PR #121, opened 2014-04-17
Added Username / Password Prompt when They're not Supplied to Login()

undercase:
This makes it easier for open source projects using Gspread (such as Reddit Bots) to use it without storing their usernames / passwords in plaintext.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -9,6 +9,7 @@ Google Data API.
 
 """
 import re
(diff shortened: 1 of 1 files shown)


burnash:
Thanks for the pull request. Sorry for the delay in reply. Since ClientLogin support was removed from Google API I'm closing this PR as obsolete.

PR #142, opened 2014-07-17
Let range accept integers

eteq:
This fairly simple PR adds an interface convenience to `Worksheet.range`.  In master, range only accepts range strings like 'A1:C4'.  As a result, I regularly find myself doing things like `ws.range(ws.get_addr_int(1, 1) + ':' + ws.get_addr_int(4, 3)`.  This PR just makes it so that `range` can accept integer sequences _in addition_ to the current behavior.  So, e.g., I can just do `ws.range((1,1,4,3))` after this.

I could also change it to use `*args` instead, allowing either `ws.range('A1:C4')` or `ws.range(1,1,4,3)`, but I wasn't sure if there's a preference against `*args` in `gspread`'s API or not.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -311,13 +311,20 @@ class Worksheet(object):
                                                   self._cell_addr(row, col))
         return Cell(self, feed)
 
-    def range(self, alphanum):
+    def range(self, alphanumorrowcol):
         """Returns a list of :class:`Cell` objects from specified range.
 
-        :param alphanum: A string with range value in common format,
-                         e.g. 'A1:A5'.
+        :param alphanumorrowcol: A string with range value in common format,
+                                 e.g. 'A1:A5', or a sequence of integers
+                                 ``(startrow, startcol, endrow, endcol)``
 
         """
+        if isinstance(alphanumorrowcol, basestring):
+            alphanum = alphanumorrowcol
+        else:
+            startalphanum = self.get_addr_int(*alphanumorrowcol[:2])
+            endalphanum = self.get_addr_int(*alphanumorrowcol[-2:])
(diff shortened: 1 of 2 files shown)


tehfink:
Thanks, I needed this functionality!

eteq:
So given that this is still unconfilicted and apparently useful... @msuozzo  or @burnash, think this can be merged?

tehfink:
Perhaps the API has changed since the pull request, but while running the tests, an error from Google popped up: 
> gspread.exceptions.HTTPError: 400: b'Neder\xc4\xabga vaic\xc4\x81juma parametra v\xc4\x93rt\xc4\xabba, kas paredz\xc4\x93ta range.'

Also, the docstring for the range function should be updated to note that it expects a tuple of values.

For the moment, I'm using the R1C1 notation, eg:
`ws.range( 'R%iC%i:R%iC%i' % (start_row, start_col, len(objects)+start_row, len(title_fields)+start_col) )`

eteq:
Oh, curious.  I'm guessing you're on py 3?  It might be that I only tested this on py 2 at the time it was written... Can you try the same on py 2.7, @tehfink?

tehfink:
Yes, I'm using Django 1.9 & Python 3.4 at the moment

burnash:
@eteq I'm sorry for the delay. Thank you very much for the contribution. 

I'm ready to merge this but I think, you can make the code more "Pythonic" if you avoid `basestring` comparison and use an [EAFP](https://docs.python.org/2/glossary.html#term-eafp) approach.

Something like this:

~~~python
try:
    (startrow, startcol, endrow, endcol) = map(int, alphanumorrowcol)
except ValueError:
    ...
~~~

PR #152, opened 2014-08-15
added test for Worksheet "updated" property

Tafkas:


burnash:
Thanks for the pull-request. I've removed unnecessary print statement and slightly simplified regexp

PR #156, opened 2014-08-20
add clear method to Worksheet

lexual:


joaofranca:
Nice helper. Works as intended.

msuozzo:
I'm not sure I agree with the `value` param here. `clear` should do exactly what one would expect: clear the spreadsheet. 

Also, the args to `get_addr_int` should be `self.row_count` and `self.col_count`.

Aside from these nits, this is definitely a good addition. If you could write a quick test, I'd love to merge this.

tracicot:
Love this addition. Works much faster than the del_row branch to clear out the sheet.

PR #171, opened 2014-10-07
Adding an insert_row function

jab416171:


canercandan:
Just tried to use this function but it's desperately slow and I dont know why, may an infinite loop be the reason ?

burnash:
I'm not aware of your exact case, but unless you have an infinite loop in your code sample there's no infinite loops. It's slow because there's an API call per each requested cell. And each API call is slow.

How many cells do you have?

MarquisT:
Does this insert_row currently still move individual cells around? It seems that the google api now supports an  insert_row function that should take care of this and speed things up.

jab416171:
Nope, @msuozzo took the time to update the code to utilize `Worksheet.range` which has significantly improved the performance of this function.

kwaqar816:
what about insert_rows()?? I guess we can add multiple rows as a list of lists. But what is row parameter inside insert_rows() method. Because it has no index parameter.

PR #188, opened 2014-12-03
Added the `numeric_value` attribute to `Cell`

ddoskind:
The `Cell` model has been extended to include a `numeric_value` attribute, which is set to the `numericValue` attribute of the cell's XML element, or `None` if the attribute does not exist on the element.

This is useful in cases where the cell's text value has been formatted with less precision than what is actually available.

It might also help to detect cells whose values have been formatted as date or date time.

Note that the test indirectly inserts a numerical value into a cell by using a formula, this is to avoid issues with spreadsheets that use a locale with a different decimal mark.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -624,6 +624,8 @@ class Cell(object):
         self._row = int(cell_elem.get('row'))
         self._col = int(cell_elem.get('col'))
         self.input_value = cell_elem.get('inputValue')
+        numeric_value = cell_elem.get('numericValue')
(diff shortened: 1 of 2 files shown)


brammator:
Are numeric_value and input_value readonly attributes?

burnash:
I don't think the term "read-only" strictly applies to these attributes. If you try to set them to a different value no exception is raised. However setting these attributes doesn't affect a spreadsheet's cell value after you call `Worksheet.update_cells()`.

PR #204, opened 2015-02-12
Clarified what goes into the OAuth2Credentials object

migurski:
The documentation for `oauth2client` is piss-poor, so I worked through [one of their examples](https://github.com/google/oauth2client/blob/master/samples/oauth2_for_devices.py) to make a valid credentials object. Because there are other libraries that speak OAuth 2.0 and feature better documentation than `oauth2client`, this change might make it easier for people to use `gspread`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ wks = gc.open("Where is the money Lebowski?").sheet1
 
 ```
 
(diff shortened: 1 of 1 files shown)


burnash:
Good point! Thank you!

migurski:
Thanks! :100:

Issue #223, opened 2015-04-21
`ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)` When opening sheet


hanleyhansen:
After successfully authenticating I'm getting an SSL error whenever I try to open a sheet.

```python
scope = ['https://spreadsheets.google.com/feeds']
credentials = SignedJwtAssertionCredentials(GOOGLE_AUTH['client_email'], GOOGLE_AUTH['private_key'], scope)
gc = gspread.authorize(credentials)

sh = gc.open('Genesis-QuestionBank')
```

```
Traceback (most recent call last):
  File "genesis/bin/import_questions.py", line 27, in <module>
    sh = gc.open('Genesis-QuestionBank')
  File "/Users/administrator2/.virtualenvs/genesis/lib/python2.7/site-packages/gspread/client.py", line 134, in open
    feed = self.get_spreadsheets_feed()
  File "/Users/administrator2/.virtualenvs/genesis/lib/python2.7/site-packages/gspread/client.py", line 220, in get_spreadsheets_feed
    r = self.session.get(url)
  File "/Users/administrator2/.virtualenvs/genesis/lib/python2.7/site-packages/gspread/httpsession.py", line 79, in get
    return self.request('GET', url, **kwargs)
  File "/Users/administrator2/.virtualenvs/genesis/lib/python2.7/site-packages/gspread/httpsession.py", line 71, in request
    uri.scheme + uri.netloc].request(method, url, data, headers=request_headers)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/httplib.py", line 1001, in request
    self._send_request(method, url, body, headers)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/httplib.py", line 1035, in _send_request
    self.endheaders(body)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/httplib.py", line 997, in endheaders
    self._send_output(message_body)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/httplib.py", line 850, in _send_output
    self.send(msg)
File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/httplib.py", line 812, in send
[...]

PR #228, opened 2015-05-06
add SSL default context to enable SSL certificate verification

danielcb:
All newer versions of python enabled certificate verification for the system
libs (e.g. httplib). Now one either needs to disable it again by hand or enable
some sort of certificate database against certificates are verified.

ssl.create_default_context() uses the system's trusted certificates.

https://docs.python.org/2/library/ssl.html#ssl.create_default_context

### DIFF
--- a/gspread/httpsession.py
+++ b/gspread/httpsession.py
@@ -14,7 +14,6 @@ try:
     from urllib import urlencode
 except ImportError:
     from http import client
-    import ssl
     from urllib.parse import urlparse
     from urllib.parse import urlencode
 
@@ -24,6 +23,7 @@ except NameError:
     basestring = unicode = str
 
 
+import ssl
 from .exceptions import HTTPError
 
 
--- a/gspread/httpsession.py
+++ b/gspread/httpsession.py
@@ -14,6 +14,7 @@ try:
     from urllib import urlencode
 except ImportError:
     from http import client
+    import ssl
     from urllib.parse import urlparse
     from urllib.parse import urlencode
 
@@ -53,7 +54,7 @@ class HTTPSession(object):
         if not self.connections.get(uri.scheme + uri.netloc):
             if uri.scheme == 'https':
                 self.connections[
-                    uri.scheme + uri.netloc] = client.HTTPSConnection(uri.netloc)
+                    uri.scheme + uri.netloc] = client.HTTPSConnection(uri.netloc, context=ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH))
             else:
                 self.connections[
                     uri.scheme + uri.netloc] = client.HTTPConnection(uri.netloc)


danielcb:
regarding issue #223

danielcb:
@burnash anything wrong/missing?

burnash:
@danielcb I'm sorry for the delay. It's not about missing something, I just didn't get enough time to figure out how to test the PR before I merge it. What steps should I follow?

danielcb:
Not a real test in the sense of a e.g. a unit test, but you can connect to a SSL backend and it will fail on all versions from [2.7.9](https://docs.python.org/2/library/ssl.html#context-creation) and [3.4](https://docs.python.org/3.4/library/ssl.html#ssl.create_default_context) on up because of the changes made regarding [PEP0476](https://www.python.org/dev/peps/pep-0476). Versions below that won't see the error from #223.

"Fixes" for that are:
1. [Opting out](https://www.python.org/dev/peps/pep-0476/#opting-out)
2. Using your own context (i.e. bring your own certificate chain, [requests](http://docs.python-requests.org/en/latest/) does this afaik), but this means you regularly need to update them, since they are only valid for a finite period of time
3. [Create a default context](https://docs.python.org/3.4/library/ssl.html#ssl.create_default_context) and thereby use the system certificate system.

I chose 3.

burnash:
@danielcb Thanks for the detailed description. This does make sense.

I tried to run this fix on 2.7.3 and I got "NameError: global name 'ssl' is not defined".

Here's a full traceback:

```python
  File "sampleoauth.py", line 26, in <module>
    sh = gc.open("aspreadsheet")
  File "../gspread/gspread/client.py", line 143, in open
    feed = self.get_spreadsheets_feed()
  File "../gspread/gspread/client.py", line 229, in get_spreadsheets_feed
    r = self.session.get(url)
  File "../gspread/gspread/httpsession.py", line 80, in get
    return self.request('GET', url, **kwargs)
  File "../gspread/gspread/httpsession.py", line 57, in request
    uri.scheme + uri.netloc] = client.HTTPSConnection(uri.netloc, context=ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH))
NameError: global name 'ssl' is not defined
```

danielcb:
mh, looks like it is not available in 2.7.3. Maybe the simplest way is to try and use the other way if it is not available on older versions of python. E.g.:

```
try:

```

burnash:
`ssl` [is available](https://docs.python.org/2/library/ssl.html) in 2.7.3, it's just the `import ssl` [is hidden by](https://github.com/danielcb/gspread/commit/2cbdeee3484cd068174a35982ae39e1577138b82#diff-652601946058f6086bc4e7944514cdcdR17) the `except` clause.

burnash:
`ssl` [is available](https://docs.python.org/2/library/ssl.html) in 2.7.3, it's just the `import ssl` [is hidden by](https://github.com/danielcb/gspread/commit/2cbdeee3484cd068174a35982ae39e1577138b82#diff-652601946058f6086bc4e7944514cdcdR17) the `except` clause.

danielcb:
sry, my bad. I moved it outside the except-block.

danielcb:
Hi, still no time for further testing or something I can do?

danielcb:
closed, since #255 seems like the more general, better approach.

PR #235, opened 2015-05-27
added workaround for python3 users

marsjoy:
@burnash I added a workaround for Python 3 users in the oauth docs.


burnash:
Thanks for the PR! I've changed the workaround slightly: removed  the `try / except` block in favor of a descriptive example. This way the `except TypeError` block won't hide any other surprises from `oauth2client`.

marsjoy:
@burnash Thanks for making this library. We use it a ton at Udemy. :+1:

PR #254, opened 2015-06-15
Paddle insert_row if len(columns) > len(values)

chenyu-yang:
If there are more columns than values, we need to paddle some values (I used empty string) to the values. Otherwise, it preserves the original cell in the end of the inserted row.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -550,6 +550,8 @@ class Worksheet(object):
         data_width = len(values)
         if self.col_count < data_width:
             self.resize(cols=data_width)
+        else:
+            values += [''] * (self.col_count - data_width)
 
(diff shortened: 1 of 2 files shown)


burnash:
Thanks for the pull request. Can you add a test case for this? 

The test suite is in https://github.com/burnash/gspread/blob/master/tests/test.py

demipenguin:
Thanks for the reply. I have added a test case for it.

burnash:

Apparently, this issue has been fixed in #283. I'm unable to reproduce this issue. 

Recently I've ported the library to API v4 which has native support for rows insertion.

I'm closing this PR now. Again sorry for the delay in response.

PR #255, opened 2015-06-15
``ssl.SSLContext`` can be passed as argument to authenticate and login

lajarre:
An `ssl.SSLContext` object as argment of `authenticate` or `login` is passed through to the `HTTPConnection` object.

After an OS X update, my computer was not ok verifying the remote SSL host, through the default SSL context.

```
SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)
```

Passing the context directly seemed to me like the simplest and most direct way to make it possible to the developer to control the way the machine is validating the Google servers certificate.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -41,6 +41,7 @@ class Client(object):
                  oauth2client library. https://github.com/google/oauth2client
     :param http_session: (optional) A session object capable of making HTTP requests while persisting headers.
                                     Defaults to :class:`~gspread.httpsession.HTTPSession`.
+    :param ssl_context: ``sslSSLContext`` object, if non-default needed.
 
     >>> c = gspread.Client(auth=('user@example.com', 'qwertypassword'))
 
@@ -50,9 +51,9 @@ class Client(object):
 
 
     """
-    def __init__(self, auth, http_session=None):
+    def __init__(self, auth, http_session=None, ssl_context=None):
         self.auth = auth
-        self.session = http_session or HTTPSession()
+        self.session = http_session or HTTPSession(ssl_context=ssl_context)
 
     def _get_auth_token(self, content):
         for line in content.splitlines():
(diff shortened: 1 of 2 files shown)


danielcb:
I think it would be easier and safer if the default would be an ssl defaul context (i.e. ```ssl.create_default_context()```) instead of ```None```.I don't think this would have any unreasonable side effects, but am not 100% sure. I've done this in #228, but the possibilty to overwrite it is a nice feature.

lajarre:
This None thing all boils down to the code in httplib.py and you can see that None becomes `ssl._create_default_https_context()` (which is the same as `ssl.create_default_context()` in my version of ssl lib. So from my point of view, passing None is the right way (as expected by httplib) to invoke the default context.

Seems ok for you?

danielcb:
Yep, sounds good. I'll close my pull request then :)

spex66:
I was hit on Windows 10, Python 2.7.10 by:
```
SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
```

Relying on context ```None``` using the default which is ```ssl.create_default_context()``` didn't made a difference on Windows (for me).

So I've used the two patched files from this pull request, and override my security context.

So I've taken the shortcut, which is the "unverified security context":
```
    # auth for gspread
```

To:
```
    import ssl
    # auth for gspread
    gc = gspread.authorize(credentials, ssl._create_unverified_context())
```

This way I can at least access the sheets as intended.
If someone has a clever idea, how to solve this properly and secure, I'm really interested :)

spex66:
And finally I'm back to the unpatched "gspread" with an ugly monkeypatch, because I've no clue how to solve this properly on Windows.

Most likely this is related to Python >= 2.7.9 and the not working fallback to the default SSL context is maybe on Windows only:
```python
# PEP https://www.python.org/dev/peps/pep-0476/
# Monkeypatch: https://github.com/abeloin/SickRage/commit/21fcea2c3155f57dba5c64a55f1f8fee02804256
import sys
if sys.version_info >= (2, 7, 9):
    import ssl
    ssl._create_default_https_context = ssl._create_unverified_context
```

meg2208:
Can we get this merged soon? I am having the same issue.

HendrikBrackmann:
upgrading python from 3.4 to 3.6 and reinstalling dependencies worked for me.

burnash:

gspread has long ago been ported to [requests](http://docs.python-requests.org/) for making HTTP requests. I'm afraid this PR is outdated. 

SSL now can be configured and managed by means of `requests` itself. Here's the link to the relevant docs: http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

I'm closing this PR. Again sorry for the delay in response.

PR #263, opened 2015-06-26
Fixed HTTPError usage

jsousa1919:
I know the usage is deprecated but this Exception is still used in httpsession.py so for now it should at least do what it was intended for


burnash:
Thanks for the PR. 

The `HTTPError` is a remnant of [urllib2](https://docs.python.org/2/library/urllib2.html) (later replaced with [httplib](https://docs.python.org/2/library/httplib.html)). I plan to move to a [requests](https://github.com/kennethreitz/requests)-compatible http session, but for now this should fix the incorrect error in `Client.put_feed`.

Issue #67, opened 2013-07-27
append_row not working.


csu:
When I run the following code, it successfully retrieves the values in cells A1 and B1, but it doesn't append the new row.

```
import gspread
import AccountDetails

gc = gspread.login(AccountDetails.GSPREAD_USERNAME, AccountDetails.GSPREAD_PASSWORD)
doc = gc.open("Development Sandbox").sheet1
print doc.acell('A1').value + ", " + doc.acell('B1').value
rowToAdd = ['test', 'test2']
doc.append_row(rowToAdd)
```

I've also tried: doc.append_row(['A','B','C']) in place of the last line, and that didn't work either.

PR #274, opened 2015-07-09
Note that append_row() may be right at the bottom of a new sheet.

longhotsummer:
This helps to clarify issues like #67.


burnash:
Indeed, thank you!

Issue #270, opened 2015-07-06
append_row adding a blank data row


Micchiu:
hello, 
I am using append_row() to append temperature and humidity readings ina google sheet.
The code makes a reading periodically, each 20 minutes, the adds a row with date, time, temp, humi, etc. etc.
Now, sometimes (apparently, with a random interval), it adds a completely blank row.
I've been on the code for long, with no solution.
Maybe someone has a clue?
Thank you in advance.

Shiva-Shinde:
append_row() basically pushes the data in the bottom most row of the sheet. I mean, if you have 1000 rows, your data should be in 1001 or 1002 (not sure of the row number) row of the sheet.

So, I suggest go with insert_row(). Thanks!!

Micchiu:
@Shiva-Shinde hello and thank you for the reply.
However, that is not the problem.
For other reasons, I need to use append_row().
I am just interested to know if others observed it to unexpectedly write blank lines, when the added variable values are not void.
thx :-)

msuozzo:
Are you able to isolate this problem any more? Have you been able to recreate it?

burnash:
I'm closing this issue for inactivity. Please feel free to reopen it if you have further input.

PR #283, opened 2015-08-14
Altered insert_row semantics to utilize range

msuozzo:
This avoids issuing one API request per cell to retrieve the Cell
objects after the insertion row. This provides a significant speed-up
for insertions at the beginning of large sheets.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -550,23 +550,33 @@ class Worksheet(object):
 
         :param values: List of values for the new row.
         """
+        if index == self.row_count + 1:
+            return self.append_row(values)
+        elif index > self.row_count + 1:
+            raise IndexError('Row index out of range')
+
         self.add_rows(1)
         data_width = len(values)
         if self.col_count < data_width:
             self.resize(cols=data_width)
 
-        all_cells = self.get_all_values()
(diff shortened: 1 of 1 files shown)


msuozzo:
Uhhh. I assume Travis failing isn't normal. Anything I can do to fix that? (Sorry but I've never done this before.)

msuozzo:
I think this also fixes #270

burnash:
Thanks for the PR. Never mind Travis, the set up is still WIP.

msuozzo:
Gotcha.

burnash:
Great work! This is a significant speed-up for the row insertion, indeed.

miklevin:
You guys ROCK! Thanks from a lurking observer who appreciates gold when he sees it.

greg-jaunt:
This change appears to have a serious negative side-effect in that it clears all formulas in the spreadsheet below the insert point.  Is there a way to prevent this?

greg-jaunt:
Looks like my problem is #284.

greg-jaunt:
#284 tracks the issue I'm seeing, which appears to pre-date this patch.  Monkey patching the fix from #338 into gspread 0.4.0 seems to resolve the issue for me.

Issue #190, opened 2014-12-19
Super slow


pryorda:
This Script takes almost 20 minutes to parse through 186 rows.

```python
bufsize = 1
with open('/cygdrive/c/hsmtest_adep/final_batch.sh', 'w', bufsize) as f:
    for row in range(3, 186):
        sentry = ''
        counter = 0
        while not sentry and counter < 3:
            try:
                gc = gspread.login('xxxxx', 'xxxxx') # If i do this outside the loop i see intermittent failures
                sht = gc.open_by_key('xxxx')
                worksheet = sht.worksheet("xxxxxx")
                values_list = worksheet.row_values(row)
                sentry = True;
            except:
                print("Row:", row, "Got Worksheet Error, Trying Again...")
                counter += 1

        try:
           print("./test_alias.sh -a", values_list[2])
           if values_list[2]:
               values = "./test_alias.sh -a " + values_list[2] + "\n"
           string = str(values)
           f.write(string)
        except IndexError:
           print("No Cert Alias for row id:", row)

f.close()
```

samuelcolvin:
Real shame that @pryorda hasn't bothered to format his example properly, it does make it unreadable. Simply putting triple backquote round the code would probably do it.

That said speed of using this package to scrape spreadsheets is a real problem. Are you sure there are no ways of improving performance?

If i do 

    cell_list = sheet.range('A1:C100')

then iterate through the result is it doing one (or multiple) request(s) per cell or getting the entire range in one request? 

I'm a big fan of gspread, thanks. I'm only being demanding because I really want to use it more.

samuelcolvin:
OK I've had a play and it does only seem to do one request per range, however (and I haven't looked through it in detail) there is vast amounts of data return when even a small range is looked up.

Would there be any way to improve the query so less data is returned? Are you sure the range query is working correctly?

samuelcolvin:
[get_all_records](http://burnash.github.io/gspread/#gspread.Worksheet.get_all_records) came to my rescue, much faster than `range` for entire sheet.

Thanks again for gspread.

burnash:
@samuelcolvin many thanks for the feedback, I'm happy the library is useful. 

Under the hood both `range` and `get_all_records` make only one HTTP API request to Google's servers: 

    GET https://spreadsheets.google.com/feeds/cells/key/worksheetId/private/full 

The only difference is that `range` passes parameters to this API call:

```python
def range(self, alphanum):
[...]

Issue #179, opened 2014-11-04
col_values extremely slow


AndyFang1998:
It takes > 20 seconds to scrape one row.

Issue #113, opened 2014-03-20
row_values() takes long time in worksheets with lots of data


acilander:
Hello,

I use row_values(1) to get the headings of a table. Now it happens that my script becomes very slow in this line. My worksheet dimension is 23 cols and 2500 rows.

In models.py there is 
cells_list = self._fetch_cells() in _list_values(). This looks to me that the complete data in the worksheet is queried before getting just one row. Is there a way around that?

PR #286, opened 2015-08-17
Optimized row_values and col_values

### LINKED ISSUE #285: Slow Retrieval of Records
Just using gspread for the first time in a django project. Got everything installed, and functioning, but veeeerrrrryyy slowly. To run the following code, it takes 40 seconds. And I'm only returning one row from a spreadsheet.

```
    gc = authenticate_google_docs()
    sh = gc.open_by_key('spreadsheet-key-here') 
    worksheet = sh.worksheet("spreadsheet_title") 
    row1values = worksheet.row_values(1)
    print row1values
```

Any idea why?

Could it have anything to do with this message that I am receiving? :

```
 InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
```
Did I ask this question in the right place?

msuozzo:
Removed the _fetch_cells call for each method. This eliminates the
adverse effect on runtime for large worksheets.

Fixes #285, #190, #179, and #113

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -378,31 +378,17 @@ class Worksheet(object):
 
         return [dict(zip(keys, row)) for row in values]
 
-    def _list_values(self, index, cell_tuple, position):
-        cells_list = self._fetch_cells()
-        cells = dict(map(cell_tuple, filter(position, cells_list)))
-
-        try:
-            last_index = max(cells.keys())
-        except ValueError:
-            return []
-
-        vals = []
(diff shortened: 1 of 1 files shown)


burnash:
This looks like a part of #283

PR #288, opened 2015-08-24
Handle CannotSendRequest exception

tacy:
Clearing httpsession's connection cache when get an CannotSendRequest exception


burnash:
@tacy sorry for the late response.

We have ported the library to use `requests.Session` for HTTP (#289). This should resolve the issue with `CannotSendRequest` exception. The code in this PR refers to the older pre-`requests` version of the library. I'm afraid it's no longer has any effect and I have to close the PR.

Again sorry for the late response and thank you for the contribution.

PR #289, opened 2015-08-24
Use Python requests

### LINKED ISSUE #81: Support clients behind a proxy
Hi,

maybe I am missing something super easy but... I can't manage to log in when standing behind a proxy. I tried to set up environment variables without success. Googling a bit it seems httplib is not handling proxy that well. Maybe that's the issue.

Thanks for this great lib!

suminb:
I'm in a situation where I need to connect to Google Docs servers via an HTTPS proxy. The previous code was set up in a way that it does not read in `HTTP_PROXY` and `HTTPS_PROXY` environment variables. I was hoping there was an easy way to pass proxy server configurations to the `http.client.HTTPSConnection` object, but it turned out there is none. You may use proxy servers with `HTTPSConnection` as follows, but it's somewhat cumbersome.

    proxy_host = 'yourproxy.com'
    proxy_port = 3128
    conn = client.HTTPSConnection(proxy_host, proxy_port)
    conn.set_tunnel(url)
    conn.request(...)

Some might argue this isn't too bad, but given there are four different possible scenarios, the code can be unnecessarily long:

- HTTP without proxy
- HTTPS without proxy
- HTTP with proxy
- HTTPS with proxy

So I thought it would be a wiser choice to replace the `HTTPSConnection` stuff with [Python requests](http://www.python-requests.org/). You will be able to use proxy servers by setting `HTTP_PROXY` and/or `HTTPS_PROXY` environment variables.

This shall resolve #81 and other proxy-related issues.


suminb:
CI failed. I'll look into it.

burnash:
@suminb Never mind the CI. I haven't finished the set up.

scottk92:
Hey guys! I'm doing a project using gspread (it's amazing btw!) but I'm running into a proxy issue that I think suminb's fix will address quite nicely. Is there an ETA for when his branch will get merged into master?

burnash:
@scottk92 Could you elaborate on the proxy issue details?

scottk92:
Sorry for the late reply @burnash! Basically my web application is running on a machine that has restricted Internet access and all HTTP calls outside of our corporate network needs to go through a proxy for auditing and security reasons. For all my 3rd party API calls, I either set an environment variable for the proxy or directly insert proxy information in an http object that I instantiate myself. I really wanted to use gspread instead of the gdata library for interacting with Google Spreadsheets because it's so simple and easy to use, but the proxy issue is preventing me from doing that. I tested my code with @suminb's branch of gspread and it works great! It would just be nice if his changes can get merged to master soon.

msuozzo:
This is a very good dependency to add. It's a fantastic package and will make adding new connection-dependent features (_e.g._ timeouts) trivial. :+1:

actioncrypto:
Please please please someone add this to the master branch so it can be easily upgraded via pip. I cannot figure out how to do this and maintain pip functionality... needs some form of timeout handling. Google API keeps dying & it kills the code without implementing this fix...

burnash:
@actioncrypto, this is how you can install @suminb's branch with pip:

~~~bash
pip install requests
pip install git+https://github.com/suminb/gspread.git@feature/use-requests
~~~

scottk92:
Are we still planning on merging @suminb's to master?

burnash:
@scottk92 yes, indeed.

gspread/client.py
```diff
@@ -112,8 +112,7 @@ def login(self):
 
             try:
                 r = self.session.post(url, data)
-                content = r.read().decode()
-                token = self._get_auth_token(content)
+                token = self._get_auth_token(r.content)
```
msuozzo: Would you mind explaining your choice of `.content` over `.text`?

requirements.txt
```diff
@@ -0,0 +1 @@
+requests>=2.2.1
```
burnash: I don't think we need `requirements.txt` in the project. As far as I know this won't implicitly tell installer to install Requests. Instead this line [should go](https://pythonhosted.org/setuptools/setuptools.html#declaring-dependencies) to `install_requires` argument of `setup()` in [`setup.py`](https://github.com/burnash/gspread/blob/bdf96e7d4d6e5c1fe355bb6d100cb6812bc25d7f/setup.py):

```python
setup(
   ... 
   install_requires=['requests>=2.2.1']
   ...
)
```

@suminb could you remove `requirements.txt` and update `setup.py`?

msuozzo: Theoretically we could add `requirements.txt` and just populate the `install_requires` list from the file's lines when setup is run. I feel like it's a pretty ubiquitous feature of python libraries at this point. That being said, there are so few requirements, it's kind of silly having it.

gspread/client.py
suminb: @msuozzo Good point. It should've been `.text`. I'll change them.

Issue #271, opened 2015-07-08
navigate to worksheet by key and gid


pita2k:
is it possible to add a feature that will return a worksheet class by key and by gid ?

msuozzo:
@m4rc1e if you want to take a crack at this, please patch away.

m4rc1e:
@msuozzo I'll have it sorted soon.

wolverdude:
Whatever happened to this? It would be super useful to get a reference to a worksheet by copy-pasting the URL, and this feature is a prerequisite to that.

burnash:
@wolverdude could you please share an example usage for this / your use case scenario?

wolverdude:
What I would like is something like

```py
gc.get_worksheet_by_url('https://docs.google.com/spreadsheet/ccc#gid=0Bm...FE
```

This would return a specific tab from a spreadsheet. I feel like that’s a per common use-case, just wanting some specific data from a specific sheet, and it’s easy to just copy-paste the url.

This issue would be a perquisite to that, and at least halfway there. I envision the API would be something like:

```py
sh.get_worksheet _by_id(1234567890)
```

wolverdude:
I’d be happy to put in a PR for it.

burnash:
Great, the PR is welcome!

lavigne958:
closed by #857

PR #298, opened 2015-09-24
Iterworksheets

m4rc1e:
Thank you for making such an incredibly easy to use api I still think certain special methods could make life just a little easier. 

My addition allows for iteration through a spreadsheet's worksheets. I understand if you do not like this implementation and would rather have people use the worksheets() method for iteration instead. 

I see issue #271 has asked for supporting keys. I will happily add this as well.

Kind regards,
Marc


burnash:
@m4rc1e Thanks for the PR and for the test. The implementation is very clear.

PR #307, opened 2015-10-11
DOC: documentation about encoding private_key

tacaswell:
Directly call `encode` on the `str` rather than implicitly through bytes.

### DIFF
--- a/docs/oauth2.rst
+++ b/docs/oauth2.rst
@@ -81,7 +81,7 @@ Depending on your system setup you may need to install PyOpenSSL:
 
     wks = gc.open("Where is the money Lebowski?").sheet1
 
-**Note**: Legacy Python users do not need to encode ``json_key['private_key']`` due to ``str`` and ``bytes`` not being differentiated.
+**Note**: Python2 users do not need to encode ``json_key['private_key']`` due to ``str`` and ``bytes`` not being differentiated.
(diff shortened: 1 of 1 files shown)


msuozzo:
Is Python 2.7 really "legacy"? I would suggest that you rephrase this to be explicit about Python 2 vs 3. Other than that, looks great.

tacaswell:
The cpython devs have made it clear that there will not be a 2.8 series and
bug fixes for 2.7 will stop in 2020 (and the only reason for _that_ is some
really broken crypto in core).

Fedora is defaulting to 3 for the next release, RHEL has committed to
switch eventually, and the debian/unubuntu world is making good progress on
switching.  It is no longer a question of _if_ things will flip to 3, it is
a question of when.

msuozzo:
Not saying that 3 won't happen. Just saying that people might not be familiar with the concept of Python 2 as "legacy". I agree that it should be changed but I'd say a good time to call it legacy is when Ubuntu and Mac come with 3 as standard. Until that time, I think the distinction need be made.

tacaswell:
Fair enough.  I have an agenda and was seeing how far I could push it ;)

Added a commit that makes the note less obnoxious.

msuozzo:
:) understandable. Merging.

PR #308, opened 2015-10-12
Added option to retry http requests upon failure

finoptimal-dev:
When creating a Client, passing in tries (an int that otherwise defaults to 1)
causes the HTTPSession.request() method to try the request that number of times.


msuozzo:
Thanks for the PR!

I think something like the `tries` parameter is a good idea. That being said, we're either deprecating HTTPSession or heavily gutting it in favor of the [requests](http://docs.python-requests.org/en/latest/) package in the very near future. If we decide to go another route or run into issues with requests, I'll be sure to merge this first thing. Thanks again.

finoptimal-dev:
Thanks; requests is good. When is that planned?

burnash:

gspread has long ago been ported to [requests](http://docs.python-requests.org/) for making HTTP requests. I'm afraid this PR is outdated. 

Retry logic can now be managed by means of `requests` itself. Here's the link to the relevant SO question: https://stackoverflow.com/questions/15431044/can-i-set-max-retries-for-requests-request

I'm closing this PR. Again sorry for the delay in response.

PR #310, opened 2015-10-12
Adding hint to share document with service-account in Readme

pc-coholic:
The Readme is missing an important hint: If you want to be able to edit a file with the .json-credentials, you need to share it with the service-account created.

Adding this hint to README - I am sure there are people out there who run into this problem, too.


burnash:
I'm not convinced that there's a point in reiterating the message from the linked guide (step 7) in the readme file. However, it might be that the sentence in the *guide* needs improvement to better convey the hint. I'm open to rephrasing or text-formatting suggestions.

PR #326, opened 2015-11-06
Use lxml if available

ptwobrussell:
Refs #2 - Attempt to import `lxml` and revert to `xml` if it's not available. I've observed roughly a 3x speedup on parsing large spreadsheets (reduction of ~24 seconds for thousands of rows down to ~8 seconds) with `lxml`. A fairly simple change and all of the tests pass.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -11,7 +11,12 @@ Google Data API.
 import re
 import warnings
 
-from xml.etree import ElementTree
+try:
+    from lxml import etree
+    fromstring = etree.fromstring # Roughly 3x faster parsing for large worksheets
+except ImportError:
+    from xml.etree import ElementTree
+    fromstring = ElementTree.fromstring
 
 from . import __version__
 from . import urlencode
@@ -229,7 +234,7 @@ class Client(object):
                             visibility=visibility, projection=projection)
 
         r = self.session.get(url)
-        return ElementTree.fromstring(r.content)
+        return fromstring(r.content)
 
     def get_worksheets_feed(self, spreadsheet,
                             visibility='private', projection='full'):
@@ -237,7 +242,7 @@ class Client(object):
                             visibility=visibility, projection=projection)
 
         r = self.session.get(url)
-        return ElementTree.fromstring(r.content)
+        return fromstring(r.content)
 
     def get_cells_feed(self, worksheet,
                        visibility='private', projection='full', params=None):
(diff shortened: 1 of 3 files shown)


gspread/models.py
```diff
@@ -12,8 +12,12 @@
 from collections import defaultdict
 from itertools import chain
 
-from xml.etree import ElementTree
-from xml.etree.ElementTree import Element, SubElement
+try:
+    from lxml.etree import ElementTree
+    from lxml.etree.ElementTree import Element, SubElement
```
burnash: This import raises `ImportError` on my system (Python 2.7.10, lxml 3.4.4):

~~~
>>> from lxml.etree.ElementTree import Element, SubElement
ImportError: No module named ElementTree

>>> import lxml
>>> lxml.etree.__version__
u'3.4.4'

~~~

So for this module the `xml.etree` version is imported in the `except` clause.

The [compatibility guide](http://lxml.de/compatibility.html) suggests this way of import:

~~~
from lxml.etree import Element
~~~

gspread/models.py
```diff
@@ -12,8 +12,12 @@
 from collections import defaultdict
 from itertools import chain
 
-from xml.etree import ElementTree
-from xml.etree.ElementTree import Element, SubElement
+try:
+    from lxml.etree import ElementTree
```
burnash: For lxml 3.4.4 this imports a wrong object:

~~~python
>>> from lxml.etree import ElementTree
>>> ElementTree.__doc__
'ElementTree(element=None, file=None, parser=None)\n\n    ElementTree wrapper class.\n    '
>>> ElementTree.tostring
AttributeError: 'builtin_function_or_method' object has no attribute 'tostring'
~~~

Wherein `etree` is actual `ElementTree` implementation.

~~~python
>>> from lxml import etree
>>> etree.__doc__
'The ``lxml.etree`` module implements the extended ElementTree API\nfor XML.\n'
~~~

This error is hidden since the `ImportError` exception in a line below leads to import of the stock `ElementTree`.

The [lxml docs](http://lxml.de/compatibility.html) suggest this way of the compatibility import:

~~~python
from lxml import etree as ElementTree
~~~

gspread/client.py
```diff
@@ -11,7 +11,12 @@
 import re
 import warnings
 
-from xml.etree import ElementTree
+try:
+    from lxml import etree
+    fromstring = etree.fromstring # Roughly 3x faster parsing for large worksheets
+except ImportError:
+    from xml.etree import ElementTree
+    fromstring = ElementTree.fromstring
```
burnash: I think both `try` and `except` can import `ElementTree` into the namespace:

~~~python
try:
    from lxml import etree as ElementTree
except ImportError:
    from xml.etree import ElementTree
~~~

This way the change will be even simpler.

burnash:
Great feature. Thank you for the PR. The tests are running almost twice as fast. 

I reviewed the code and found some issues. I believe after eliminating those issues gspread will be even more faster.

burnash:
I'm closing this PR due to inactivity. Also, the latest version of gspread uses Sheet API v4 by default which eliminates the need to parse XML.

PR #337, opened 2015-12-19
Added del_row method (and test for it)

a-rodin:
The implementation uses that same idea as the one in `insert_row` method (just shifts rows below up and removes the last row).


synth3tk:
@burnash Is there anything that this is waiting on? It seems like a very important piece missing.

tracicot:
I need this for a project I'm working on. My workaround right now is to loop through the rows and columns and clear out each _cell_, which is extremely time consuming!

Issue #284, opened 2015-08-14
`Worksheet.insert_row` removes formulae below insertion point


msuozzo:
Because the `insert_row` performs the copy with the `Cell.value` attribute (as opposed to `Cell.input_value`), all formulae at and below row `index` are lost.

I tried to simply replace `value` with `input_value` but the request failed because of what I believe to be a Sheets API error (the cell input values contained cell references in the format `R[#]C[#]` instead of the usual `A#`)

burnash:
I went with replacing `value` with `input_value` and merged #338. Since `insert_row` copies formulas verbatim (as opposite to the row insertion in the Sheets UI, where cell refs are recalculated) it could lead to incorrect formulas. For now, I don't know any solution other than parsing formulas and essentially replicating Sheets functionality locally. But this will go way outside the scope of the library. The API v4 should change this. I think the simplest way now is to let the user handle the edge cases.

I'm closing the issue for now. Please feel free to reopen it.

PR #338, opened 2015-12-19
Fix losing input_value data from following cells in insert_row

a-rodin:
Let's suppose that one have a spreadsheet which values after the inserted row are prefixed by single quote (to avoid formula's evaluation). Then someone inserts a row before it. And then the single quote become lost. To avoid it there is a one-line patch in the pull request.


msuozzo:
Great fix. Would you be able to add a test or two for this? I vaguely remember trying this fix previously and there being an issue.

Otherwise, definitely a good feature to add. This would also address #284.

PR #348, opened 2016-01-26
Use celementtree when available to reduce memory usage

drussel:
This cuts memory usage by half for large spreadsheets.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -11,7 +11,10 @@ Google Data API.
 import re
 import warnings
 
-from xml.etree import ElementTree
+try:
+    import xml.etree.cElementTree as ElementTree
+except:
+    from xml.etree import ElementTree
 
 from . import __version__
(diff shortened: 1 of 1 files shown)


burnash:
Thanks for the contribution. How do you calculate the memory improvement?

I haven't had any experience with `cElementTree`. Are there any limitations when using it instead of `ElementTree`?

Another method to improve performance is to fallback to `lxml` when it's available. (#326)

drussel:
I'd been running jobs with large spreadsheets and just using `top` or its kin to monitor memory usage (so very crude), but it was enough of a difference to make things fit on my heroku instances that didn't before.

My understanding from reading various docs on the two is that celementtree should be a drop in replacement for elementtree other than
- requiring a compiler to build
- being faster/smaller

lxml looks to be basically equivalent for these purposes. I think I just happened on celementtree first.

PR #356, opened 2016-02-23
oauth2 documentation update

JMensch:
Oauth2client version 2.0.0 removed SignedJwtAssertCredentials, which is referenced in your authorization readme. I've updated OAuth2.rst to now properly describe how to authorize with oauth2client 2.0.0 using a replacement method, ServiceAccountCredentials.

### DIFF
--- a/docs/oauth2.rst
+++ b/docs/oauth2.rst
@@ -61,6 +61,19 @@ Depending on your system setup you may need to install PyOpenSSL:
 
 6. Now you can read this file, and use the data when constructing your credentials:
 
+::
+
+    import gspread
+    from oauth2client.service_account import ServiceAccountCredentials
+
+    scope = ['https://spreadsheets.google.com/feeds']
+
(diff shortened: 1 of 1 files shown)


msuozzo:
Thanks for the PR! A few notes:
* You can remove the `import json` because it's no longer used
* Would you mind adding a section for `oath2client`'s previous semantics? The latest version might not be universally supported yet so we should keep a note of the previous way of doing things.

Thanks again!

JMensch:
@vagelim Thanks for the catch!
@msuozzo Makes sense, I've added the section for oauth2client < 2.0.0

JMensch:
CI will pass when fix for issue #363 is integrated.

msuozzo:
Fantastic! Last nit: The most up-to-date instructions should come first and the old ones should come second prefaced with a note like "For older versions of the oauth2client (< 2.0.0):"

Thanks again!

JMensch:
@msuozzo thanks, updated!

PR #361, opened 2016-02-28
Update oauth2.rst

vagelim:
changes to support `oauth2client==2.0.0`


msuozzo:
Hey thanks for the PR! Would you mind adding another section mentioning the old semantics in case people are running older versions of the library?

Thanks again!

Issue #357, opened 2016-02-24
oauth2client error cannot import name SignedJwtAssertionCredentials


perrycannon:
I receive an error
from oauth2client.client import SignedJwtAssertionCredentials
ImportError: cannot import name SignedJwtAssertionCredentials

I have received the same error on different raspberry pis. Each had a fresh install of Jessie.

I received the error. Then installed Librium-dev, upgrade PyOpenSSL, with the same result.

I don't know how to resolve this error

Thank you
Perry

vagelim:
Ran into this issue myself. 

The solution that worked for me:
`pip install oauth2client==1.5.2`

perrycannon:
Vagelim thank you for your help. Installing oauthclient 1.5.2 resolved the error.

Now I receive this error.
"Unable to login and get spreadsheet.  Check OAuth credentials, spreadsheet name, and make sure spreadsheet is shared to the client_email address in the OAuth .json file!
Google sheet login failed with error: No crypto library available"

I have rechecked all of my entires, everything is correct. I have previously installed libffi-dev and PyOpenSSL.

Thank you
Perry

JMensch:
Please see my correction to their documentation [here](https://github.com/burnash/gspread/pull/356). This will solve your problem.

perrycannon:
Thank you. I really appreciate your help. I am now able to write to google sheets from my raspberry pi. I've been working on this for awhile. I wasn't able to get help from the creator of the project I emulated. He would only refer me to blog posts or generic guides.

PR #369, opened 2016-03-16
Update method to read credentials from file - closes #343, #356, #357

### LINKED ISSUE #343: please update your docs on setting up google oauth
it says that PyOpenSSL may be required, but I find that with python 2x it is absolutely required. The error message you get back does not make the source of the problem clear:

> > > from oauth2client.client import SignedJwtAssertionCredentials
> > > returns `ImportError: cannot import name SignedJwtAssertionCredentials`

looking on stackoverflow, I found:
http://stackoverflow.com/questions/14063124/importerror-cannot-import-name-signedjwtassertioncredentials

that oauth2client depends on PyOpenSSL but doesn't tell the user it does, so best to let people know in your otherwise exquisite documentation what this error message means.

Libraries oauth2client.client
~~~
if HAS_OPENSSL:
  # PyOpenSSL is not a prerequisite for oauth2client, so if it is missing then
  # don't create the SignedJwtAssertionCredentials or the verify_id_token()
  # method.
~~~
thanks!

vmahedia:
SignedJwtAssertionCredentials is deprecated, replaced that with
new method with reads credentials directly from the file.

https://github.com/google/oauth2client/issues/401

### DIFF
--- a/docs/oauth2.rst
+++ b/docs/oauth2.rst
@@ -65,21 +65,17 @@ Depending on your system setup you may need to install PyOpenSSL:
 
     import json
     import gspread
-    from oauth2client.client import SignedJwtAssertionCredentials
-
-    json_key = json.load(open('gspread-april-2cd … ba4.json'))
+    from oauth2client.service_account import ServiceAccountCredentials
+    
     scope = ['https://spreadsheets.google.com/feeds']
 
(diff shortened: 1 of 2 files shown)


msuozzo:
Fixed by #356

PR #370, opened 2016-03-17
Retry requests on Internal Server Errors

ddoskind:
Since last weekend, we've been getting significantly higher number of 500 error codes from the Spreadsheet API. These errors are usually transient and retrying helps.
This PR adds a very simple exponential backoff strategy to `HTTPSession` which helped a lot in our case.
Additionally, the test script is slightly modified to work with non-Service Account OAuth2 credentials as well.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -38,13 +38,16 @@ class Client(object):
                  oauth2client library. https://github.com/google/oauth2client
     :param http_session: (optional) A session object capable of making HTTP requests while persisting headers.
                                     Defaults to :class:`~gspread.httpsession.HTTPSession`.
+    :param max_retries: (optional) If using the default `http_session`, how many times to retry a
+                                   request when encountering internal server error responses.
+                                   Default is to retry four times.
 
     >>> c = gspread.Client(auth=OAuthCredentialObject)
 
     """
-    def __init__(self, auth, http_session=None):
+    def __init__(self, auth, http_session=None, max_retries=None):
         self.auth = auth
-        self.session = http_session or HTTPSession()
+        self.session = http_session or HTTPSession(max_retries=max_retries)
 
     def _ensure_xml_header(self, data):
         if data.startswith(b'<?xml'):
@@ -239,7 +242,7 @@ class Client(object):
         return ElementTree.fromstring(r.content)
 
 
-def authorize(credentials):
+def authorize(credentials, max_retries=None):
     """Login to Google API using OAuth2 credentials.
 
     This is a shortcut function which instantiates :class:`Client`
@@ -248,6 +251,6 @@ def authorize(credentials):
     :returns: :class:`Client` instance.
 
     """
-    client = Client(auth=credentials)
+    client = Client(auth=credentials, max_retries=max_retries)
     client.login()
     return client
--- a/gspread/httpsession.py
+++ b/gspread/httpsession.py
@@ -27,7 +27,7 @@ except NameError:
 from .exceptions import HTTPError
 
 
-MAX_RETRIES = 4
+DEFAULT_MAX_RETRIES = 4
 
 
 class HTTPSession(object):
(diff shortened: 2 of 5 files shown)


ddoskind:
@burnash @msuozzo Any chance you could have a look at this? Currently we have to use our own branch as we're seeing `500` errors constantly.

rai2048iiit:
@ddoskind I also faced this issue. Exact error message was "500: We are sorry, a server error occurred. Please wait a bit and try reloading your spreadsheet."  but recently I'm not getting these errors. But we should have this fix in master. @burnash @msuozzo

ddoskind:
@rai2048iiit Same here, the situation with Google Spreadsheets returning 500 error codes has become much better again. Would still be nice if this could be merged.

bryandunbar:
Agree, the 500's seem widespread.

burnash:
@ddoskind I'm sorry for the delay. I've bit more time and ready for review. Is the issue still persist? What are the steps to test it?

ddoskind:
@burnash The server errors are more sporadic now, but every now and then we see one:
````
gspread.exceptions.HTTPError: 500: <!DOCTYPE html><html lang="en" >
[... lots of HTML ...]
Google Docs encountered an error. Please try reloading this page, or coming back to it in a few minutes.
[... lots of HTML ...]
</html>
````
Unfortunately the only way to test except from mocking the HTTP call would be to keep on accessing a spreadsheet repeatedly until you run into a 500 error.
I'll try to update the PR with a test that mocks the response by the end of the week and rebase on master.

**Comment by burnash** [comment]
@ddoskind thanks! Look forward to the update.

**Comment by ddoskind** [comment]
Didn't update the mock tests, sorry. Will add another commit.

**Comment by ddoskind** [comment]
It's a bit of a hack, the test for the retry logic now only runs for the full tests, not the mock ones. It would probably be possible to run it for both, but I'm not sure if it's worth the effort.

**Comment by ddoskind** [comment]
@burnash Let me know if this makes sense to you.

**Comment by burnash** [comment]
@ddoskind Looks good to me. I think it does make sense to allow the user to set `MAX_RETRIES` somehow. Need to think about it. 

I also noticed similar functionality in `requests`: https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L88, although this is not related to 50x errors.

**Comment by ddoskind** [comment]
@burnash This commit makes the number of retries configurable, it's still set to 4 by default though. If you'd prefer the default to be not to retry at all, let me know and I'll change.

**Comment by ddoskind** [comment]
@burnash Bump :)

**Comment by burnash** [comment]
@ddoskind Thank you again for the contribution. I'm really sorry for such a late reply. API v3 has been deprecated for quite a while. I have put all of the time I had for gspread into porting it to API v4.

I'm afraid the changes in this PR are not relevant anymore. 

Since #536 you can pass a custom [`requests.Session`](http://docs.python-requests.org/en/master/api/#request-sessions) object to `Client`. Retry is already implemented in requests' default [`HTTPAdapter`](http://docs.python-requests.org/en/master/api/#requests.adapters.HTTPAdapter) and can be customized with urllib3's [`Retry`](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.retry.Retry). Here's an example I adapted from https://www.peterbe.com/plog/best-practice-with-retries-with-requests

```python
import requests

from requests.adapters import HTTPAdapter
from requests.packages.urllib3.util.retry import Retry

retries=3
backoff_factor = 0.3
status_forcelist = (500, 502, 504)

session = requests.Session()
retry = Retry(
    total=retries,
    read=retries,
    connect=retries,
    backoff_factor=backoff_factor,
    status_forcelist=status_forcelist,
)
adapter = HTTPAdapter(max_retries=retry)
session.mount('https://', adapter)

# load `credentials`

gc = gspread.Client(credentials, session=session)
...
```

Similar to #377 I think this functionality is out of scope for gspread and should be addressed at the lower level. I'm closing this pull request.

ddoskind:
@burnash Don't worry! I appreciate your taking the time to explain and understand that there's only so much effort you can spend on open source projects 👍

PR #372, opened 2016-03-21
Added mechanism to write pandas dataframe to google worksheet

sebastianvidal:
I was using gspread to write small-ish pandas dataframes to google sheets and felt it could be a nice feature for folks used to working with dataframes. 

import_dataframe resizes the worksheet to the dimensions of the dataframe it is exporting, replacing any data existing in the worksheet. It will process up to 25,000 rows in its current form - I thought this was a good enough size that it would cover most applications. If there is demand for larger dataframes I'm happy to add some chunking logic (taking into account a worksheet's 2 million cell limit).

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -18,7 +18,7 @@ from xml.etree.ElementTree import Element, SubElement
 from . import urlencode
 from .ns import _ns, _ns1, ATOM_NS, BATCH_NS, SPREADSHEET_NS
 from .urls import construct_url
-from .utils import finditem, numericise_all
+from .utils import finditem, numericise_all, number_to_column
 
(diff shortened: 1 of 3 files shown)


maybelinot:
Hi, @sebastianvidal 
I solved this problem, and it was much easier for me to keep it out of gspread package..
Take a look on [df2gspread](https://github.com/maybelinot/df2gspread), which is based on gspread and has few extra features

burnash:
Thanks for the PR. Although the feature can be useful I'm not sure if it fits the concept of gspread. I suggest to share the code either in [gspread's cookbook](https://github.com/burnash/gspread/wiki/Cookbook), create an external library for this feature or (probably the best way) to collaborate with @maybelinot on df2gspread.

PR #377, opened 2016-04-24
added timeout option

HoonJin:
I have experience connection error while using gspread. 
sometimes, I don't receive any feedback more than 30 minutes. 
so I added 'timeout' option of 'requests' library to gspread. 

if you have any question, send message or comment.

Thank you!

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -224,23 +224,23 @@ class Client(object):
 
         return result
 
-    def get_spreadsheets_feed(self, visibility='private', projection='full'):
+    def get_spreadsheets_feed(self, visibility='private', projection='full', timeout=None):
         url = construct_url('spreadsheets',
                             visibility=visibility, projection=projection)
 
-        r = self.session.get(url)
+        r = self.session.get(url, timeout=timeout)
         return ElementTree.fromstring(r.content)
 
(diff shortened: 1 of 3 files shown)


burnash:
Thanks for the PR. Could you please share a test case for this? (If it is not feasible to make unit test, you can just describe how-to steps in a comment)

burnash:
I'm closing this issue due to inactivity. I also think that potentially cleaner way to manage timeouts can be accomplished by using [`requests.adapters`](http://docs.python-requests.org/en/master/api/#requests.adapters.HTTPAdapter):

Here's a sample code by @Lukasa from a [relevant issue](https://github.com/requests/requests/issues/3341) in the requests repo:

```python
class MyHTTPAdapter(requests.adapters.HTTPAdapter):
    def send(self, *args, **kwargs):
        kwargs['timeout'] = 10
        return super(MyHTTPAdapter, self).send(*args, **kwargs)
```
https://github.com/requests/requests/issues/3341#issuecomment-226764983 

You can then use this adapter and mount it to the custom session object.

```python
session = requests.Session()
session.mount('https://',  MyHTTPAdapter())
```

Then you can pass this session object to gspread's `Client`

PR #385, opened 2016-06-01
Updated update_cells() method to better batch cells

spenceral:
It creates update feeds of 50000 or fewer cells so that the connection does not crash on large updates.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -472,8 +472,12 @@ class Worksheet(object):
         :param cell_list: List of a :class:`Cell` objects to update.
 
         """
-        feed = self._create_update_feed(cell_list)
-        self.client.post_cells(self, ElementTree.tostring(feed))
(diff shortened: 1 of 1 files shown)


burnash:
Thanks for the PR. Could you please share a test case for this? (If it is not feasible to make unit test, you can just describe steps in a comment)

aiguofer:
ohh this would be nice, I had to do something similar [in my gspread-pandas wrapper](https://github.com/aiguofer/gspread-pandas/blob/master/gspread_pandas/client.py#L320)

looking at the commit, wouldn't `xrange` be problematic for python 3?

burnash:
Thank you again for the pull request. I've ported the library to Sheets API v4 recently and, unfortunately, this PR is no longer compatible with the new API. I'm afraid I have to close it.

PR #404, opened 2016-07-30
Improve regex pattern for cell_addr

maybelinot:
'A0' cell label should not work with `get_int_addr()`
@burnash, can you review please?


burnash:
Looks good to me. Thank you!

Issue #421, opened 2016-11-22
Set worksheet contents using Pandas DataFrame


robin900:
For my own project work, I've written a function that sets the contents of a worksheet using a Pandas DataFrame. While similar in behavior to the feature offered by the `pgsheets` package, it has additional features, especially the automatic chunking of DataFrame content to avoid the "API call limit" on large Google Sheet writes.

May I offer a pull request to `gspread` to add this functionality?

burnash:
Thanks for the suggestion and for the PR. I do use Pandas myself although not in conjunction with gspread.

We try to stick to the single responsibility principle in developing this library. At its best, it should only contain functionality related to accessing data and managing spreadsheets. Although widely accepted and used, Pandas is still not a core Python package. I don't see a reason why gspread should support DataFrames. 

That being said, I really don't want to discourage you. You can make your own package, import gspread, inherit and extend the functionality you need. You will own the DataFrames functionality and maintainers of gspread will have less code to support.

I also plan to rework the docs, and maybe put an examples / cookbook section. We have the ancient [GitHub Wiki pages](https://github.com/burnash/gspread/wiki/) for the project with a Cookbook section listing an example of handling DataFrames. We can update example and a link to your future library if it does better handling of DataFrames.

robin900:
If I make this small, small package to allow gspread<->dataframe work, may I call it `gspread-dataframe`?

burnash:
Looks good to me.

robin900:
An alternative would be to include this in gspread, but as an optional subpackage install, so that people could put

`gspread[dataframe]`

in their setup.py requirements or requirements.txt files to have the dataframe module installed.

I can also change the dataframe test to use a mock DataFrame and thus provide test coverage without needing the Travis CI host to have pandas installed.

burnash:
I never used subpackages before and I need some time to study this approach. I do aware of this syntax however and was planning to use it to provide a simple way to substitute `xml.etree` with `lxml`.

robin900:
Their official name is "extras", and the implementation pattern is that an "extra" is in fact a list of additional dependencies needed to use the extra; the "extra" implementation code is always included in the package installation. In setup.py [one specifies these dependencies via `extras_requires`](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies). For example, for the dataframe extra, we could add

`extras_require={ 'dataframe': ['pandas>=0.14.0'] },`

[...]

PR #422, opened 2016-11-23
Add graceful, optional support for Pandas DataFrames

robin900:
Adds gspread.dataframe module with functions `get_as_dataframe` and `set_with_dataframe`.

Gracefully handle absence of pandas with sufficient version; add tests with dataframe equality assertion. add batch support for very large dataframes.

Implements feature requested in #421.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,13 @@ set_with_dataframe(worksheet, df)
 df2 = get_as_dataframe(worksheet)
 ```
 
+You can install Pandas as part of the gspread installation using the `pandas`
+extra:
+
+```sh
+pip install gspread[pandas]
+```
+
 ## Installation
 
 ### Requirements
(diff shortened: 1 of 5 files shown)


burnash:
I outlined my opinion in #421, however, I think that chunking the huge content could be very very useful. Do you consider generalizing your code to support chunking for Python list of lists (or an iterator of lists) instead of DataFrame?

NTAWolf:
@robin900, this looks great! It is much better than what I hacked together for the same purpose :) Really nice!
@burnash , any chance this will be pulled soon?

robin900:
@NTAWolf I've made the code available as a PyPI package: https://pypi.python.org/pypi/gspread-dataframe

Version 1.1.0 is the latest. Github is at

https://github.com/robin900/gspread-dataframe

```
pip install gspread-dataframe
```

Issue #415, opened 2016-10-18
Make get_all_records return empty cells as None


valericus:
Class `gspread.Worksheet` has nice method `get_all_records`. That method allows replace empty cells with zeros (parameter empty2zero) and I assume that would be great to teach it do same thing with `None`. Cause sometimes zero is not the same that nothing :-)

burnash:
Good point. I think a better way would be for this method to provide something like `default_blank` parameter which defaults to `None` and could be set to zero or any other value.

valericus:
Added pull request #423

PR #423, opened 2016-11-23
Make get_all_records return empty cells as None

valericus:
Since now `gspread.Worksheet.get_all_records` method can handle parameter default_blank that will be used to convert blank cells to something except empty string (default behavior) or zero (if `empty2zero=True`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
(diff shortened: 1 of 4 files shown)


valericus:
Pull request for issue #415.

burnash:
Looks good. Thanks for the contribution. Now it's time to deprecate `empty2zero`.

PR #431, opened 2016-12-20
Fix error when inserting permissions.

mjdudak:
There was an error that the request to insert permissions defaulted to form type, I changed it to be 'application/json' which is accepted by Google.


burnash:
Thank you!

PR #439, opened 2017-02-25
Added gid property for worksheets

laike9m:
I need to use this feature in my work and found there's no api to get gid. Luckily there are solutions on [SO](http://stackoverflow.com/questions/11290337/how-to-convert-google-spreadsheets-worksheet-string-id-to-integer-index-gid). I tested it and it works perfect!

Not sure if you're interested in adding this feature. If you do, I'll add some test cases.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -95,6 +95,14 @@ class UtilsTest(unittest.TestCase):
                 (r, c) = utils.a1_to_rowcol(addr)
                 self.assertEqual((row, col), (r, c))
 
+    def test_get_gid(self):
+        gid = 'od6'
+        self.assertEqual(utils.wid_to_gid(gid), '0')
+        gid = 'osyqnsz'
+        self.assertEqual(utils.wid_to_gid(gid), '1751403737')
+        gid = 'ogsrar0'
+        self.assertEqual(utils.wid_to_gid(gid), '1015761654')
+
 
 class GspreadTest(unittest.TestCase):
 
@@ -521,14 +529,6 @@ class WorksheetTest(GspreadTest):
                            for line in exported_data.splitlines()]
 
         self.assertEqual(exported_values, value_list)
-
-    def test_get_gid(self):
-        self.sheet._id = 'od6'
-        self.assertEqual(self.sheet.gid, '0')
-        self.sheet._id = 'osyqnsz'
-        self.assertEqual(self.sheet.gid, '1751403737')
-        self.sheet._id = 'ogsrar0'
-        self.assertEqual(self.sheet.gid, '1015761654')
 
 
 class WorksheetDeleteTest(GspreadTest):
--- a/tests/test.py
+++ b/tests/test.py
@@ -524,11 +524,11 @@ class WorksheetTest(GspreadTest):
 
     def test_get_gid(self):
         self.sheet._id = 'od6'
-        self.assertEqual(self.gid, '0')
+        self.assertEqual(self.sheet.gid, '0')
         self.sheet._id = 'osyqnsz'
-        self.assertEqual(self.gid, '1751403737')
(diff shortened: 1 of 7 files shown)


burnash:
Thanks for the PR.

What is the gid property used for in your case? If it's not a secret and you can share this information, it could be useful to other people searching for this functionality.

laike9m:
That's not a secret, basically I need to create links for each worksheet, so that when people clicks a link it brings people to a certain worksheet. And as you know, `#gid=xxx` is the last part of a worksheet's url. To build those urls, I need to retrieve gid for each worksheet.

laike9m:
What do you think?

burnash:
Great thank you. Could you add test cases? & I'll merge the PR right away.

laike9m:
Added tests based on a newly created sheet:
https://docs.google.com/spreadsheets/d/1tycoxErSg0JP7fCG3GlOgMbLtFjCR8U0PBg5NCrfVow/edit#gid=1015761654

burnash:
@laike9m You put it in a right place. Worksheet is not in mock tests yet.

tests/test.py
```diff
@@ -521,6 +522,14 @@ def test_export(self):
 
         self.assertEqual(exported_values, value_list)
 
+    def test_get_gid(self):
+        self.sheet._id = 'od6'
+        self.assertEqual(self.gid, '0')
```
burnash: `self` refers to `WorksheetTest` instance, so this results in `AttributeError: 'WorksheetTest' object has no attribute 'gid'`.

laike9m:
My fault, fixed it.

laike9m:
Anything else I should do to get it merged?

tests/test.py
```diff
@@ -521,6 +522,14 @@ def test_export(self):
 
         self.assertEqual(exported_values, value_list)
 
+    def test_get_gid(self):
+        self.sheet._id = 'od6'
+        self.assertEqual(self.sheet.gid, '0')
+        self.sheet._id = 'osyqnsz'
+        self.assertEqual(self.sheet.gid, '1751403737')
+        self.sheet._id = 'ogsrar0'
```
burnash: I'm getting  '400: Invalid query argument for grid_id.' error when I run the tests:
~~~
  File "~/b/p/gspread/tests/test.py", line 203, in tearDown
    self.spreadsheet.del_worksheet(self.sheet)
  File "~/b/p/gspread/gspread/models.py", line 156, in del_worksheet
    self.client.del_worksheet(worksheet)
  File "~/b/p/gspread/gspread/client.py", line 203, in del_worksheet
    self.session.delete(url)
  File "~/b/p/gspread/gspread/httpsession.py", line 76, in delete
    return self.request('DELETE', url, params=params, **kwargs)
  File "~/b/p/gspread/gspread/httpsession.py", line 69, in request
    response.status_code, response.content))
RequestError: (400, '400: Invalid query argument for grid_id.')
~~~

My guess is that this happens because of setting `self.sheet._id` to a value different from what it was before the test started. It could work for you since you're using a different document for testing. I see two possible ways on how to make the test document independent:

1. to extract the gid calculation function and move it to [utils](https://github.com/burnash/gspread/blob/4ad90b4523214b14e61ef889098603fa5d449c49/gspread/utils.py) module (since the only value the method is accessing now is `_id`). Since this function will be pure it can be tested without a spreadsheet document, as [other utils functions](https://github.com/burnash/gspread/blob/4ad90b4523214b14e61ef889098603fa5d449c49/tests/test.py#L51). 

2. to store `self.sheet._id` value before the test code and reset the value after. But that's an inferior approach, imho.

laike9m:
Agreed, this is much easier to test.

burnash:
Works like a charm, thank you, @laike9m

PR #454, opened 2017-04-18
get_all_values() detects empty rows and columns at end of worksheet

### LINKED ISSUE #453: get_all_values() misses empty rows at bottom of worksheet
### Environment info

Operating System: Ubuntu 14.04 LTS
Python version: 2.7.12
gspread version: v0.6.2

### Steps to reproduce
1. Create a worksheet that ends with N empty rows (arbitrarily many, as long as they're the bottom rows in the sheet)

e.g., the worksheet would correspond to an array like the following:

```python
[["A", "B", "C", "D"],
 ["E", "F", "G", "H"],
 ["", "", "", ""]]
```

For this worksheet, the actual row count would be 3 but get_all_values() would only return the top 2 rows.

### Stack trace or other output that would be helpful

Assuming wks is a Worksheet object corresponding to the worksheet created in Step 1,
```python
len(wks.get_all_values())
```

will not equal
```python
wks.row_count
```

and get_all_values will miss the last N rows (the empty ones) in the sheet.

If the bottom row of the worksheet is non-empty, (as far as I know) this issue does not occur. It's only when the bottom row is purely empty. The row_values method still functions fine in this case.

I can't tell from the get_all_values() documentation whether this is intentional.

knyte:
Resolves #453 

Instead of inferring the worksheet size from the non-empty cells, get_all_values() directly checks the row and column counts so it doesn't miss any purely empty rows or columns that aren't followed by non-empty rows or columns.

I haven't been able to determine from the documentation whether the get_all_values() method intentionally ignores empty edge rows/columns. Having it intentionally ignore those would make sense given that Google Sheets worksheets are default sized at 1000x26, although I've got a use case for which this is a bug*.

If the exclusion is intentional, it might be useful to note that in the documentation so users can process the output of get_all_values() themselves or add an optional parameter to get_all_values() to allow users to decide whether they want those empty edge rows/columns.

\* use case: I'm using gspread to pull, process, and modify some data that had been entered into a somewhat large set of worksheets in the past where empty cells are significant and some rows consist entirely of empty cells.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -365,7 +365,6 @@ class WorksheetTest(GspreadTest):
         self.assertEqual(list_len, len(result_list))
 
     def test_get_all_values(self):
-        print self.sheet
         self.sheet.resize(7, 5)
         # put in new values, made from three lists
         rows = [["A1", "B1", "", "D1", ""],


burnash:
@knyte Thank you for your contribution. I'm really sorry for such a late reply. API v3 has been deprecated for quite a while. I have put all of the time I had for gspread into porting it to API v4.

I'm afraid the changes in this PR are not relevant anymore. 

As you have correctly pointed out the omission of empty rows and columns is intentional. In fact, this is how Sheets API returns it in API v4 and (if I'm not mistaken in v3).

I've updated `get_all_values()` docs with a note on empty rows and columns exclusion: c58d0e4fc0686f74939200936f3c69f55d120387

I'm closing this pull request.

PR #459, opened 2017-05-10
Added ability to copy a spreadsheet

wtluke:
This requires an additional scope, https://www.googleapis.com/auth/drive:
`scope = ['https://spreadsheets.google.com/feeds','https://www.googleapis.com/auth/drive']`

The existing `create()` function works as normal. This includes two new parameters, `copy` and `key`. Key is the key or file ID of the original spreadsheet (obtained from the URL of the Google Spreadsheet you want to copy from) Example:

`gc = gspread.authorize(creds)`

`n=gc.create(copy='yes',key='key/fileID',title='New spreadsheet (copy)')`

`n.share('you@gmail', perm_type='user', role='writer')`

Thoughts/improvements welcome. Would be nice to have an additional "copy title" option to grab the existing title from the originating file, and append "(copy)" on the new one. Will do that when I get time.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -242,7 +242,7 @@ class Client(object):
 
         return ElementTree.fromstring(r.content)
 
-    def create(self, title):
+    def create(self,title,copy='no',key=''):
         """Creates a new spreadsheet.
 
         :param title: A title of a new spreadsheet.
@@ -265,6 +265,9 @@ class Client(object):
            when you try to create a new spreadsheet.
 
         """
+        endpoint=DRIVE_FILES_API_V2_URL
+        if copy =='yes':
+            endpoint="https://www.googleapis.com/drive/v2/files/" + key + '/copy'        
 
         headers = {'Content-Type': 'application/json'}
         data = {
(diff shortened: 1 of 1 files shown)


Drachenfels:
What I personally think instead of extra argument copy on creation, maybe you should go for gc.copy() where yo specify fileId of what you want to copy.

robsgithub:
Any chance this can get pushed through?

burnash:
@robsgithub absolutely. This is a good feature. Please stay tuned.

arudmin:
So when it will be availible?

Grism:
this does not appear to be available yet, will it be?

scythargon:
+1

scythargon:
@burnash Hello! Can you please merge this?

burnash:
@wtluke Sorry for the delay, it was a busy year. Thank you for the contribution.

I like the suggested feature. But I think the implementation could be improved. I agree with @Drachenfels `gc.copy` could be a better fit for this feature (Separation of concerns + in line with underlying Drive API design).

My suggestion is a method with a following usage:

```
copy_of_spreadsheet = gc.copy(
    file_id='<spreadsheet_id>', 
    title='A copy of the spreadsheet'
)
```

I understand that it's been a year and a half since you submitted the PR. Would you still be interested in reworking the PR?

dsask:
@burnash I have implemented something very similar to this, I just submitted is as a separate Pull Request (#625)

burnash:
I'm closing this PR. The feature is implemented in #625. It is available in gspread version 3.1.0

PR #479, opened 2017-08-25
Added a clarification about which API to auth

avivo:
I just ran into this, and figured it was worth sharing. This is my first contribution here, let me know if I missed something.


burnash:
Sorry for the late response. There were many changes to the API console over these years and this PR is unfortunately no longer relevant. @manasouza is correct, right now we need to enable Sheets API.
I'm closing this PR.

PR #493, opened 2017-11-02
allow setting the withLink property of permissions

### LINKED ISSUE #492: Set permissions 'withLink' property?
Is it possible to set the 'withLink' property of new permissions? This is allowed in the DRIVE api insert [here](https://developers.google.com/drive/v2/reference/permissions/insert). This seems handy when making files editable by anyone (which might be inadvisable in general, but is helpful for working with random groups of people)

Just hacked up client.py like this, maybe there's a better way?

    def insert_permission(
        self,
        file_id,
        value,
        perm_type,
        role,
        notify=True,
        email_message=None,
        withLink=False
    ):

jsfenfen:
Closes #492. No tests or anything, but seems to work.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -371,7 +371,7 @@ class Client(object):
             'value': value,
             'type': perm_type,
             'role': role,
-            'with_link': with_link
+            'withLink': with_link
         }
 
         params = {
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -317,7 +317,7 @@ class Client(object):
         role,
         notify=True,
         email_message=None,
-        withLink=False
+        with_link=False
     ):
         """Creates a new permission for a file.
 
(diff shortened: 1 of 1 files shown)


gspread/client.py
```diff
@@ -316,7 +316,8 @@ def insert_permission(
         perm_type,
         role,
         notify=True,
-        email_message=None
+        email_message=None,
+        withLink=False
```
burnash: Can you use `lower_case_with_underscores` style for the new argument name? So it's in line with other arguments' names?

gspread/client.py
```diff
@@ -360,6 +371,7 @@ def insert_permission(
             'value': value,
             'type': perm_type,
             'role': role,
+            'withLink':withLink,
```
burnash: There's a missing space between a colon and a variable name.

burnash:
Thank you for the PR. I've reviewed it. Please read the comments above.

jsfenfen:
Thanks for the review @burnash. I've fixed the style problems you mentioned in 
2364b3d.

burnash:
Great! Thanks again for the contribution!

PR #496, opened 2017-12-21
Decode request data as utf-8 by default

drbig:
Found a problem when trying to upload data containing unicode, and this is a working solution.

**Now**, I can't say I did any sort of proper testing (only Python 2.7.11, no more cases than my one), though as later in the code there is explicitly `data.encode('utf-8')` this doesn't strike me as an immediately stinky solution.


burnash:
Thanks for the contribution and sorry for the pause in replying.

PR #502, opened 2018-02-03
Fix default behaviour of numericise function

### LINKED ISSUE #499: Ambiguity in the `numericise` function depending on Python version.
https://github.com/burnash/gspread/blob/217b43073e1abe9ceb1d65d2d0719d56f0a14642/gspread/utils.py#L58

I have run into some issues when fetching data from spreadsheets using the `get_all_records()` method when using Python 3.6+.

If a worksheet has data in the format `12_34`, which should be interpreted as a string (or numeric if _explicitly_ desired) the `numericise` function will convert it to an `int` because of https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep515 introduced in Python3.6, using Python3.5.4 it parses it as the string `12_34`, as expected.

I don't think following PEP515 should be the default behaviour in this case.

Example

Python3.6
```
>>> from gspread.utils import numericise
>>> numericise('18_29')
1829
```

Python3.5.4
```
>>> from gspread.utils import numericise
>>> numericise('18_29')
'18_29'
```

danthelion:
Fixes #499


danthelion:
@burnash What do you think about this implementation? Details are in the referenced issue.

burnash:
@danthelion thank you for the issue and the PR.

As I said in the referenced issue I agree. I'm sorry for the delay in responding. Since you submitted the PR the code of gspread has significantly changed. If you have time, could you rebase your changes to resolve the conflicts? Also please don't forget to add a test to cover this fix.

danthelion:
Sure, I'll get on it as soon as I can. 👍

danthelion:
@burnash Rebased and added some unittests, let me know if theres more needed to be done.

danthelion:
Is there a chance of this change being merged @burnash?

burnash:
Thanks for the contribution and sorry for the delay. I've merged the PR. 
Note: tests pass locally but still fail on Travis due to an unrelated issue.

PR #506, opened 2018-02-27
Added update_worksheet_format in order to format a worksheet

elinnore:
Hi @burnash,

Thanks a lot for implementing gspread. I found it really easy to use.
I am currently working with it and I needed to add a method to apply formatting to a worksheet at some point.
I have added it in my local package and I thought to submit it to you, perhaps it helps.
Then I noticed you already have a new branch for api v4 and my modifications are based on the current master.
I haven’t had the time to check the apiv4 branch properly, but if you think my modifications are fine I can fix them to fit in this branch.
Also, I am new to github so I hope I don’t cause any troubles here :)

Anyway I have performed the following:
The formatting is contained in a json file.
I have modified urls.py file to include the URL used in APIv4.
I added update_format method in client.py that performs the POST containing the json as dictionary on the APIv4 url.
I added update_worksheet_format method in models, under worksheet class, that will open the json, modify the worksheet id and pass it as dictionary to the update_format method. This method receives as param the json file name/path.
These modifications appear to work fine for me, even though I am not 100% sure whether the transformations on the json are really reliable :)
Update_format just depends on the spreadsheet id so it can be used to add a method to update the format of the entire spreadsheet.

Regards,
Elinnore

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -390,6 +390,19 @@ class Client(object):
         headers = {'Content-Type': 'application/json'}
 
         self.session.delete(url, headers=headers)
+        
+    def update_format(self, file_id, data_file):
+        """Updates format of a worksheet.
+
+                :param file_id: a spreadsheet ID (aka file ID.)
+                :param data_file: json with formatting.
+                """
+        headers = {'Content-Type': 'application/json'}
+        with open(data_file) as json_file:
+            data = json.load(json_file)
+        url = construct_url('cells_batch_format', spreadsheet_id=file_id)
+        r = self.session.post(url, json=data, headers=headers)    
+        
 
 
 def authorize(credentials):
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -395,13 +395,11 @@ class Client(object):
         """Updates format of a worksheet.
 
                 :param file_id: a spreadsheet ID (aka file ID.)
-                :param data_file: json with formatting.
+                :param data: dictionary created from json containing formatting details.
                 """
         headers = {'Content-Type': 'application/json'}
-        with open(data_file) as json_file:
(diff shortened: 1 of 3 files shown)


gspread/client.py
```diff
@@ -390,6 +390,17 @@ def remove_permission(self, file_id, permission_id):
         headers = {'Content-Type': 'application/json'}
 
         self.session.delete(url, headers=headers)
+        
+    def update_format(self, file_id, data_file):
```
TelmoM: change 'data_file' to 'data'

burnash: @TelmoM thanks for weighing in. However, I think it would be better if you can provide a more elaborated comment on the suggested change.

elinnore: Hi @TelmoM ,
yes, you are right. The argument in the update_format method should be named "data" instead of "data_file". The POST in the end uses "data" as var name for the content of the json. Sorry about this.

elinnore: Hi @TelmoM ,
yes, you are right. The argument in the update_format method should be named "data" instead of "data_file". The POST in the end uses "data" as var name for the content of the json. Sorry about this.

burnash:
Hello @elinnore, I'm really sorry for such a late response. 

Unfortunately, I have to close this PR. It's really hard to deny PRs from first-time contributors. 
I try to be very careful with extending gspread API with new methods and I have been reluctant in adding formatting support at all. Recently, I changed my mind and decided to add a lightweight method [`format()`](https://github.com/burnash/gspread/blob/1e6882c36d1026a93192762d4fc4bb831e57d863/gspread/models.py#L1108) to support formatting.

This PR, unfortunately, is out of date. Again sorry for the late response and thank you for your contribution.

PR #522, opened 2018-03-23
Fetch all spreadsheets in Spreadsheet.list_spreadsheet_files

aiguofer:
The current implementation only fetches 100 sheets. I have significantly more than that so I ran into this issue. This commit should allow to fetch all of the sheets for a given user.

It's possible that in the future someone will want to also be able to leverage the `corpora` param which does:
```
Comma-separated list of bodies of items (files/documents) to which the query applies. Supported bodies are 'user', 'domain', 'teamDrive' and 'allTeamDrives'. 'allTeamDrives' must be combined with 'user'; all other values must be used in isolation. Prefer 'user' or 'teamDrive' to 'allTeamDrives' for efficiency.
```

For now I left it out so it's using the default

### DIFF
--- a/gspread/v4/client.py
+++ b/gspread/v4/client.py
@@ -73,12 +73,23 @@ class Client(BaseClient):
             raise APIError(response)
 
     def list_spreadsheet_files(self):
-        url = (
-            "https://www.googleapis.com/drive/v3/files"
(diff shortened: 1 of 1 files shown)


burnash:
@aiguofer Thank you for the contribution. I agree that the method could be extended even further in future to leverage `corpora` and other team drive related features.

PR #523, opened 2018-03-24
Raise SpreadsheetNotFound in Client.open_by_key if not found

aiguofer:
I catch this exception on https://github.com/aiguofer/gspread-pandas to let users pass in Title, URL, or ID and cascade through methods until it succeeds.

### DIFF
--- a/gspread/v4/client.py
+++ b/gspread/v4/client.py
@@ -122,12 +122,17 @@ class Client(BaseClient):
 
         """
         try:
-            return Spreadsheet(self, {'id': key})
-        except APIError as err:
-            if 'NOT_FOUND' in err:
-                raise SpreadsheetNotFound
-            else:
-                raise err
+            properties = finditem(
+                lambda x: x['id'] == key,
+                self.list_spreadsheet_files()
+            )
+
+            # Drive uses different terminology
+            properties['title'] = properties['name']
+
+            return Spreadsheet(self, properties)
+        except StopIteration:
+            raise SpreadsheetNotFound
 
     def openall(self, title=None):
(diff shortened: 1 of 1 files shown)


aiguofer:
Hmm, nvm this doesn't quite work because instantiating a `Spreadsheet` object doesn't actually call out to the API.

I see two alternatives:

1. Try to call something that hits the API in the `Spreadsheet` constructor
2. Do a check similar to `Client.open` but looking at the `id` instead

What do you think?

burnash:
I'm still figuring out a proper API for gspread that will be aligned with Sheets API v4. In this case, while porting the library to v4 I deliberately omit the fetching of a list of the user's spreadsheets from `Client.open_by_key`. I believe it's a redundant call and an expensive way of checking if the key is correct.

It's a tradeoff. You don't make unnecessary API call if you have the correct spreadsheet ID (aka _the key_) but if something's wrong with the id you'll get an API error down the line anyway when trying to fetch the data.

Again the API is in flux right now. However, if there's a need to check a spreadsheet id provided by the user, calling [`Spreadsheet.fetch_sheet_metadata()`](https://github.com/burnash/gspread/blob/d32398495ff335b725aac30f14be0528505cf0b8/gspread/v4/models.py#L115) could be a more performant alternative to `Client.list_spreadsheet_files()` — especially if the user has a lot of spreadsheets.

aiguofer:
Ahh yeah that makes sense. I've since changed my code to do just that. I actually also realized this would break the use-case of opening public spreadsheets that are not saved in your own drive. 

I'll close this PR since there's likely better ways to do this and I've solved it in my code for now.

PR #527, opened 2018-04-02
Open team drive sheets by name

ryantuck:
Currently, I can open a spreadsheet located in a team drive by key or url, but not by name. This issue occurs due to the `client.list_spreadsheet_files()` function not having the ability to list team drive files by default.

This PR fixes that issue by enabling team drive files to be listed by default. The ability to list existing spreadsheets outside of team drives is unaffected by this change.

More documentation here: https://developers.google.com/drive/v3/web/enable-teamdrives

### DIFF
--- a/gspread/v4/client.py
+++ b/gspread/v4/client.py
@@ -84,7 +84,9 @@ class Client(BaseClient):
         url = "https://www.googleapis.com/drive/v3/files"
         params = {
             'q': "mimeType='application/vnd.google-apps.spreadsheet'",
-            "pageSize": 1000
+            "pageSize": 1000,
+            'supportsTeamDrives': True,
+            'includeTeamDriveItems': True,
         }
 
         while page_token is not None:


burnash:
@ryantuck Thanks for the PR. I haven't worked with Team Drive yet. I wonder what's the downside of enabling Team Drive access by default?

ryantuck:
@burnash it's not apparent that there is any downside, especially given that `open_by_key()` and `open_by_url()` are *already* able to access sheets within team drives. As a result, my take is that this PR is fixing a bug/inconsistency instead of adding new functionality.

A particular sheet, even within a team drive, still needs to be shared with whatever account you are passing into the `gspread.authorize()` function in order to be returned by the drive API's `files.list` call.

burnash:
@ryantuck thank you for the elaboration. Looks like TeamDrive can be enabled by default.

PR #530, opened 2018-04-04
URL encode the range in the value_* functions

aiguofer:
This fixes using this endpoing when the worksheet name has forward
slashes or other strange characters.


burnash:
That's a great finding and fix. Thank you!

PR #542, opened 2018-04-26
Fix Worksheet.title being out of sync after using update_title

ryanpineo:


burnash:
Thank you for the contribution!

Issue #545, opened 2018-05-04
Adding new rows doesn't increase row count property


cgmorton:
### Environment info

Operating System: MacOS Sierra
Python version: 3.6.3 (Miniconda)
gspread version: 3.0.0

### Steps to reproduce
ws = sheet.worksheet('Sheet1')
print(ws.row_count)
ws.add_rows(1)
print(ws.row_count)

uneasyguy:
I too am having this issue if anyone can lend a hand

crash5936:
The issue is actually larger than that - the `Worksheet` instance seems to never update its properties after changing them, the same goes for `self.col_count` and `self.title` properties.

I think the issue could be fixed by updating the `Worksheet` instance properties every time after calling `self.batch_update()`

cgmorton:
You're right!  Let me know if it would help if I modified the title of this issue to better describe the problem.  Or I could close this issue if you want to start a new one.

crash5936:
I tried fixing the issue and opened a pull request, it fixed all the issues for me, let's see if it's viable and we can get this merged into master :)

python-engineer:
> I tried fixing the issue and opened a pull request, it fixed all the issues for me, let's see if it's viable and we can get this merged into master :)

I would love to see this one merged

tlhinman:
Any updates on this? This is a pretty big impediment for us right now.

Acendro:
Can't you just bypass the issue with saving row_count into a variable before appending and then incrementing it everytime you append a row during the code afterwards?
Works fine for me at least

lavigne958:
Hi this is a bigger issue in fact, as none of the properties are up to date after changing them. 

I think the simplest way of solving this for now is to update local properties when an update is successful.
So for instance calling `add_rows` will increase the local `row_count` value. this can lead to mistakes still because someone could be updating the sheet from an other script or from a web browser. but still it solves the issue for single access.

kuznetsov-m:
I have the same problem like issue starter.

alifeee:
Fixed by #1211?

lavigne958:
yes closed by #1211

alifeee:
this specific method is fixed, but there is still the same issue if you use other methods such as `append_rows` or `insert_rows`

see https://github.com/burnash/gspread/issues/881#issuecomment-1591337773

PR #548, opened 2018-05-16
FIX: updating Worksheet properties after calling methods that can change them

crash5936:
Digging into issue #545 I found out, that there were multiple problems caused by this. 
Essentially, when any method that modifies the properties of a sheet online is called, the Worksheet instance properties didn't get updated accordingly.

For example, when you called `gspread.models.Worksheet.add_cols(x)` twice in a row, it only added this number of columns to the initial number of columns (ie. the number of columns from when the Worksheet() class was instantiated)

I tried fixing this by adding a decorator, which causes the function to read the actual metadata from the API and update the instance properties accordingly - testing this, it seems to fix all the issues.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -22,7 +22,8 @@ from .utils import (
     numericise_all,
     finditem,
     fill_gaps,
-    cell_list_to_rect
+    cell_list_to_rect,
+    update_worksheet_properties
 )
 
 from .urls import (
@@ -530,6 +531,7 @@ class Worksheet(object):
         except KeyError:
             return []
 
+    @update_worksheet_properties
     def update_acell(self, label, value):
         """Sets the new value to a cell.
 
@@ -544,6 +546,7 @@ class Worksheet(object):
         """
         return self.update_cell(*(a1_to_rowcol(label)), value=value)
 
+    @update_worksheet_properties
     def update_cell(self, row, col, value):
         """Sets the new value to a cell.
 
@@ -570,6 +573,7 @@ class Worksheet(object):
 
         return data
 
+    @update_worksheet_properties
     def update_cells(self, cell_list, value_input_option='RAW'):
         """Updates cells in batch.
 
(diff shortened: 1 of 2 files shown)


aiguofer:
Nice catch, I encountered an issue caused by the same problem. The only concern with this approach is that it might increase the number of API calls significantly.

aiguofer:
I got to thinking about this and got an idea. Instead of calling fetch_sheet_metadata each time we update a property on a worksheet, we hold the sheets_metadata in the Spreadsheet instance and add a flag to indicate that it needs updating. When accessing the sheets metadata, we can update it if it needs it, and each Worksheet's properties are a reference to the Spreadsheet, so they would be updated accordingly.

I'm not entirely sure how much this would actually reduce the number of calls for most usecases, but it seems like it could save some calls when working with multiple worksheets from the same spreadsheet.

see https://github.com/aiguofer/gspread/commit/21c87dd04848abe28e9d2deee95bdd9ede1c36ad (forked off your branch)

lavigne958:
Hi this is a great idea. and it will prevent some issues for sure.  thank you @crash5936. 

As you mentioned @aiguofer, this will increase the number of calls toward the API. 

We must think this through and find a solution that brings: robustness (the local metadata are up to date) and efficiency (we don't do more HTTP call toward the API than necessary).

I think at some points we won't have a choice, the metadata might change (some user change the column number using a web browser) and we are not up to date.

A solution could be to offer the user a choice:
- Do not fetch metadata every times, rely on what we know locally, the script is the only one updating the spreadSheet
- Get metadata every times after an update to make sure we are up to date, some else might be updating the spreadSheet outside the script (this could benefit from your idea @aiguofer to prevent having too many requests)

alifeee:
fixed by #1211?

lavigne958:
yes closed by #1211 

this was one way to make it happen too but it's less specific. so far we can continue with our current method if we have so many properties to update at some point we'll change it for something like this.

alifeee:
yes. this way also increased the API calls.

PR #552, opened 2018-05-21
Add Authlib usage

lepture:
Added a use case with [Authlib](https://authlib.org/). Here is a blog post of it: https://blog.authlib.org/2018/authlib-for-gspread


burnash:
Thank you very much for the example. I'm sorry for the late reply. 
I've reworked the authentication docs. I think this actually belongs to the "Custom Authentication" section. Later on, I'll move this to its own file.

Issue #538, opened 2018-04-11
update_cells no longer supports non-contiguous cells (overwrites unlisted cells within the range)


CianH:
This appears to be a regression in the gpsread use of the v4 API (was previously working: see #177)

### Environment info
Operating System: Windows / Linux
Python version: 2.7.14
gspread version: 2.1.1

### Steps to reproduce
1. Make a list of cells that are non-contiguous, e.g. [<Cell R7C2 '2018'>, <Cell R20C2 '2004'>]
2. Call update_cells, passing the list
3. Cells R7C2 _through_ R20C2 inclusive get updated, with only R7C2 and R20C2 having values, R8-R19C2 all get zeroed out.

### Stack trace or other output that would be helpful
Compare update_cells implementation in `/v4/models.py` with `models.py`

```
>>> worksheet.range('B7:B20')
[<Cell R7C2 ''>, <Cell R8C2 ''>, <Cell R9C2 u'2013'>, <Cell R10C2 u'2013'>, <Cell R11C2 u'2015'>, <Cell R12C2 u'2015'>, <Cell R13C2 ''>, <Cell R14C2 u'2017'>, <Cell R15C2 u'2015'>, <Cell R16C2 u'2017'>, <Cell R17C2 u'2016'>, <Cell R18C2 u'2014'>, <Cell R19C2 u'2012'>, <Cell R20C2 ''>]
>>> cells = worksheet.range('B7:B20')
>>> other_cells = []
>>> other_cells.append(cells[0])
>>> other_cells.append(cells[len(cells)-1])
>>> other_cells
[<Cell R7C2 ''>, <Cell R20C2 ''>]
>>> other_cells[0].value = "test"
>>> other_cells[1].value = "testing"
>>> len(other_cells)
2
>>> worksheet.update_cells(other_cells)
{u'spreadsheetId': u'**removed**', u'updatedRange': u'Collection!B7:B20', u'updatedCells': 14, u'updatedRows': 14, u'updatedColumns': 1}
>>> worksheet.range('B7:B20')
[...]

PR #553, opened 2018-05-23
Bugfix 538: default cell values to None instead of empty string

ADraginda:
#538 

The google API specifies that if the cell `value` is None then no updates will happen in the sheet. Thus, I swapped out the defaultdicts for regular dictionaries and used `.get` for accessing the values (thereby defaulting to None instead of an empty string).

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -595,8 +595,8 @@ class Worksheet(object):
 
         values_rect = cell_list_to_rect(cell_list)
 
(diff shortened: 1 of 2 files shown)


burnash:
@ADraginda thank you very much for contributing the bugfix. I'm checking it now.

burnash:
All good, thank you again. I've merged the pr. I've also added a test case.

PR #556, opened 2018-06-08
add append_rows function

martinwarby:
Added an append_rows function. This was added to avoid hitting quota limits when adding multiple rows


burnash:
Thank you very much for your contribution.

Issue #560, opened 2018-06-21
Update cell by column name


adace123:
It would be nice to not have to rely on column order and instead reference the column by name.

E.g. `sheet.update_col(col_name, row, value)`

lavigne958:
I don't think this function is very needed. As workaround if really needed one can use the function `a1_range_to_grid_range`.
One can provide only a column name and get its index.

Closing this issue.

PR #566, opened 2018-07-10
Adds method requested in issue #560

MayconGesser:


burnash:
Thank you for the contribution, however, this PR lacks context. Even though this feature has been requested in #560, I'm not convinced it's needed enough to add more methods to the API.

PR #572, opened 2018-07-28
Fix usage of client.openall when a title is passed in

aiguofer:
This allows user to get all spreadsheets that contain (case
insensitive) the string that is passed in.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -170,8 +170,9 @@ class Client(object):
         spreadsheet_files = self.list_spreadsheet_files()
 
         if title:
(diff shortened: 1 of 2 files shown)


ADraginda:
I just ran into this issue as well. I would argue for implementing with exact match instead of `in`.

aiguofer:
Hey all, sorry had forgotten about this. I implemented it as case sensitive `in` search for a bit more flexibility, but I agree it probably makes sense to do an exact match. Will update PR

burnash:
Thank you very much!

PR #574, opened 2018-07-28
Add freeze and add_filter functions to Worksheet

### LINKED ISSUE #543: Feature Request: Freezing Rows/Columns and Other Worksheet Formatting
I was thinking it would be really cool to implement a way to format worksheets using the gspread Worksheet class, like freezing rows/columns.  They could be class methods and calling something like worksheet-instance.freeze_row(n) could freeze the first n rows.

aiguofer:
These allow you to freeze columns and/or rows and to create a basic filter on a set of data.

This would fix #543.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -333,6 +333,16 @@ class Worksheet(object):
         """Number of columns."""
         return self._properties['gridProperties']['columnCount']
 
+    @property
+    def frozen_row_count(self):
+        """Number of frozen rows."""
+        return self._properties['gridProperties'].get('frozenRowCount', 0)
+
+    @property
+    def frozen_col_count(self):
+        """Number of frozen columns."""
+        return self._properties['gridProperties'].get('frozenColumnCount', 0)
+
     def acell(self, label, value_render_option='FORMATTED_VALUE'):
         """Returns an instance of a :class:`gspread.models.Cell`.
 
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -813,6 +813,83 @@ class Worksheet(object):
         """
         return list(self._finder(filter, query))
 
+    def freeze(self, rows=None, cols=None):
+        """Freeze rows and/or columns on the worksheet.
+
+        :param rows: Number of rows to freeze.
+        :param cols: Number of columns to freeze.
+        """
+        grid_properties = {}
+
+        if rows is not None:
+            grid_properties['frozenRowCount'] = rows
+
+        if cols is not None:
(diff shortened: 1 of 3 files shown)


gspread/models.py
```diff
@@ -813,6 +813,96 @@ def findall(self, query):
...
+                    },
+                    'fields': fields
+                }
+            }]
+        }
+
+        return self.spreadsheet.batch_update(body)
+
+    @cast_to_a1_notation
+    def add_basic_filter(self, name=None):
```
robin900: A basic filter also has (optionally) a SortSpec and a collection of FilterCriteria (optionally one criterion for each column included in the range). I think if we want to add basic filter support, we would want to support the full feature set of basic filters.

gspread/models.py
```diff
@@ -813,6 +813,96 @@ def findall(self, query):
         """
         return list(self._finder(filter, query))
 
+    def freeze(self, rows=None, cols=None):
```
robin900: You probably want to include some `@property` methods `frozen_row_count` and `frozen_column_count` so that these values are readable from the Worksheet object.

aiguofer: Fair point. The only issue is that this will be affected by #545 (see #548). Added them anyway for now.

gspread/models.py
aiguofer: Yeah, I thought about those but I figured I can sort and filter the data before pushing it... I find the basic filter most useful for allowing users of the sheet to do their own exploration.

However, that was purely thinking on my usecase... I think it makes sense to add the options here, but I'm not sure how to do it while still leveraging @cast_to_a1_notation. I'll think on it, but I'm thinking it might need some major re-work of @cast_to_a1_notation.

aiguofer: Looking at these, they seem to be 'non-trivial' parameters. 

It's easy to implement their addition with no checking and letting the user build the correct JSON. An alternative is to create an object to represent each of those parameters so the user can build them programmatically and pass in the object. The latter implementation seems 'better' but it seems like too much effort for something that I'm not convinced will be used very often and I don't have the time to implement that.

However, I still think the addition of the basic functionality as proposed by this PR is quite useful to many (I had already implemented this in https://github.com/aiguofer/gspread-pandas and 'back-ported' it here for others to use).

If the simple param option (user passes in the dict) is good enough then I'm happy to implement it, otherwise I think this could be merged for now and the better implementation could be added later.

burnash:
@aiguofer thank you for your contribution
@robin900 thank you for the insights

PR #579, opened 2018-08-07
Out of order Cell list cause update_cells to fail

buzzlaker:
This may be more of an edge case situation, but for me, it is a common situation I run into while using update_cells to keep some business documents in sync with other systems.

This problem is most commonly cause by someone shifting cells in a worksheet.

I compiled a crude code sample which demonstrates this issue below. To take this a step further as a real world example: A sheet would get updated by external sources (as well as manually), with the tabs showing different views/analytics for different departments.

```python
credentials = get_credentials()

main_sheet = gc.open_by_key('sheet-id').worksheet('Main')
# main_sheet_headers = ["Reference Col C", "Reference Col D", "Reference Col A", "Reference Col B"]
main_sheet_data = main_sheet.get_all_values()
main_sheet_headers = []
for row in main_sheet_data[0]:
    main_sheet_headers.append(row)

update_sheet = gc.open_by_key('sheet-id').worksheet('Update Here')
# update_sheet_headers = ["Reference Col D", "Reference Col A", "Reference Col B", "Reference Col C"]
update_sheet_data = update_sheet.get_all_values()
update_sheet_headers = []
for row in update_sheet_data[0]:
    update_sheet_headers.append(row)

updated_data = {}

for row, col in enumerate(main_sheet_data):
    updated_data[row] = {
        'Reference Col A': '=Main!' + gspread.utils.rowcol_to_a1(row + 2, main_sheet_headers.index('Reference Col A') + 1),
        'Reference Col B': '=Main!' + gspread.utils.rowcol_to_a1(row + 2, main_sheet_headers.index('Reference Col B') + 1),
        'Reference Col C': '=Main!' + gspread.utils.rowcol_to_a1(row + 2, main_sheet_headers.index('Reference Col C') + 1),
        'Reference Col D': '=Main!' + gspread.utils.rowcol_to_a1(row + 2, main_sheet_headers.index('Reference Col D') + 1),
        }

updates = []
for row in updated_data:
    for col in updated_data[row]:
        updates.append(
            gspread.models.Cell(row + 2, update_sheet_headers.index(col) + 1, value = updated_data[row][col]))

print("Raw Cell list out of order")
for update in updates:
    print(update)

#  out data is out of order
#  <Cell R2C2 '=Main!C2'>
#  <Cell R2C3 '=Main!D2'>
#  <Cell R2C4 '=Main!A2'>
#  <Cell R2C1 '=Main!B2'>
#  This will cause not all updates to happen for update_cells
#  because of cell.row cell.col offsets in cell_list_to_rect
#  may cause cell col offset to be a negative

update_sheet.update_cells(updates, value_input_option = 'USER_ENTERED')
```

This issue arises due to the need to offset rows and columns in utils.py  cell_list_to_rect().

```python
row_offset = cell_list[0].row
col_offset = cell_list[0].col
```

Since it is using index 0, it is offsetting by the incorrect amount in this edge case. In cases where cell_list is constructed in the correct sequence, this is not an issue.

A workaround that I've used in the past would be something like this:

```python
updates = []
for row in updated_data:
    for header in update_sheet_headers:
        if header in updated_data[row]:
        # for col in updated_data[row]:
            updates.append(
                gspread.models.Cell(row + 2, amazon_data_headers.index(header) + 1, value = updated_data[row][header])
                )
```

Ideally, I would restructure this and recreate a Cells object which could be indexable.

Tweaking the function cell_list_to_rect() to sort cell_list by rows and columns prior to formatting to send to the Google API gives the most flexibility.

```python
def cell_list_to_rect(cell_list):
    if not cell_list:
        return []
    
    cell_list.sort(key = lambda x: (x.row, x.col))
    ...
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+# Docstrings and comments use max_line_length = 79
+[*.py]
+max_line_length = 119
--- a/README.md
+++ b/README.md
@@ -172,6 +172,45 @@ for cell in cell_list:
 worksheet.update_cells(cell_list)
 ```
 
+### Formatting Cells
+ Basic formatting of a range of cells is supported. All basic formatting components 
+of the v4 Sheets API's `CellFormat` are present as classes in the `gspread.format` module, 
+available both by `InitialCaps` names and `camelCase` names: for example, the background color 
+class is `BackgroundColor` but is also available as `backgroundColor`, while the color class is `Color`
+but available also as `color`. Attributes of formatting components are best specified as 
+keyword arguments using `camelCase` naming, e.g. `backgroundColor=...`. Complex formats 
+may be composed easily, by nesting the calls to the classes.  
+ See [the CellFormat page of the Sheets API documentation](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets#CellFormat) 
+to learn more about each formatting component.
+ ```python
+ from gspread.format import *
+ fmt = cellFormat(
+    backgroundColor=color(1, 0.9, 0.9),
+    textFormat=textFormat(bold=True, foregroundColor=color(1, 0, 1)),
+    horizontalAlignment='CENTER'
+    )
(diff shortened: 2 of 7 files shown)


burnash:
Sorry for the late response and thank you for the contribution. I believe this problem has been addressed in #613. Could you please update gspread and check if the problem is resolved?
I'm closing this PR for now. Please open a new issue if the problem persists.

PR #613, opened 2018-10-28
Fix update_cell on non rect cell list

skaparis:
This fix is for the case then list of cell passed to sheet-update_cell are not top-left conner of upate rect

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -216,8 +216,8 @@ def cell_list_to_rect(cell_list):
 
     rows = defaultdict(lambda: {})
 
-    row_offset = cell_list[0].row
-    col_offset = cell_list[0].col
+    row_offset = min(c.row for c in cell_list)
+    col_offset = min(c.col for c in cell_list)
 
     for cell in cell_list:
         row = rows.setdefault(int(cell.row) - row_offset, {})


burnash:
Thank you for the contribution. Could you please give a more detailed code example of the problem? Could you also add a regression test case?

SolusAU:
```
c1 = Cell(3,1,'ABC')
c2 = Cell(4,2,'ABC')
c3 = Cell(5,3,'DEF')

cell_list = [c3,c1,c2]
wks.update_cells(cell_list)

 'updatedRange': 'rofl_test!A3',
 'updatedRows': 1,
 'updatedColumns': 1,
 'updatedCells': 1}
```

It doesn't create a range based on the minimum and maximum row,cols available.

emlazzarin:
> Thank you for the contribution. Could you please give a more detailed code example of the problem? Could you also add a regression test case?

Right now, the code checks the first cell provided in the `cell_list` to determine the "top-left corner" (minimum value) of the range to-be-updated:

    row_offset = cell_list[0].row
    col_offset = cell_list[0].col

If you ever submit a cell list where the "top-left" cell is not first, this will fail to produce a valid range, and will mangle the sheet.

The submitted diff ensures that the "top-left" (minimum row and minimum column) are used to begin drawing the rectangle-being-updated. Although this may not always be a valid cell, it is always the appropriate point to begin drawing the rectangle-being-updated.

I also checked @SolusAU's example, and it seems to work - nor do I understand the issue they are presenting.

hexvolt:
just merge it, it's been a year, hitting the same issue

emlazzarin:
I would but @burnash hasn't given anyone permission to commit. He's held the repo hostage.

This repo is a great candidate for https://www.codeshelter.co/.

hexvolt:
@emlazzarin yes thanks, that was a message to @burnash :)

burnash:
@skaparis Thank you very much for the fix.

@SolusAU, @emlazzarin thank you for the example and for the detailed explanation.

PR #621, opened 2018-11-03
Add with_link attribute when sharing / adding permissions

epicfaace:
The with_link attribute (https://developers.google.com/drive/api/v2/reference/permissions) (API doc) 
This PR:
- Adds a `with_link` kwarg to `gspread.client.insert_permission`
- Adds a `with_link` kwarg to `gspread.models.Spreadsheet.share`
- By default, `with_link` is set to false (so publicly accessible spreadsheets are publicly findable), but when it is sent to true, then publicly accessible spreadsheets are only accessible for users with the link.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -275,7 +275,8 @@ class Client(object):
         perm_type,
         role,
         notify=True,
-        email_message=None
+        email_message=None,
+        with_link=False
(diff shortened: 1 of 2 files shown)


burnash:
Thank you for the contribution!

Issue #158, opened 2014-08-24
Feature Request: duplicate spreadsheet


felagund:
Hi, it would be great if it were posible to duplicate a spreadsheet (including formulas) - I can code to copy values, but formatting on google's side is lost (I was trying to first save a backup of a spreadsheet I am working on before proceeding furhter).

burnash:
Implemented in #625 
Available in version 3.1.0.

PR #625, opened 2018-11-10
A method to copy spreadsheets

dsask:
Adds a method to copy spreadsheets, as requested in Issue #158. This is the same functionality as Pull Request #459 but reworked much like @burnash suggested.

Usage example:
```
spreadsheet_copy = client.copy(key, title="Copied Sheet", copy_permissions=True)
```

A title for the new spreadsheet can be specified optionally; the default is "Copy of <original_title>".

`copy_permissions=True` will share the new spreadsheet with everyone who had access to the original. This can be slow for sheets that are shared with many users and is disabled by default.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -214,6 +214,48 @@ class Client(object):
         spreadsheet_id = r.json()['id']
         return self.open_by_key(spreadsheet_id)
 
+    def copy(self, key, title=None):
+        """Copies a spreadsheet.
+
+        :param key: A key of a spreadsheet to copy.
+        :type title: str
+
+        :param title: A title of a new spreadsheet.
+        :type title: str
+
+        :returns: a :class:`~gspread.models.Spreadsheet` instance.
+
+        .. note::
+
+           In order to use this method, you need to add
+           ``https://www.googleapis.com/auth/drive`` to your oAuth scope.
+
+           Example::
+
+              scope = [
+                  'https://spreadsheets.google.com/feeds',
+                  'https://www.googleapis.com/auth/drive'
+              ]
+
+           Otherwise you will get an ``Insufficient Permission`` error
+           when you try to copy a spreadsheet.
+
+        """
+        if title is None:
+            original = self.open_by_key(key)
+            title = "{} (copy)".format(original.title)
+        payload = {
+            'title': title,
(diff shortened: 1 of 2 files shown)


gspread/client.py
```diff
@@ -214,6 +214,75 @@ def create(self, title):
...
+            original = self.open_by_key(key)
+            title = 'Copy of {}'.format(original.title)
+
+        payload = {
+            'title': title,
+            'mimeType': 'application/vnd.google-apps.spreadsheet'
+        }
+        r = self.request(
+            'post',
+            'https://www.googleapis.com/drive/v2/files/' + key + '/copy',
```
burnash: Please use `DRIVE_FILES_API_V2_URL` instead of hardcoded API url. Here's an example from `Client.list_permissions`:

https://github.com/burnash/gspread/blob/3f343fd0e440fb8c57c44ee00adbf312fe9b92d2/gspread/client.py#L270

gspread/client.py
```diff
@@ -214,6 +214,75 @@ def create(self, title):
...
+              scope = [
+                  'https://spreadsheets.google.com/feeds',
+                  'https://www.googleapis.com/auth/drive'
+              ]
+
+           Otherwise you will get an ``Insufficient Permission`` error
+           when you try to copy a spreadsheet.
+
+        """
+        if title is None:
```
burnash: I believe this entire `if` block is not needed. I just tested it locally and if you omit title from the API request or pass None, Sheets API will name the copy of the spreadsheet "Copy of _ORIGINAL_NAME_".

dsask:
Thanks for the feedback, this is my first contribution so I appreciate it. I'll make the changes you suggest.

dsask:
Not sure why tests are failing btw, they pass for me locally.

burnash:
Thank you! The tests are failing because on TravisCI tests rely on recorded HTTP sessions ([Example](https://github.com/burnash/gspread/blob/3f343fd0e440fb8c57c44ee00adbf312fe9b92d2/tests/cassettes/ClientTest.test_create.json)). There are no real HTTP requests. This is done with [Betamax](https://betamax.readthedocs.io/en/latest/). The setup is still pretty complex and some tests are still broken (you can see it in other PRs).

I'm merging the PR for now. I'll get back to sessions recording issue for this test case as I fix #610

Issue #629, opened 2018-11-28
sheet.get_all_records throws IndexError on empty sheets


AivanF:
### Environment info

Operating System: Mac OS X 10.14.1
Python version: 3.6.5
gspread version: 3.1.0

### Steps to reproduce
1. Create an empty sheet at Google Sheets, set up access
2. Call `sheet.get_all_records()`

### Sample code

```
client = gspread.authorize(creds)
document = client.open('SomeName')
worksheets = document.worksheets()
for sheet in worksheets:
	content = sheet.get_all_records()
	print(content)
```

### Output:

>1 sections:
Traceback (most recent call last):
  File "/Users/aivanf/.../sheets.py", line 26, in <module>
    content = sheet.get_all_records()
  File ".../python3.6/site-packages/gspread/models.py", line 629, in get_all_records
    keys = data[idx]
IndexError: list index out of range

### Expected behaviour:

The method should returns an empty list. Otherwise we need to use `try .. except ...` with manual setting the result to an empty list.

---

By the way, someone [created similar issue](https://github.com/burnash/gspread/issues/590), but with less description, so I decided to open a new one.

---

**Addition question**: why this method (and `append_row()` too) isn't presented in examples? I just worry about how many other useful things aren't presented, and I don't know about them.

burnash:
Thank you for the exemplary bug report. That's indeed is a bug.

burnash:
> Addition question: why this method (and append_row() too) isn't presented in examples? I just worry about how many other useful things aren't presented, and I don't know about them.

It's unintentional. Probably historical reasons. All the examples used to live in [README](https://github.com/burnash/gspread/blob/master/README.md) only, however recently they got [a dedicated section in the docs](https://gspread.readthedocs.io/en/latest/user-guide.html). The documentation contributions to this section are more than welcome!

PR #632, opened 2018-12-02
Fixed IndexError in get_all_records

AivanF:
Added data length checking to prevent `IndexError` in `.get_all_records()` method of `Worksheet` class. This is a small fix for the issue #629.


AivanF:
I just tested the code again and found that there is a type, comparison must be by `<=`, not just `=`

PR #633, opened 2018-12-02
 Fixed IndexError in Worksheet.get_all_records()

AivanF:
Added data length checking to prevent `IndexError` in `.get_all_records()` method of `Worksheet` class. This is a small fix for the issue #629.

<sub>
// Sorry for the second similar pull request
</sub>

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -628,7 +628,7 @@ class Worksheet(object):
         data = self.get_all_values()
 
         # Return an empty list if the sheet doesn't have enough rows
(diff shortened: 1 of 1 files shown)


AivanF:
I don't know why the Travis CI fails, but I think this is because of some other code and test which aren't related to my code changes.

burnash:
Thank you for the contribution. Please ignore Travis failure, it uses recorded data and there's an unrelated issue with it.

PR #634, opened 2018-12-04
Exception message

lordofinsomnia:
I have noticed that when an error occurs that user does not receive user friendly message. Before my change user has received an json response in args as string.

`error.exception.args[0]` has value of 
```json
'{
  "error": {
    "code": 404,
    "message": "Requested entity was not found.",
    "status": "NOT_FOUND"
  }
}
'
```

after my change `args` has populated with dict with values from json.
```python
self.assertEqual(error.exception.args[0]['code'], 404)
self.assertEqual(error.exception.args[0]['message'], 'Requested entity was not found.')
self.assertEqual(error.exception.args[0]['status'], 'NOT_FOUND')
```

Thanks for great lib. :+1:

### DIFF
--- a/gspread/exceptions.py
+++ b/gspread/exceptions.py
@@ -45,6 +45,6 @@ class APIError(GSpreadException):
     def _text_from_detail(self, response):
         try:
             errors = response.json()
-            return errors['detail']
+            return errors['error']
         except (AttributeError, KeyError, ValueError):
             return None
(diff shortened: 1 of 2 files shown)


burnash:
Thanks for the PR. Please ignore Travis failure, it uses recorded data and there's an unrelated issue with it.

lordofinsomnia:
wuhu. :)
tnx.

PR #640, opened 2018-12-23
Add quotes to title for get_all_values - fixed #636

### LINKED ISSUE #636: get_all_values returns single value when worksheet name is in A1 notation
Thanks @burnash for the great work! I've been using gspread a lot lately and think I came across a bug when a worksheet is named in A1 notation.

### Steps to reproduce
1. Create a new worksheet "A2" and populate with data
2. Call get_all_values() on the worksheet
3. Only the value in cell A2 will be returned

I have tried to fix the issue in https://github.com/grlbrwrg/gspread/commit/de5c3b4fc7cf56436226862e3c5e6c1d4adfdf8a but not sure if this is the right way to do things (it's my first contribution!)

Cheers
Walter

grlbrwrg:
@burnash as discussed in #636, added the quotes and a test.

I played around with worksheets that contain single and double quotes - looks like it's not an issue. Names I've tried that returned the correct data:

'"
A1"'
'"A1'
ASD'"


burnash:
Sorry for the late response. Thank you very much for your contribution.

PR #643, opened 2019-01-04
Stop keyerror on permissions list comprehension

wooddar:
If there is an item in a worksheet's permissions list that does not have an `emailAddress` attribute, a `KeyError` will be raised in this listcomp. This issue can be solved by using dict.get() notation that will return `None` if the emailAddress key does not exist.


markopy:
Can this be merged please? It's an obvious bug with a trivial fix.

burnash:
Thank you very much for fixing this, @wooddar.

PR #646, opened 2019-01-09
Fix for sharing from Team Drives

wooddar:
This PR adds the supports team drives parameter to permissions insertion method for GSpread, allowing GSheets to be shared that are in Team Drives


burnash:
Thank you, @wooddar

PR #622, opened 2018-11-03
Update numericise with underscores test to work in python2 and <python3.6

epicfaace:
Otherwise, that single test is failing in Travis.

PR #649, opened 2019-01-12
Add with_link attribute for sharing

epicfaace:
This way, one can share a spreadsheet only if people have the link, for example.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -275,7 +275,8 @@ class Client(object):
         perm_type,
         role,
         notify=True,
-        email_message=None
+        email_message=None,
+        with_link=False
     ):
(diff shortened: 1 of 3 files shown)


burnash:
This looks like a duplicate of #621
Closing.

epicfaace:
@burnash oops -- the PR description / name was duplicate, but the actual code change was different -- I've just updated the PR name / description to accurately reflect the change I made.

burnash:
@epicfaace Thank you for the clarification, in this case, I believe, it got merged in #622 isn't it?

epicfaace:
@burnash yeah, you're right -- must have just made this PR by accident then! Thanks

PR #651, opened 2019-01-18
Clarify supported versions

hugovk:
* `python_requires` helps pip install the right version for the running Python
* Trove classifiers makes it clear on PyPI which versions are supported


burnash:
Thank you!

PR #666, opened 2019-02-27
Fixed offset computing in cell_list_to_rect

fsonntag:
First thanks for the great library!

The input for `cell_list_to_rect` is assumed to be sorted, as it uses the first cell in the given list for the offset. If it is not, there will be cell updates with negative indices, that will be just omitted.

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -230,8 +230,8 @@ def cell_list_to_rect(cell_list):
 
     rows = defaultdict(lambda: {})
 
-    row_offset = cell_list[0].row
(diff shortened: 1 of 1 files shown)


emlazzarin:
This same bug is addressed here: https://github.com/burnash/gspread/pull/613

Would love @burnash to take a look at these. I already made a comment on the other PR.

fsonntag:
@emlazzarin Sorry, didn't see that. The change is the same, so I'll close this one.

burnash:
Thank you for your contribution.
The change has been merged in #613

PR #679, opened 2019-05-06
Update models.py

David-NS:
When finding a worksheet by title, you should specify the worksheet title 'str', but in the function docs title is specified as a 'int'.


burnash:
Sorry for the late response and I'm sorry I've overlooked your fix. This has been fixed as part of #690

PR #684, opened 2019-06-06
Speed up Client.open

aiguofer:
When you have lots of files in your drive, list_spreadsheet_files can
be slow (and can make unnecessary requests if it has to
paginate). Letting Google do the title query should speed things up a
bit and keep the number of requests to a minimum.


burnash:
Works like a charm. Thank you very much!

Issue #687, opened 2019-06-30
Feature to set position of a worksheet while addition


djmgit:
Presently while adding a worksheet, the position of the worksheet (index) in the spreadsheet cannot be set. By default the new worksheet is added at the end.
It would be helpful if gspread allows to set the position of the worksheet.

burnash:
Thanks again for the suggestion and implementation.
Added in #688
Closing the issue.

PR #688, opened 2019-06-30
Issue #687 : Add feature to set position of worksheet

djmgit:
Solves issue #687. Added param index in add_worksheet method
to set index of the newly created worksheet

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -242,7 +242,7 @@ class Spreadsheet(object):
         except (StopIteration, KeyError):
             raise WorksheetNotFound(title)
 
-    def add_worksheet(self, title, rows, cols):
+    def add_worksheet(self, title, rows, cols, index=None):
         """Adds a new worksheet to a spreadsheet.
 
         :param title: A title of a new worksheet.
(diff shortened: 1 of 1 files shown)


djmgit:
@burnash  please have a look.

lucasklaassen:
This is a good change to make. Would be interested in seeing this be merged!

gspread/models.py
```diff
@@ -269,6 +271,9 @@ def add_worksheet(self, title, rows, cols):
             }]
         }
 
+        if index:
```
lucasklaassen:
```suggestion
        if index or index == 0:
```

If `index` is set to 0, this will return false and not send the index to google. Setting an `index` of 0 is very useful when you want your current worksheet to be the first worksheet within the spreadsheet.

djmgit:
@lucasklaassen thanks for the review, I will update my PR as soon as possible

djmgit:
@lucasklaassen done with the change you suggested.

lucasklaassen:
@burnash this change is valuable and the code looks good. Can you please review and merge?

burnash:
@djmgit thank you for your contribution.
@lucasklaassen thank you for the review.
Sorry to everyone for the late reply.

PR #690, opened 2019-07-07
Fix typos and docstring :type

pedrovhb:
Fixes a few "gsperad" typos and the docstring :type for `worksheet`'s `title` param (from int to str)


burnash:
Thanks for fixing this!

PR #694, opened 2019-08-10
Added add_rows method to simplify the procedure of adding multiple rows

fabytm:
In order to make the process of adding multiple rows to a spreadsheet easier, I've added an add_rows method that allows the user to add multiple rows at once, thus decreasing execution time (less API calls) and overcoming the API write limit.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -896,7 +896,7 @@ class Worksheet(object):
         """Adds rows to the worksheet and populates it with values.
         Widens the worksheet if there are more values than columns.
 
-        :param values: Values for new rows.
+        :param values: Values for new rows. Values must be a list of lists.
         :param value_input_option: (optional) Determines how input data should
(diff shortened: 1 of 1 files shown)


gspread/models.py
```diff
@@ -891,6 +891,29 @@ def append_row(self, values, value_input_option='RAW'):
         }
 
         return self.spreadsheet.values_append(self.title, params, body)
+    
+    def append_rows(self, values, value_input_option='RAW'):
+        """Adds rows to the worksheet and populates it with values.
+        Widens the worksheet if there are more values than columns.
+
+        :param values: Values for new rows.
```
mukund-murali: I think it would be better to mention that this has to be a `list of list`

burnash:
@fabytm thank you very much for your contribution. @mukund-murali thank you for the review.

PR #598, opened 2018-09-27
Ignore Numericising Specific Column

benjamindhimes:
Adds functionality to specify a column to ignore numericising. By specifying the column # in numericse_ignore in get_all_records, it will not numericise a specific column. Fixes #581.

Issue #568, opened 2018-07-18
Option to turn off automatic type conversion with get_all_records


adace123:
I have a case where I need to keep the leading zeros of a numeric column. Therefore, I'd like to have the option to turn off type conversion. Thanks.

gnilrets:
I ran into this one today and would also appreciate a way to handle this.  Looks like there was attempt to solve it here - https://github.com/burnash/gspread/issues/581, but the PR was closed for some unknown reason.

burnash:
Implemented in #701

Issue #231, opened 2015-05-11
Cells with leading 0's - Gspread doesn't receive data as formatted on screen?


Ralithune:
Hello all, this is my first time posting an issue to Github.

I am using Gspread in my Python script to pull down firmware versions of hard drives.  For many Seagate drives, those firmware revisions begin with 3 leading 0's - as in "0003" or "0004".

Google Spreadsheets is aggressive about reformatting these and attempts to make them just straight numbers - "3" and "4".  I can avoid this by setting the column to "Plain Text" via the menu:

Highlight Column -> Format -> Number -> Plain Text

I then enter the new values "4" -> "0004" and wait for the changes to be saved.

These values are not coming to Gspread literally, however - when they arrive in the hash, they arrive without the leading 0's: "4".

It should be noted that I can change the value to something like "ai3oerjgeij" and have it show up properly in my script, just not a number with leading 0's.

Is there some way I can avoid this aggressive data changing?  Google is sure assuming a lot when they do this by default - messing with data is dangerous.

burnash:
Sorry for the late reply. Could you please send a snippet of your code? Particularly a place where you're retrieving those values. This can help.

Thanks for the feedback :)

Ralithune:
Hey Burnash.  I'm just using sheet.get_all_records and looping through that - the values in the dict are altered when they arrive.

I've taken to adding "googleisdumb" to my numbers to work around this, so "googleisdumb0004" and then manually parsing out the "googleisdumb" in my python script.  Hacky.

msuozzo:
This is desired behavior for spreadsheets so, unfortunately, I feel that it's out of scope to adjust this behavior in gspread. What data are you representing with leading zeros?

Ralithune:
@msuozzo I'm representing drive firmware levels.  For Seagate SAS drives, their firmware rev appears as "0003" or "0004" when queried by Smartctl and proprietary SAS controller binaries.

The spreadsheet serves as documentation for these values, so when they appear as "4" and not "0004", that is not an exact match to the disk firmware level.

Furthermore, when my software is comparing these values, it tries to compare "0004" to "4" and finds them inequal.  There's absolutely no reason for me to convert them to int - I just need the literal string to be uploaded and displayed as I wrote it with no alteration.

I'd have thought that would be default behavior when it comes to any kind of spreadsheet software, but apparently I'm wrong.

Is there a way to accomplish this?  Is Gspread munging the values or is Google?

[...]

PR #701, opened 2019-10-16
Ignore Numericising Specific Columns in get_all_records

### LINKED ISSUE #581: specify text columns so get_all_records does not run numericise
I have an id column in my spreadsheet that contains alphanumeric strings. Some of these strings are being interpreted as numbers. Take the following example:
```
from gspread.utils import numericise_all

id_column = ['8085807dc0000000', '8085807e40000000', '8085807ec0000000', '8085805e40000000']
numericise_all(id_column)
```
Returns:
```
['8085807dc0000000', inf, '8085807ec0000000', inf]
```
Of course this means that the id column is no longer useful for uniquely identifying records.

I would like to be able to specify that `id_column` should be kept as a string. Maybe something like `sheet.get_all_records(text_columns=['id_column'])`?

benjamindhimes:
Allows passing in a list of column indices (1-indexed) to avoid numericising them in the get_all_records method, also allows for the special case ['all'] to avoid numericising all columns. Fixes #581,  #568, #231 Improvement over my previously-closed PR of #598.


vijay-shanker:
@burnash this looks like an interesting problem, but the PR has been closed, should be picked up, can i ?

bhimesuber:
@vijay-shanker this PR is not closed

burnash:
@benjamindhimes Thank you very much for your contribution. I'm really sorry for the late response and I appreciate your persistence. I'll review the PR shortly.

gspread/models.py
```diff
@@ -744,14 +744,15 @@ def get_all_records(
         head=1,
         default_blank="",
         allow_underscores_in_numeric_literals=False,
+        numericise_ignore=[]
```
burnash, line 747: A [mutable default argument](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) may lead to surprising behavior.

benjamindhimes: You're quite right. Fixed.

gspread/utils.py
```diff
@@ -141,10 +141,27 @@ def numericise(value, empty2zero=False, default_blank="", allow_underscores_in_n
 
     return value
 
+def numericise_all(input, empty2zero=False, default_blank="", allow_underscores_in_numeric_literals=False,
+                   ignore=[]):
```
burnash, line 145: Sorry, the same problem is here.

gspread/utils.py
```diff
@@ -141,10 +141,27 @@ def numericise(value, empty2zero=False, default_blank="", allow_underscores_in_n
...
+    :param input: list row
+    :param empty2zero: bool indicating whether or not to return empty cells as 0, default False
+    :param default_blank: str determining which value to use for blank cells, default empty str
+    :param allow_underscores_in_numeric_literals: bool indicating whether or not to allow visual underscores
+                                                  in numeric literals
+    :param ignore: list of ints of indices of the row (index 1) to ignore numericising
+    """
+    _numericise = partial(numericise, empty2zero=empty2zero, default_blank=default_blank,
+                          allow_underscores_in_numeric_literals=allow_underscores_in_numeric_literals)
+    if not ignore:
```
aiguofer, line 158: You could just set `ignore = ignore or []` and get rid of the `if` statement

gspread/models.py
```diff
@@ -778,15 +783,21 @@ def get_all_records(
...
-                row,
-                empty2zero,
-                default_blank,
-                allow_underscores_in_numeric_literals,
-            )
-            for row in data[idx + 1:]
-        ]
+
+        if numericise_ignore == ['all']:
+            values = [row for row in data[idx + 1]]
```
burnash, line 788: It looks like in this case it only iterates over one row. Shouldn't it be `data[idx + 1:]` like in an `else` clause?

burnash:
Works well. Thank you very much! Kudos for expanding the docstring for `numericise_all()`

mmsbr:
Hello friends! How can I use this fix in my code? I'm using gspread for python3 and I really need the leading zeros

mmsbr:
Just to answer my own question and to help others that may have the same problem. Now you can keep leading zeros just using get_all_records() like this: get_all_records(numericise_ignore=['all']) and gspread won't convert your strings! Thank so much for this @benjamindhimes !!! You are awesome!

Ehayanarle:
@mmsbr I was looking for the solving of a problem for a couple of days but it was so easy. I should read the fu***** documentation and code of the module. Thank you very much for the solution!

PR #704, opened 2019-10-27
Set default argument to add_worksheet method

katsuhisa91:
First thanks for the great library!

I want to set default argument to add_worksheet method.
We execute `gc.create('A new spreadsheet')` , processing succeeds and new spreadsheet have 1,000 rows and 26 columns.
But, we execute `spreadsheet.add_worksheet('A new worksheet')`,  an error ( `TypeError: add_worksheet() missing 2 required positional arguments: 'rows' and 'cols'` ) occurs. We have to set argument `rows` and `columns` using add_worksheet method.

I think most gspread library's users want to use add_worksheet method more simply.
What are your thoughts on that?

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -242,7 +242,7 @@ class Spreadsheet(object):
         except (StopIteration, KeyError):
             raise WorksheetNotFound(title)
 
-    def add_worksheet(self, title, rows, cols):
+    def add_worksheet(self, title, rows=1000, cols=26):
         """Adds a new worksheet to a spreadsheet.
 
         :param title: A title of a new worksheet.


burnash:
Very good suggestion, @katsuhisa91 

I think it's a clear API improvement. However, the implementation adds two "magic numbers": 1000 and 26. The problem with these magic numbers is that whenever Google Sheets API backend decides to use different default values, these numbers become outdated.
 
A more future-proof solution is to omit the defaults completely, making `rows` and `cols` optional arguments. That way we could specify `gridProperties` in `body` conditionally and let Google Sheets API backend decide on the default sheet dimensions (as in the case of `gc.create`).

katsuhisa91:
@burnash I am convinced of your opinion, so I will withdraw this PR. Thank you very much for your answer.

burnash:
@katsuhisa91 You're welcome. 
Let me know if you're interested in reworking this PR as opposed to closing it. It's a valuable improvement and I'm happy to help to introduce it.

PR #705, opened 2019-11-06
Add `Spreadsheet.values_batch_get` function

aiguofer:
This enables hitting the Google API method
`spreadsheets.values.batchGet`. See
https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/batchGet
for more details.

If it seems like something you might be interested in and able to incorporate I'd be happy to write tests and document if needed.


burnash:
Thank you for the PR, @aiguofer. I'm interested. I would only suggest making the method name consistent with its peers (e.g. `Spreadsheet.batch_update`). 

The idea behind this naming for low-level methods like this is to mirror the relevant API endpoints converting camel case to snake case where needed.

aiguofer:
Great, yeah that makes sense. I noticed all `*.values.*` functions are `values_` so I went with `values_batch_get`. Looking at adding a test, if I do `GS_RECORD_MODE=all` I believe it re-writes all the cassettes, is that the proper way to do it?

aiguofer:
Nvm, I figured out how to run just that one test. I've added a test and the cassete to the PR as well.

burnash:
Excellent, thank you very much!

PR #706, opened 2019-11-10
Update oauth2.rst - deprecated menus and outdated images

manasouza:
"APIs & auth" menu deprecation + fix exception value + remove outdated images


burnash:
Thank you very much for the contribution.

PR #711, opened 2019-12-05
Convert oauth2client credentials to google-auth

aiguofer:
The new google-auth library has google.auth.transport.requests.AuthorizedSession which is basically a requests Session with attached Credentials. It will handle token refreshes on its own and will set the headers. 

In order to support backward compatibility with `oauth2client`, I added a simple wrapper that converts `OAuth2Credentials` to `google.oauth2.credentials.Credentials` and `ServiceAccountCredentials` to `google.oauth2.service_account.Credentials`.

Using `AuthorizeSession` handles refreshing tokens and retrying automagically so login should not be needed. I wasn't sure how to handle the scenario when a user passes in their own `Session` object, so I kept  login so they can still call it to refresh the token.


gspread/client.py
```diff
@@ -38,20 +31,16 @@ class Client(object):
 
     """
     def __init__(self, auth, session=None):
-        self.auth = auth
-        self.session = session or requests.Session()
+        self.auth = convert_credentials(auth)
+        self.session = session or AuthorizedSession(self.auth)
```
deargle: AuthorizedSession is not defined

deargle: I'm probably overlooking something, but all I've tried so far is using https://developers.google.com/sheets/api/quickstart/python , returning `creds`, and using this PR instead of normal gspread, like so:

`gspread.authorize(creds)`

deargle: adding `from google.auth.transport.requests import AuthorizedSession` at the top of `def __init__` seems to fix it.

aiguofer: Hey, thanks for reviewing! I should have probably re-checked this before submitting. You're right, I was just missing the import. I've added it now and it should be working.

aiguofer:
Updated docs to discuss new library and the fact that `oauth2client` credentials will be converted.

artemrys:
I have prepared a very small change that will fix Travis CI: https://github.com/aiguofer/gspread/pull/1
@aiguofer you can merge it.

aiguofer:
@artemrys great thanks, I merged it. However, I find it odd that it is required since I did add `google-auth` as a requirement for `gspread` in `setup.py`. That being said, your PR reminds me I didn't change the tests... 

I'm not sure what the best approach is here. For adding a test for this change, I'm thinking a basic assert to make sure that the `ServiceAccountCredentials` are correctly converted to `google.oauth2.service_account.Credentials` could suffice. In this case, we'd still need `oauth2client` as a test requirement. Alternatively, I could change the test to just use the new `Credentials` from `google-auth` and not test the conversion (which would allow us to remove the `oauth2client` dependency). That being said, it's possible that using `google-auth` instead of `oauth2client` will change every single cassette. Should I re-record all of these?

aiguofer:
Interesting, it looks like `nosetests` has a hard time with using full submodule paths unless they're imported correctly, which is what was causing the error. I've changed the test to use `google-auth` instead of `oauth2client` and changed the imports to be more explicit so `nosetests` can import correctly.

However, the use of `DummyCredentials` breaks the tests based on my implementation...

One option is to remove the checks in `convert_credentials` that force the creds to come from `google-auth`.. but that would probably break the `Client.login` code when non `google-auth` credentials are used.

Not sure what the best path is here... I wonder who actually uses their own custom credentials? is this a feature that needs to be preserved? and if so... what's the best way to address this? The old code forces `refresh` to meet the `oauth2client` signature, but the `google-auth` signature expects a `Request` object instead of an `http` object.

aiguofer:
@burnash Do you have any thoughts on this? I think moving over to `google-auth` is probably the right thing to do, but backward compatibility is probably pretty important (at least until next major version).

xinbinhuang:
Thanks for the work @aiguofer @burnash ! Will love to see this merged in

artemrys:
[oauth2client](https://github.com/googleapis/oauth2client) is being deprecated quite a lot time since now. Should we still support it?

aiguofer:
@artemrys although it has been deprecated for a while, `gspread` has continued to use it. If we fully remove support for it a lot of people's stuff might break. It'd be ideal if there was at least a transition period.

burnash:
@aiguofer Thank you very much for the PR. 
@deargleTthank you for the review.

Totally agree to move on from oauth2client. I really like the idea of gracefully dropping its support with backward compatibility (I think I'll also add deprecation warnings).

@aiguofer 
>  That being said, it's possible that using google-auth instead of oauth2client will change every single cassette. Should I re-record all of these?

I'm preparing a release for 3.4.0 which will include this PR. Some unrelated unittests need cassette updates so I'll re-record everything anyways.

gspread/client.py
```diff
@@ -39,15 +37,13 @@ class Client(object):
...
 
     def login(self):
-        """Authorize client."""
-        if not self.auth.token or (hasattr(self.auth, 'expired') and self.auth.expired):
```
burnash, line 47: @aiguofer I'm fixing the tests now, and I noticed that dropping this condition has increased the number of token refresh requests (there's `login()` in `setUp`). You probably had a good reason for removing the condition, but I can't figure it out. Why is it not needed any longer?

aiguofer: The new Google auth library handles refreshes automatically, which renders this `login` function unnecessary. I think I wasn't sure how to handle that when I submitted the PR and I thought it could be used when someone wants to force a refresh for some reason.

We could probably just remove it altogether and add a deprecation notice.

PR #719, opened 2020-01-07
add insertDataOption to prevent data overwrite

lobatt:
Referencing Google API V4 document: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/append#InsertDataOption


burnash:
Thank you very much for your contribution. Sorry for the late reply. 
I agree it's valuable to handle `insertDataOption`. However, to make `append_row` backward compatible we have to use the Sheets API default, which appears to be `OVERWRITE`.

Issue #447, opened 2017-03-16
Feature Request: 'Protect range...'-equivalent. Manage permissions on cell level.


tomrijntjes:
To use Google Sheets as data acquisition tool, it would be nice to have some sort of guarantee the cells did not change after retrieval, ie. by setting the permissions of the processed rows at read-only. 

Google Sheets provides this feature in their UI (right click: 'Protect Range...'). I briefly scanned the Drive API docs but couldn't find anything on cell level permissions. If this would be possible otherwise this would open up some interesting use cases.

jeremyherbert:
Looks like this is now possible: https://developers.google.com/sheets/api/samples/ranges

nineleaps-vijay:
@burnash can we start working on this one, I suppose many people might be needing this one and will be a good add-on to existing set of features. We can 

- create, delete and update a protectedRange
- clear sheet of all protectedRange
- get list of all protectedRange
- get list of protected Cells with each range

nineleaps-vijay:
@burnash can we work on this feature, as this is quite useful and well needed. We can have create, update and delete a protectedRange now with the API.

burnash:
@nineleaps-vijay sorry for the late response. Let me know if you are still interested in contributing.

vijay-shanker:
hey @burnash yes will very much like to do it. Do you have any suggestions/input on starting it up. (nineleaps-vijay is my office GitHub profile )

burnash:
Great, thanks for the response. When I left the comment I didn't notice that there's #720 which I believe is relevant. I haven't yet looked deeply into it though. If you have time, could please you review/test the PR?

PR #720, opened 2020-01-17
add worksheet protected range for issues #447

KesterChan01:
Regarding issue #447, add protect ranges in worksheet
https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/sheets#ProtectedRange

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -986,23 +986,33 @@ class Worksheet(object):
         }
 
         return self.spreadsheet.batch_update(body)
-    def add_protect_ranges(self, start_row_index,end_row_index,start_column_index,end_column_index,
-                           description=None,  request_user_edit=False):
+    def add_protect_ranges(self, start_row_index,end_row_index,start_column_index,end_column_index, 
+                           editors_emails=None, description=None,  request_user_edit=False):
         """"Add protect ranges into the selected worksheet. Only the editors can edit the protected ranges
         :param start_row_index: Index of the start row.
-        :type name_range: int
+        :type start_row_index: int
         :param end_row_index: Index of the end row.
-        :type name_range: int
+        :type end_row_index: int
         :param start_column_index: Index of the start column.
-        :type name_range: int
+        :type start_column_index: int
         :param end_column_index: Index of the end column.
-        :type name_range: int
+        :type end_column_index: int
+        :param editors_emails: List for more editors email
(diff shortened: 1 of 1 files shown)


gspread/models.py
```diff
@@ -986,6 +986,65 @@ def delete_row(self, index):
         }
 
         return self.spreadsheet.batch_update(body)
+    def add_protect_ranges(self, start_row_index,end_row_index,start_column_index,end_column_index,
```
aiguofer, line 989: For consistency, using the `cast_to_a1_notation` decorator might be better. This is currently used in a few other places and allows end users to either use A1 range notation or start/end indexes. For example, look at the `range` function.

gspread/models.py
```diff
@@ -986,6 +986,65 @@ def delete_row(self, index):
...
+        :type end_column_index: int
+        :param editors_emails: List for more editors email
+        :type editors_emails: list
+        :param description: description for the protected ranges
+        :type description: str
+        :param request_user_edit: True if the user who requested this protected range can edit the protected area.
+        :type request_user_edit: boolean
+        """
+        emailAddresses = []
+        if (editors_emails != None) & (isinstance(editors_emails,list)):
```
aiguofer, line 1008: You might be able to greatly simplify this code by making the default for `editors_emails=[]`. Then you could simply extend `emailAddresses` with `editors_emails`

gspread/models.py
```diff
@@ -986,6 +986,65 @@ def delete_row(self, index):
...
+            for email in editors_emails:
+                emailAddresses.append(str(email))
+        elif editors_emails == None:
+            pass
+        else:
+            raise Exception('Invalid type for editors_emails, only list is accepted')
+            
+        for permission in self.client.list_permissions(self.spreadsheet.id):
```
aiguofer, line 1016: This could be written more succinctly and with no try/except... for example:
```
emailAddresses.extend(permission.get('emailAddress') for permission in self.client.list_permissions(self.spreadsheet.id) if permission.get('emailAddress'))
```

Generally, a try/except without explicitly catching the type of exception you're interested in is not a great design choice.

KesterChan01:
@aiguofer Thanks you for the suggestions. I just made the changes in new commit.

burnash:
Thank you very much for your contribution!

PR #717, opened 2019-12-20
Fix get_all_value does not follow A1 notation.

zynaxsoft:
Resolve #716 

Regarding the issue, when a worksheet is named similar to A1 notation the requesting URL to the google API v4 could be misunderstood.

For example, if the worksheet is named `v1` the request in `get_all_value` would be
https://sheets.googleapis.com/v4/spreadsheets/[sheet]/values/v1

This breaks the method since the request would only return one value which is the value of the cell `V1`.

A simple fix is to make `get_all_value` follows the [A1 notation](https://developers.google.com/sheets/api/guides/concepts#a1_notation) that the worksheet name should be enclosed by a single quotation.

```
data = self.spreadsheet.values_get("'%s'" % self.title)
```

### Tests
Originally from master branch, the test already fails with 2 errors on
 * `test_values_batch_get`
 * `test_values_get`

(The result of the two tests is exactly the same as the test results below.)

With this pull request it results in 7 errors (including the above 2).
I am not sure if the existing bug effect this pull request or not. The fix works with `get_all_values()` in my environment. If there is any suggestion please let me know.

Here are the test results.

```
$ nosetests -vv tests/test.py

nose.config: INFO: Ignoring files matching ['^\\.', '^_', '^setup\\.py$']
test_numeric_value (tests.test.CellTest) ... ok
test_properties (tests.test.CellTest) ... ok
test_access_non_existing_spreadsheet (tests.test.ClientTest) ... ok
test_copy (tests.test.ClientTest) ... ok
test_create (tests.test.ClientTest) ... ok
test_import_csv (tests.test.ClientTest) ... ERROR
test_no_found_exeption (tests.test.ClientTest) ... ok
test_openall (tests.test.ClientTest) ... ok
test_add_del_worksheet (tests.test.SpreadsheetTest) ... ok
test_get_worksheet (tests.test.SpreadsheetTest) ... ok
test_properties (tests.test.SpreadsheetTest) ... ok
test_sheet1 (tests.test.SpreadsheetTest) ... ok
test_values_batch_get (tests.test.SpreadsheetTest) ... ERROR
test_values_get (tests.test.SpreadsheetTest) ... ERROR
test_worksheet (tests.test.SpreadsheetTest) ... ok
test_worksheet_iteration (tests.test.SpreadsheetTest) ... ok
test_a1_to_rowcol (tests.test.UtilsTest) ... ok
test_addr_converters (tests.test.UtilsTest) ... ok
[...]

PR #730, opened 2020-02-21
When Sheet names in A1 Notation, get_all_values fetching from one cell alone #636

vijay-shanker:
Hi,
When the sheet name is in A1 Notation, get_all_values only return value from one single cell.
Wrapping them with quotes (like this 'A1') fixes the issue. In case of sheet name already contains a single quotes, we need to escape it with another single quote.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -690,8 +690,9 @@ class Worksheet(object):
             Empty trailing rows and columns will not be included.
         """
 
+        title = self.title.replace("'", "''")
         data = self.spreadsheet.values_get(
-            "'{}'".format(self.title),
+            "'{}'".format(title),
             params={'valueRenderOption': value_render_option}
         )
 


gspread/models.py
```diff
@@ -606,8 +607,15 @@ def get_all_values(self):
 
             Empty trailing rows and columns will not be included.
         """
-
-        data = self.spreadsheet.values_get(self.title)
+        try:
```
burnash: I don't think `try/except` block is needed. As far as I understand it should not be a problem always escape single quotes.

burnash:
Thank you for your contribution. This has been also reported and fixed in #640 and #717 
However, both solutions lack escaping. Could you rebase your PR and implement escaping?

burnash:
Let me know if you're still interested to go on with this PR. I understand it's been a month since you've submitted it and priorities may have shifted.

vijay-shanker:
sure, i will rebase and do escaping

vijay-shanker:
hey, made the changes, sorry for being late on this, usually it does not happens.

burnash:
No problem, thank you @vijay-shanker

burnash:
FYI: I cleaned up unrelated commits that got carried over to this PR, kept only the relevant and rebased it on top of master.

burnash:
Backlinking the issue: #636 

Thanks again, @vijay-shanker

PR #733, opened 2020-03-21
Drop Python 2.7 support.

artemrys:
Changes:
* Updated docs
* Removed Python 2.7 from Travis configuration 
* Removed imports that are not needed in Python 3
* Removed u'' strings
* Removed `Python 3 support` mentions in docs because it is not a feature any more :)

For Python 2 to 3 migration [2to3](https://docs.python.org/2/library/2to3.html) tool was used.


burnash:
Hello and thank you for your contribution. I think it's a very good idea. 

Could you please provide some background for this change in the PR description? I'm aware of EOL for 2.7 but it's better to be very explicit with compatibility changes.

artemrys:
Hi, I have updated the PR description.
I assume that part of the users is still using this library with Python 2.7. But it's like almost 3 months already passed from Python 2.7 EOL so it would not be so bad from our side to break the dependency of Python 2.7 and remove part of the code. WDYT?

P.S. since Python 3.8 is already released, I think that I can create a separate PR with adding Python 3.8 to Travis build configuration. WDYT?

burnash:
Thanks for the update. I've scheduled this for inclusion in version 4.

burnash:
Regarding the Travis build config I would be very grateful if you can include 3.8.
(As you may have noticed Travis tests are currently failing. I haven't yet got time to figure out why.)

artemrys:
> Regarding the Travis build config I would be very grateful if you can include 3.8.
> (As you may have noticed Travis tests are currently failing. I haven't yet got time to figure out why.)

Regarding Python 3.8. I have created a PR https://github.com/burnash/gspread/pull/736.

Regarding failing tests I have a created a separate issue for that: https://github.com/burnash/gspread/issues/737

lavigne958:
Hi @artemrys  I think we should get your PR up to date and merge it. it has been a while since python2.7 is dropped. 
Is it ok with you if you rebase your PR on master branch and this time once ready will get it merge before any conflict happen.

thank you.

lavigne958:
Old PR no response from PR owner, closing it in favor of #879 .

PR #734, opened 2020-03-21
Adds multiple rows with index

tr-fi:
Adds multiple rows to the worksheet at the specified index and populates it with values

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1355,7 +1355,7 @@ class Worksheet(object):
                       "sheetId": self.id,
                       "dimension": "ROWS",
                       "startIndex": index - 1,
-                      "endIndex": len(values) + 1
+                      "endIndex": len(values) + index - 1
                     }
                 }
             }]
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1287,43 +1287,6 @@ class Worksheet(object):
 
         return self.spreadsheet.values_append(range_label, params, body)
 
-    def append_rows(self, values, index=1, value_input_option='RAW'):
-        """Adds multiple rows to the worksheet at the specified index
-        and populates it with values.
-        The input should be a list of lists, with the lists each
-        containing one row's values.
-        Widens the worksheet if there are more values than columns.
-        :param values: List of row lists.
-        """
-
-        body = {
-            "requests": [{
-                "insertDimension": {
-                    "range": {
-                      "sheetId": self.id,
-                      "dimension": "ROWS",
-                      "startIndex": index - 1,
-                      "endIndex": len(values) + 1
-                    }
-                }
-            }]
-        }
-
-        self.spreadsheet.batch_update(body)
-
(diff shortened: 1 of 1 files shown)


burnash:
Thank you very much for your contribution. My bad I've missed this PR and added `append_rows()` from #556 and #694. 
However, I think your contribution is could be used as `insert_rows()` method that gspread is currently missing. For that, we would need to change the method name in the PR and the API call.

Let me know if you're interested in changing the PR.

garethsb-sony:
It would be very nice if the `"insertDimension"` and the values were put atomically. I think it would be possible by using `"updateCells"` ?

burnash:
@garethsb-sony very good point. Thank you for the suggestion. 

For the record, here's the link to `updateCells` docs: [UpdateCellsRequest](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/request#updatecellsrequest)

@tr-fi let me know if you need help with reworking the PR.

garethsb-sony:
Thanks, @burnash ! (I assume `insert_row` should be reworked to call the new `insert_rows` as well, like `append_row` in master calls `append_rows`?)

burnash:
@garethsb-sony yes, exactly. I'm also trying to figure out if the scalar versions of these methods are still needed. Maybe it would be a good idea to deprecate the scalar versions and reduce the API surface a bit.

tr-fi:
@burnash You right insert_rows is more reasonable. So I changed the method name to insert_rows.

gspread/models.py
```diff
@@ -1339,6 +1339,43 @@ def insert_row(
 
         return data
 
+        def insert_rows(self, values, index=1, value_input_option='RAW'):
```
burnash, line 1342: Thank you for the update. It looks like the indentation of this line is wrong.
Could you please fix this?

gspread/models.py
```diff
@@ -1339,6 +1339,43 @@ def insert_row(
...
+        :param values: List of row lists.
+        """
+
+        body = {
+            "requests": [{
+                "insertDimension": {
+                    "range": {
+                      "sheetId": self.id,
+                      "dimension": "ROWS",
+                      "startIndex": index - 1,
```
burnash, line 1357: It looks like there's a problem with insert indices.

Steps to reproduce:

1. I have a sheet with consecutive values:

Screenshot 2020-04-27 at 15 00 55

2. I call 

```python
wks.insert_rows([['r1.1','r1.2'],['r2.1', 'r2.2']], index=3)
```

Where `wks` is my worksheet.

3. I get 

Screenshot 2020-04-27 at 15 01 20

Where the expected result is 1, 2, r1.1, r2.1, 3, 4, etc in the column A. 
It looks like the method overwrites the cell with value "3".

tr-fi: Sorry I did not test thoroughly. I fixed it but need you test again

burnash: No problem, thank you for fast response

PR #738, opened 2020-03-25
Add Python 3.8 to Travis CI

artemrys:
Python 3.8 was released Oct. 14, 2019.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,11 @@ language: python
 
 matrix:
   include:
-  - python: "2.7"
   - python: "3.4"
   - python: "3.5"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,13 +2,11 @@ language: python
 
 matrix:
   include:
-  - python: "2.7"
```
burnash: I think it's better to keep 2.7 until we officially drop the support.

artemrys: Done. Sorry about that, I'll be more accurate next time :)

burnash:
Thank you!

Issue #328, opened 2015-11-09
find_in_row/col


DWiechert:
It would be nice to have functions on the Worksheet which allow for find/findall that only search in a given column/row.

Example worksheet:

| A | B | C |
| --- | --- | --- |
| foo | bar | something |
| bar | something2 | foo |

Example queries (row/col is zero based):

```
sheet.find_in_row("foo", 1)
sheet.find_in_col("bar", 0)
```

Expected results:

```
Cell(row = 1, col = 2)
Cell(row = 1, col = 0)
```

PR #739, opened 2020-03-28
Adding col code to find

alfonsocv12:
I have a problem with the find() function i cannot specify a column so i made this changes to the function

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1425,7 +1425,7 @@ class Worksheet(object):
         """
         return self.spreadsheet.values_clear(self.title)
 
-    def _finder(self, func, query):
+    def _finder(self, func, query, col):
         data = self.spreadsheet.values_get(self.title)
 
         try:
@@ -1433,11 +1433,7 @@ class Worksheet(object):
         except KeyError:
             values = []
 
-        cells = [
-            Cell(row=i + 1, col=j + 1, value=value)
-            for i, row in enumerate(values)
-            for j, value in enumerate(row)
-        ]
+        cells = self.get_cells(values, col)
 
         if isinstance(query, basestring):
             match = lambda x: x.value == query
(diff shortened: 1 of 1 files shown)


burnash:
Thank you for your contribution. The review is pending. Please stay tuned.
Linking the relevant issue: #328

burnash:
@alfonsocv12 I noticed that you added more find_* functions in 0c1b1e8. I prefer the first version of the function in 81406c2. Could you please keep it that way? I think it's a bit cleaner to have a single function with parameters as modifiers.

alfonsocv12:
O sorry I thought that you preferred the code as the issue that you send me https://github.com/burnash/gspread/issues/328 <https://github.com/burnash/gspread/issues/328> but yes I can revert the modifications

burnash:
No problem :) No, I've just found a similar issue and linked it so the PR has more context. Also, it would be easier to find this issue and close is after the PR is merged.

alfonsocv12:
Ready

gspread/models.py
```diff
@@ -1446,15 +1442,32 @@ def _finder(self, func, query):
 
         return func(match, cells)
 
-    def find(self, query):
+    def get_cells(self, values, col):
```
burnash: 1. I think `get_cells` is a too generic name for this method. 
2. Refactoring is good, but I'm not convinced we need this method as a public method of `Worksheet` class. The more public methods we have, the harder it would be to provide backward compatibility should we decide to change anything down the line. More on this in PEP8's [Designing for Inheritance](https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance).

alfonsocv12:
I think thats good solution, if you know a better name for the function let me know

burnash:
Thank you again for your contribution!

lavigne958:
This closes #328 too.

PR #741, opened 2020-03-31
Fill in same value for all merged cells when calling get_all_values

### LINKED ISSUE #700: Same value for all merged cells
Currently, when you read from a spreadsheet that has merged (combined) cells, the value is only kept for the first of the individual cells that compose the merged cell.

It would be great to have a flag, when reading a spreadsheet wih `get_all_records()`, to duplicate that initial value for all individual cells that compose the merged cell.

aiguofer:
Fixes #700

Thought about making this an option, but it seems like it should just be the default behavior.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -23,7 +23,8 @@ from .utils import (
     filter_dict_values,
     absolute_range_name,
     a1_range_to_grid_range,
-    accepted_kwargs
+    accepted_kwargs,
+    fix_merge_values
 )
 
 from .urls import (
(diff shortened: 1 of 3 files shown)


burnash:
@aiguofer I never used gspread with a sheet containing merged cells so I'm not sure how frequent is the scenario where you fetch all values from such a sheet.

I have two concerns with marking this a default behavior: 
  1. Wouldn't this change brake any existing code that uses gspread?
  2. Even if there are no merged cells, we do a second API call, doubling the number of API calls for `get_all_values()`

aiguofer:
Weird, I had added a comment which related to your 2nd point but hadn't actually submitted it. Here's my take on both:

> Wouldn't this change brake any existing code that uses gspread?

I don't think it would... if there's no merged cells then this code won't do anything. If there are, I would imagine the expectation is that the same value is displayed for all cells in the merged range. It'd be interesting to know how people are currently handling this and whether they're manually filling in the values. That being said, I would not be opposed to making a param to handle it only when desired.

> Even if there are no merged cells, we do a second API call, doubling the number of API calls for get_all_values()

This is certainly more concerning. I have an idea on how to handle this and will make a PR. I think keeping a copy of the entire Spreadsheet metadata on the Spreadsheet, and using a reference to the specific Worksheet metadata within a Worksheet might help. We could then update that when any API calls that change it happen. I think the hardest thing there will be to ensure that gets updated on any new metadata-updating function calls that get implemented.

Let's not merge this in yet, and I'll play around with how to handle the metadata.

aiguofer:
Sorry, been pretty busy recently... I have some local changes I haven't pushed yet. I'll try to get to it this week

burnash:
No problem :)

PR #742, opened 2020-03-31
Improve merge_cells and add merge_type parameter

aiguofer:
Various fixes to the `merge_cells` function including:

- Allowing A1 range name notation
- Fixing tests so it actually tests it works
- Add merge_type parameter

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1751,23 +1751,28 @@ class Worksheet(object):
             new_sheet_name
         )
 
-    def merge_cells(
-        self,
-        startRowIndex,
-        endRowIndex,
-        startColIndex,
-        endColIndex
-    ):
-        """Merge cells, using index coordonates.
+    @cast_to_a1_notation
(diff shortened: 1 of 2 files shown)


burnash:
Wow, that was fast! Thank you very much!

aiguofer:
Absolutely :) it's a little easier when I'm bringing over functionality I added to https://github.com/aiguofer/gspread-pandas/

burnash:
@aiguofer Thank you again for the PR. 
You made a lot of very valuable contributions to gspread. Would you be interested in joining gspread as co-maintainer? I've sent you an invitation to the project. If you're not interested that’s totally ok :)

lavigne958:
Nice improvement, if I may advise, in the repository the functions interacting with cells are divided in 2 sets, the first one which accept row/col indexes and the second one with the A1 notation, I believe it brings more clarity if you create a second function that accepts only A1 notation. 

But the test improvement is awesome mate :+1:

Issue #678, opened 2019-05-02
Feature request: Folder option for create and copy functions


fernandeznico:
Hi, how I explain here: https://github.com/burnash/gspread/issues/665#issuecomment-488540191 it's posible specify a folder (by your id) in "copy" or "create" functions from easy way, adding a few lines of code.

Correction (tested):
`'parents': [{'id': folder_id}]`

Abdellam1994:
Hello,
I also think that this feature would be really useful.
It will allow to create directly a spreadsheet into a folder. It will therefore avoid polluting the drive when creating spreadsheets at scale.
I can work on the feature :)

burnash:
@Abdellam1994 would love to see this implemented, please do a PR.

Abdellam1994:
Hello :)

I noticed that I only implemented the folder option for "create function" and not for the "copy function".
Could we reopen the issue ?
I will work on a PR asap :)

burnash:
Hi there! Sure, would love to get a PR.

PR #754, opened 2020-04-08
[#678] Add option folder_id when creating a spreadsheet

Abdellam1994:
Quick development in order to be able to create a spreadsheet within an existing drive folder.
It will allow to have an organized drive when creating spreadsheets at scale.


burnash:
Thank you very much for the contribution. Works great!

PR #756, opened 2020-04-11
Enable open by title for spreadsheets on G Suite Team drives

geoffbeier:
add `corpora` parameter to `Client.list_all_files()` so that files on all drives (including team drives) will be found.

Inside `Client.list_all_files()`, `files/list` is being called with the `supportsAllDrives` and `includeItemsFromAllDrives` parameters both set to `True`. The drive API will still not return items from Team drives unless the `corpora` parameter is present and set to `allDrives`.

Unless `corpora` is included and set, gspread can't open sheets on team drives by title.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -83,6 +83,7 @@ class Client(object):
 
         params = {
             'q': q,
+            'corpora': 'allDrives',
             'pageSize': 1000,
             'supportsAllDrives': True,
             'includeItemsFromAllDrives': True,


burnash:
@geoffbeier Thank you very much for your contribution. 

Unfortunately, I don't have access to a Team drive at the moment so I'm unable to test this PR.

Could anyone from the community take a look at the PR and review it?

geoffbeier:
OK. Here's where I found the parameter:

https://developers.google.com/drive/api/v3/reference/files/list

and for anyone verifying, using the "Try It!" button on the right side of that page while logged into a G Suite account (this works well on .edu G Suite accounts if you have a shared drive with a spreadsheet on it available to you) is very useful for understanding the interaction between these parameters.

For now accessing sheets by url/key works fine from shared drives, so there's a relatively painless if not-as-readable workaround.

sanmarval:
@burnash I have reviewed the solution proposed by @geoffbeier and it works perfectly for my Shared Drive.

geoffbeier:
@sanmarval Thanks for trying it out! It's still holding up well here, too.

lavigne958:
Hi @geoffbeier thank you for you contribution.

I noticed the following mention on Google API documentation:
_When able, use 'user' or 'drive', instead of 'allDrives', for efficiency._

Can you please make it as an option in the function `list_spreadsheet_files`, this way if someone wants to keep the current efficiency he/she can and if someone want to open a SpreadSheet in a shared drive it can too but at (apparently) some efficiency cost.

Adding the option will imply adding the option to the following calling methods too: `open` and `openall` as they call `list_spreadsheet_files`.

lavigne958:
I ran some tests again using shared drives. I tried most combinations of:
- `includeItemsFromAllDrives` (already set to True by default)
- `supportsAllDrives` (already set to True by default)

What I got as result is:
- `supportsAllDrives` = True:
  - `includeItemsFromAllDrives = True`: gives me access to files in shared drive
  - `includeItemsFromAllDrives = False`: returns: SpreadsheetNotFound error
- `supportsAllDrives` = False:
  - `includeItemsFromAllDrives = True`: returns an API Error: `supportsAllDrives` was not set to True, which means: `includeItemsFromAllDrives = True` implies `supportsAllDrives = True`
  - `includeItemsFromAllDrives = False`: returns: SpreadsheetNotFound error

In all these test I never set `corpora` and it works fine.

when I try with `corpora = user` then it works with `supportsAllDrives = True` and `includeItemsFromAllDrives = True`. Because it already works with the default value of `corpora` it becomes useless to set it.

when I try with `corpora = allDrives` then I must set `includeItemsFromAllDrives = True` or I get an API Error.

My conclusion is:
the `corpora` parameter is not necessary to be set. The current implementation out of the box (on the latest master branch) works and let me access shared drive files.
The `corpora` entry is a duplicate of `supportsAllDrives` which is already set to `True`.

Therefore I will close this PR.

PR #759, opened 2020-04-15
Added delete_column() method

rafa-guillermo:
delete_column(self, index):
Deletes the column from the worksheet at the specified index.
param index: Index of a column for deletion.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1598,7 +1598,28 @@ class Worksheet(object):
         }
 
         return self.spreadsheet.batch_update(body)
+        
+    def delete_column(self, index):
+        """Deletes the column from the worksheet at the specified index.
 
+        :param index: Index of a column for deletion.
+        :type index: int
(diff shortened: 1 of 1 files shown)


burnash:
Thank you very much for your contribution.

Linking the relevant Stack Overflow question: https://stackoverflow.com/questions/61213417/delete-remove-column-in-google-sheet-over-gspread-python-like-sheet-delete-row

The method will be available in the next release v3.5.0.

rafa-guillermo:
No worries! Glad to help. I was looking to add a delete_columns() method, if it interests you, as a modification of the current delete_rows(). I also noticed the range of indexes for the delete_rows() method is incorrect (should be from `start_index -1` to `end_index`, not to `end_index - 1`).

burnash:
Both the `delete_columns()` and the fix for `delete_rows()` are very welcome :)

PR #760, opened 2020-04-15
Fix: delete_rows() index error

rafa-guillermo:
Fixed index error for `endIndex` object in `delete_rows(start_index, end_index)`:
`end_index - 1` -> `end_index`
Fixed comment standard


burnash:
Very good, thank you so much!

PR #761, opened 2020-04-15
Added delete_columns(start_index, end_index)

rafa-guillermo:
Method `delete_columns(start_index, end_index)`:
Deletes multi columns from the worksheet at the specified index.
param `start_index`: Index of a first column for deletion.
param `end_index`: Index of a last column for deletion.


burnash:
Thank you very much!

PR #762, opened 2020-04-15
Simplified OAuth2 flow

burnash:
This is a first attempt to implement a simplified authentication for gspread. Inspired by excellent [gspread-pandas](https://github.com/aiguofer/gspread-pandas) I'd like to cut down the boilerplate for constructing credentials, so you can get the authorized client instance in your code with just:

```
import gspread
gc = gspread.oauth()
```

On the first run, this will look for OAuth 2.0 Client credentials in `~/.config/gspread/credentials.json`, open a browser window with the application authorization request, get the user's consent and store authorized credentials in the config directory. Next runs will use stored credentials.

Open questions:
1. Credentials security.
2. Default scope.
3. With this added convenience, should client credentials replace service accounts as recommended flow in the docs. This is related to (1) and (2) since a service account does not have access to any spreadsheet unless a spreadsheet is explicitly shared with this account. Using OAuth 2.0 Client credentials, on the other hand, gives the script direct access to the user data.

### DIFF
--- a/gspread/__init__.py
+++ b/gspread/__init__.py
@@ -13,6 +13,7 @@ __version__ = '3.4.2'
 __author__ = 'Anton Burnashev'
 
 
+from .auth import oauth
 from .client import Client
 from .models import Spreadsheet, Worksheet, Cell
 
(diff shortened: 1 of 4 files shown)


burnash:
Ok, let's merge it and try a soft launch :)

PR #763, opened 2020-04-16
[#757] Add ability to fetch cell address in A1 notation

### LINKED ISSUE #757: Cell class should support conversion to A1 notation
Currently, there is no straightforward way to fetch data in `A1` notation from `Cell` object (of course one can use `utils` package explicitly). It will be a great to have this feature. 
Let's say you have the following steps to be performed:
1. Find some specific Cell by content.
2. Get a list of cells starting from previously found Cell.

tivaliy:
Now it is possible to fetch cell addres in A1 notation by
accessing 'address' attribute of a cell instance

closes #757

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1918,6 +1918,11 @@ class Cell(object):
             return None
 
     @property
+    def address(self):
+        """Cell address in A1 notation."""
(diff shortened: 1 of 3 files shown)


burnash:
Perfect! Thank you very much.

PR #765, opened 2020-04-21
Add ability to create cell instance using A1 notation

tivaliy:
Naturally when we deal with Google Spreadsheets we'd like
to operate with letters (in A1 notation), rather then decimal
row/col addressing mode. This patch adds an ability to create
cells using A1 notation, eg.:

   cells = [Cell.from_address(f'A{i}', value=i) for i in range(1, 10)]
   self.sheet.update_cells(cells)

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1894,6 +1894,10 @@ class Cell(object):
         #: Value of the cell.
         self.value = value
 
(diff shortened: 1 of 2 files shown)


burnash:
Good idea, thank you very much!

PR #768, opened 2020-04-24
Add `auth.service_account()`

burnash:
Simplify auth with service accounts. This PR adds `gspread.service_account()` which wraps `Credentials.from_service_account_file` cutting down initialization code.

So instead of 

```python
import gspread
from google.oauth2.service_account import Credentials

scopes = ['https://spreadsheets.google.com/feeds',
         'https://www.googleapis.com/auth/drive']

credentials = Credentials.from_service_account_file(
    'path/to/the/downloaded/file.json', 
    scopes=scopes
)

```

You put your service account key in `~/.config/gspread/service_account.json` and `service_account()` does the rest:

```python
import gspread

```

### DIFF
--- a/gspread/__init__.py
+++ b/gspread/__init__.py
@@ -13,7 +13,7 @@ __version__ = '3.5.0'
 __author__ = 'Anton Burnashev'
 
 
-from .auth import oauth
+from .auth import oauth, service_account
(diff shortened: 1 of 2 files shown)


honzajavorek:
Awesome!

PR #775, opened 2020-05-19
Add get_values() to Worksheet

burnash:
This PR adds a more flexible alternative to `get_all_values()`:

```python
    # Return all values from the sheet
    worksheet.get_values()

    # Return all values from colums "A" and "B"
    worksheet.get_values('A:B')

    # Return values from range "A2:C10"
    worksheet.get_values('A2:C10')

    # Return values from named range "my_range"
    worksheet.get_values('my_range')
```

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -704,30 +704,105 @@ class Worksheet(object):
             for j, value in enumerate(row)
         ]
 
-    def get_all_values(self, value_render_option='FORMATTED_VALUE'):
-        """Returns a list of lists containing all cells' values as strings.
+    @accepted_kwargs(
+        major_dimension=None,
+        value_render_option=None,
+        date_time_render_option=None,
+    )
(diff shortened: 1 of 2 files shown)


AlexeyDmitriev:
There's worksheet.get(). Is it different?

burnash:
@AlexeyDmitriev that's a good point. The only difference is that `get_values()` calls `fill_gaps()`, while `get()` does not do any post-processing. 

`get_all_values()` is an old method and it's heavily used in the wild. I wanted to generalize it a bit and replace it with `get_values` that returns all values by default. A similar story goes for `get_all_records()`. Sometimes you don't really need *`_all_`* of them.

lavigne958 (approved):

lavigne958:
Fixed test suite by protection function `fill_gaps` in case we give it an empty list.

lavigne958 (approved):

PR #783, opened 2020-06-01
fix editor_users_emails - get only from list not all who had access t…

Lukasz87:
Fix for editor_users_emails.
It was setup to extend list of editor_users_emails for all users who had access to spreadsheet.
Change to users only from editor_users_emails.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1629,14 +1629,7 @@ class Worksheet(object):
             who requested this protected range can edit the protected cells.
             Defaults to ``False``.
         """
-        email_address = [
-            permission.get('emailAddress')
-            for permission in self.client.list_permissions(self.spreadsheet.id)
-            if permission.get('emailAddress')
-        ]
-
-        editors_emails = editor_users_emails or []
-        email_address.extend(email for email in editors_emails)
+        editor_users_emails = editor_users_emails or []
 
         editor_groups_emails = editor_groups_emails or []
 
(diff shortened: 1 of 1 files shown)


aiguofer:
I agree that's probably the right thing to do, my only concern is changing behavior for anyone that might be expecting this "append" style functionality. @KesterChan01 since you created the original PR, do you have any objections with this?

KesterChan01:
@aiguofer I suggest add an argument for choosing "append" or "replace" style.  And suggest that the editors email needs to include a client email provided by the service account. Otherwise, there is no rights for the API to modify that ranges.

lavigne958:
@KesterChan01 thank you for this PR.

I looked at it in details and I agree that:
- the given list of email addresses should be the only set of user allowed for that protected range.
- There is no reason to expect the current list of users with access to be in that protected range.
- This makes an extra call that we could skip.

My issue is that previous users would have a behavior change and might not notice it.

One easy way to do it is to take the problem the other way:
- add a new option (could be names like `add_current_users: bool`) which will offer the choice to add the current list of users to the given set
- **do not provide default values**, this will ensure anyone previous user will need to make a choice when upgrading the library.
- The default behavior would be to only set the users list given as argument of that function. but one can keep its old behavior.

What do you think ?

lavigne958:
I played a little bit with the protected range API call and using the Speadsheet web UI, noticed:

- you cannot remove yourself from the users access => we should always at least add the email used to access the files as editor
- you cannot add `@gmail.com` address that does not exists, gmail checks them and returns an error if this is the case
- by default the UI pre-select all current editors
- you can add external e-mails that are not part of the current users, the protected range will be created but the UI offers to share that spreadSheet with that e-mail. While that new external e-mail has not validated the sharing it is listed but not validated.
- Using the API call to add external emails, if the email is not part of the current editors then the protected range is created but still returns an error mentioning that the email address won't access to the file.

Thinking about it, I will push back that update, and try to find the right solution that offer some flexibility for end-users and follows the google API rules (current editor cannot remove itself, etc)

PR #790, opened 2020-06-23
Changed find() return to None if nothing found

GastonBC:
Very minor tweak, find() should return something if nothing was found. It's likely that sometime you won't find the cell and returning None seems more practical than raising an exception

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1780,7 +1780,7 @@ class Worksheet(object):
         try:
             return self._finder(finditem, query, in_row, in_column)
         except StopIteration:
-            raise CellNotFound(query)
+            return None
 
     def findall(self, query, in_row=None, in_column=None):
         """Finds all cells matching the query.


lavigne958:
Hi this is a design choice and can be discussed. Returning `None` or `raise` an exception leads to the same situation: _the user must protect the call to the method  `find`._

This change could break other users code as it changes the behavior but on the other hand does not introduce any major improvement/bugfix unfortunately. 

@burnash what do you think ?

lavigne958:
@GastonBC I looked back at your PR.

I agree we should return `None` if nothing is found, this would make user code easier.

Could you please update:
- the docstring of the `find` function, mention something like `returns the first matching cell, None otherwise`
- the doc about the `find` function in `docs/user-guide.rst`, mention the "returns None if nothing is found" with a example
- the test suite named `test_find` in `tests/test.py`, add a new call to the `find` function that will fail and check the return value is `None`

thank you :-) 

if you don't want to or don't have the time to, let me know I'll do it :wink:

lavigne958:
closing this PR in favor of #899.
current commit has been used in #899 .

PR #794, opened 2020-07-08
Update oauth2.rst

elnjensen:
- Slight labeling change in the initial step for Enable API Access
- Add step of configuring consent screen - I found I was redirected to this even when I chose "Credentials" first, so I think there's no way around it and I added it as an explicit step. 
- There was no "Other" option for the application category when creating credentials, so I chose "Desktop app" which seemed to work fine.


burnash:
Thank you very much!

PR #797, opened 2020-07-27
fix: no exception if file with credentials doesn't exist

Birdi7:
This fix will tell python to create a file with credentials if it doesn't exist. I believe this is expected behaviour because the client shouldn't manually create an empty file to store them

### DIFF
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -100,7 +100,7 @@ def load_credentials(filename=DEFAULT_AUTHORIZED_USER_FILENAME):
 def store_credentials(
     creds, filename=DEFAULT_AUTHORIZED_USER_FILENAME, strip='token'
 ):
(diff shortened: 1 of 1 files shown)


burnash:
Could you please elaborate on how to produce this exception? What steps should I follow to see it? The environment info (OS and Python version) will be helpful too. 
Also [`open`](https://docs.python.org/3.8/library/functions.html?highlight=open#open) with plain `w` mode creates the file if it does not exist. If I understand it adding the plus (`+`) to `w` tells open to open the file for reading as well and this is not needed in this particular case.

Birdi7:
Hmmm, strangely, I am not able to reproduce the bug now. I'll submit an issue if it will occur

burnash:
Great, thank you!

PR #799, opened 2020-07-28
Simplify implementation of insert_rows via insert_row

AlexeyDmitriev:
I've noticed that insert_row and insert_rows implemented identically, so one can be implemented in terms of the other.


AlexeyDmitriev:
I've run tests locally in master and in my branch, and the set of failed tests in the same.
I see that you have .travis.yml. How do I trigger travis build?

burnash:
Thank you very much!

Issue #576, opened 2018-08-02
Feature request: insert columns to a worksheet


qilinkang:
gspread has a function to resize a worksheet to append new columns. No function to insert columns at specific column.

I have implemented a method insert_cols(start_col, number_cols) in Worksheet class.

Please review this request and let me know if it's good feature.

lavigne958:
closed by #802

PR #802, opened 2020-07-29
Add insert_cols method to worksheet

### LINKED ISSUE #798: Inserting columns
I want to insert new columns (between existing columns)

It seems that it's supported by underlying API but not in the provided

I just need inserting empty column, but it's probably makes sense to make the API the same as with insert_row/insert_rows

I'm ready to provide a PR for that if it's fine

AlexeyDmitriev:
Closes #798

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -1553,6 +1553,43 @@ class Worksheet(object):
 
         return self.spreadsheet.values_append(range_label, params, body)
 
+    def insert_cols(self, values, col=1, value_input_option='RAW'):
+        """Adds multiple new cols to the worksheet at specified index and
+        populates them with values.
+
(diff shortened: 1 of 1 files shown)


AlexeyDmitriev:
cols is used more often then columns in this class so I went ahead and called it insert_cols, not insert_columns. Let me know if you think that I should rename it.

Also, I think it makes sense to add either add insert_col(umn) or make insert_row deprecated for consistency

iisulop:
Hi,
I would love to get to use this. Is there still something requiring action?

burnash:
Thank you very much for the PR. Yes, I think it makes more sense to deprecate `insert_row()` to remove duplicate functionality.

lavigne958:
closes #576

PR #809, opened 2020-08-27
Allow fetching named ranges from worksheets.

agatti:
Seems like worksheets only allow ranges in A1 form when retrieving data, although the v4 API works just fine if a named range is passed.  I've just tweaked range extends retrieval a bit to rely on what is returned by the API server if no A1 range is being passed.


lavigne958 (approved):
Tested it, works fine.

Thank you @agatti  for you contribution :slightly_smiling_face:

PR #838, opened 2021-01-07
Updated user-guide.rst - correct header spell

jayeshmanani:
Updated in Using gspread with NumPy

hearder --> header (with correct spelling of header)


burnash:
Thank you very much!

PR #844, opened 2021-01-19
Add include_values_in_response to append_row & append_rows

martimarkov:


martimarkov:
Can we merge this?

punnerud:
There is several parameters that the API-support. Would it be better to add **kwargs so include_values_in_response and others can be supported?

Something like this:
```
append_row(
        value_input_option='RAW',
        insert_data_option=None,
        table_range=None,
        , **kwargs
    ):

if kwargs is not None:
            payload = {**payload, **kwargs}
```

gspread/models.py
```diff
@@ -1500,6 +1509,7 @@ def append_rows(
         params = {
             'valueInputOption': value_input_option,
             'insertDataOption': insert_data_option,
+            'includeValuesInResponse': include_values_in_response,
```
martimarkov, line 1512: @punnerud But in here we still would need custom logic as for example `value_input_option` is not used here. I don't remember the API specs on the top of my head but maybe there are other things like that?

punnerud: Yes, we need to keep every parameter that have logic. All others could be passed directly to the API, and let the API give errors if they don’t exists.

Should make it a bit better than my previous code example, so the parameters can be placed in random order.

lavigne958:
closing this PR in favor of #957 .

@martimarkov  I cherry-picked your commit, this way you keep your contribution.

Issue #849, opened 2021-03-01
Feature Request: get_note()


marcusbeckerman:
There is a a clear_note() method, and insert_note() method, and an update_note() method, but there doesn't seem to be a method to get the note of a cell(s). Can this be added?

burnash:
Open for a PR

water-ghosts:
I've created a PR for this feature, added test cases, and confirmed that the test suite runs successfully.

This is my first time contributing to this project, so let me know if there are additional steps I'm missing, or if you'd prefer I step aside and let someone more experienced handle this request 🙂

PR #855, opened 2021-03-26
Add get_note method (#849)

water-ghosts:
Adds the worksheet.get_note() method requested in #849, with relevant test cases.

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -149,6 +149,22 @@ class Spreadsheet(object):
 
         return r.json()
 
+    def notes_get(self, range):
+        """Lower-level method that calls `spreadsheets.get <https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/get>`_
+        and retrieves the notes field.
+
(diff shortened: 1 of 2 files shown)


gspread/models.py
```diff
@@ -149,6 +149,22 @@ def batch_update(self, body):
 
         return r.json()
 
+    def notes_get(self, range):
```
burnash, line 152: I'm not convinced that there is a need for this public method in the `Spreadsheet` class. It'd be great to have as minimal api surface area as possible and I actually regret that I added other convenience methods as public methods in `Spreadsheet`.

Let me know if you have a specific reason for this public method. If you don't, please consider moving this code to `Worksheet.get_note`.

burnash (changes requested):
Perfect, thank you very much! Please see the comment for `Spreadsheet.notes_get`.

gspread/models.py
water-ghosts: I just added that method for consistency with values_get(). I’ve removed it and transferred the logic to worksheet.get_note()

burnash:
Great, thank you very much again!

PR #857, opened 2021-03-29
Added feature get_worksheet_by_id

a-crovetto:
Enables the possibility to get the worksheet by it's Id. This approach could be more robust because the worksheets ids never changes and it´s really easy to get it from the URL by looking the for the "**gid**" parameter value

New method:
get_worksheet_by_id

### DIFF
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -306,16 +306,13 @@ class Spreadsheet(object):
         sheet_data = self.fetch_sheet_metadata()
 
         try:
-            sheets_length = len(sheet_data['sheets'])
-            print(sheet_data['sheets'])
-            for i in range(0,sheets_length):
-                
-                if id == sheet_data['sheets'][i]['properties']['sheetId']:
-                    properties = sheet_data['sheets'][i]['properties']
-                    return Worksheet(self, properties)
-            raise WorksheetNotFound
-        except (KeyError):
-            return None
+            item = finditem(
+                lambda x: x['properties']['sheetId'] == id,
+                sheet_data['sheets'],
+            )
+            return Worksheet(self, item['properties'])
+        except (StopIteration, KeyError):
+            raise WorksheetNotFound(id)
 
     def worksheets(self):
         """Returns a list of all :class:`worksheets <gspread.models.Worksheet>`
--- a/gspread/models.py
+++ b/gspread/models.py
@@ -288,6 +288,34 @@ class Spreadsheet(object):
             return Worksheet(self, properties)
         except (KeyError, IndexError):
             return None
+    
+    def get_worksheet_by_id(self, id):
+        """Returns a worksheet with specified `worksheet id`.
+
+        :param id: The id of a worksheet. it can be seen in the url as the value of the parameter 'gid'.
+        :type id: int
+
+        :returns: an instance of :class:`gspread.models.Worksheet`
+                  or `None` if the worksheet is not found.
+
+        Example. To get first worksheet of a spreadsheet:
+
+        >>> sht = client.open('My fancy spreadsheet')
+        >>> worksheet = sht.get_worksheet_by_id(0):rtype: dict
+        """
+        sheet_data = self.fetch_sheet_metadata()
+
+        try:
+            sheets_length = len(sheet_data['sheets'])
+            print(sheet_data['sheets'])
+            for i in range(0,sheets_length):
+                
+                if id == sheet_data['sheets'][i]['properties']['sheetId']:
+                    properties = sheet_data['sheets'][i]['properties']
+                    return Worksheet(self, properties)
+            raise WorksheetNotFound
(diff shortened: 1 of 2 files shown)


gspread/models.py
```diff
@@ -288,6 +288,34 @@ def get_worksheet(self, index):
...
+        sheet_data = self.fetch_sheet_metadata()
+
+        try:
+            sheets_length = len(sheet_data['sheets'])
+            print(sheet_data['sheets'])
+            for i in range(0,sheets_length):
+                
+                if id == sheet_data['sheets'][i]['properties']['sheetId']:
+                    properties = sheet_data['sheets'][i]['properties']
+                    return Worksheet(self, properties)
```
lavigne958, line 315: Here it could be useful to use the same way of iteration as the function [workshet](https://github.com/burnash/gspread/blob/8c31c6df40f51728a82ee26e5b369a765618d191/gspread/models.py#L317) that uses the util function `finditem` and a labmda to filter.

If you need any help with this let me know I can guide you.

Then just below catch the following exceptions `StopIteration` and `KeyError`.

This will make it more consistent accress functions.

lavigne958:
closes #271

gspread/models.py
```diff
@@ -288,6 +288,34 @@ def get_worksheet(self, index):
...
+        Example. To get first worksheet of a spreadsheet:
+
+        >>> sht = client.open('My fancy spreadsheet')
+        >>> worksheet = sht.get_worksheet_by_id(0):rtype: dict
+        """
+        sheet_data = self.fetch_sheet_metadata()
+
+        try:
+            sheets_length = len(sheet_data['sheets'])
+            print(sheet_data['sheets'])
```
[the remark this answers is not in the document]
a-crovetto: Fixed

gspread/models.py
a-crovetto: Used the finditem fuction to perform the operation

a-crovetto:
Let me know I there is something else to change, thanks for the feedback

gspread/models.py
```diff
@@ -288,6 +288,31 @@ def get_worksheet(self, index):
...
+        :param id: The id of a worksheet. it can be seen in the url as the value of the parameter 'gid'.
+        :type id: int
+
+        :returns: an instance of :class:`gspread.models.Worksheet`
+                  or `None` if the worksheet is not found.
+
+        Example. To get first worksheet of a spreadsheet:
+
+        >>> sht = client.open('My fancy spreadsheet')
+        >>> worksheet = sht.get_worksheet_by_id(0):rtype: dict
```
lavigne958, line 304: Minor detail, could you use an example where we access a sheet that is not the first one please ? the first sheet can be obtained using `sht.sheet1` as a shortcut, so the most common use of this function will be to access another sheet.

tests/test.py
```diff
@@ -297,6 +297,10 @@ def test_sheet1(self):
     def test_get_worksheet(self):
         sheet1 = self.spreadsheet.get_worksheet(0)
         self.assertTrue(isinstance(sheet1, gspread.Worksheet))
+    
+    def test_get_worksheet_by_id(self):
+        sheet1 = self.spreadsheet.get_worksheet_by_id(0)
+        self.assertTrue(isinstance(sheet1, gspread.Worksheet))
```
lavigne958, line 303: Same here, if you could please adjust the test to reflect what the users would be likely to do :
- add a couple of sheets (3 for example seems enough)
- get the second sheet (id = 1)

as mentioned above the first sheet is already access via a shortcut so most likely the test would not reflect the use case of that function.

thank you.

lavigne958 (approved):
Thank you @a-crovetto .

These are minor changes if you have time to do them. otherwise I can do it.

Anyway thanks again for this feature :+1:

gspread/models.py
joaocasarin: He used `id=0` just to show how it would work, but the gid on the URL is often a big number, so it would become something like `sht.get_worksheet_by_id(2146239925)` for example. Instead of the user/developer use the method to get the worksheet by it's name, which he would need to copy its name or already know it, he's just need to copy the gid from url.

lavigne958: Hi @JoachimKoenigslieb , thank you for reviewing this PR.

I understand it is an example. :slightly_smiling_face: 

My point is:
the provided example does not reflect how the user is most likely going to use the method. We already have a shortcut to access the first sheet (using: `sht.sheet1`). So the user is most likely going to access another sheet, like the second one or the _Nth_ one. This is why I recomend to update the example.

PS: I did not noticed that sheet ID are not incremental (1st => `0`, 2nd => `1`, etc), but it is a big number with no relation one from another. Having the example reflecting this (using a example sheet ID like `123456789`) would have teach me this detail :wink:

lavigne958:
closes #271

PR #785, opened 2020-06-09
Add auth.service_account_from_dict

mahenzon:
Closes #784

PR #860, opened 2021-04-16
Document dictionary credentials auth

dmytrostriletskyi:
Hello,

the library provides two methods of auth: [service_account](https://github.com/dmytrostriletskyi/gspread/blob/master/gspread/auth.py#L158) and [service_account_from_dict](https://github.com/dmytrostriletskyi/gspread/blob/master/gspread/auth.py#L187). `service_account_from_dict` was added last year in [this pull request](https://github.com/burnash/gspread/pull/785), but wasn't documented. At least, I have not found anything:

Screenshot 2021-04-16 at 12 07 46

So, this pull requests adds some documentation for the `service_account_from_dict` function to [auth using service account documentation section](https://gspread.readthedocs.io/en/latest/oauth2.html#for-bots-using-service-account). I have built it locally, it looks like:

Screenshot 2021-04-16 at 12 19 59

If there is something I should add, just tell me. Thank you in advance! :)

### DIFF
--- a/docs/oauth2.rst
+++ b/docs/oauth2.rst
@@ -103,6 +103,27 @@ that has been used for authentication prio to the gspread version 3.6:
 
     gc = gspread.authorize(credentials)
 
+There is also the option to pass credentials as a dictionary:
+
+::
+
+    import gspread
+    
+    credentials = {
+        "type": "service_account",
+        "project_id": "api-project-XXX",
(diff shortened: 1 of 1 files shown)


burnash:
Brilliant! Thank you very much! Kudos for the screenshot of the final result :)

PR #864, opened 2021-05-03
Update oauth2.rst

londoniscold:
Changed "Desktop" to "Windows UI" as that appears to be a change in the process.


burnash:
Thank you for the PR. Could you please share more details on why this change is needed?
I don't see any changes related to "Windows UI" in the dropdown in question.


Screenshot 2021-05-03 at 11 20 40

londoniscold:
Ahh, ok, I maybe got waylaid at the wrong step ? I definitely never came across Desktop app in my dropdown, but i managed to make gspread work when I chose Windows UI in my dropdown. I never saw the dropdown you've shown ?

burnash:
Where do you see the dropdown with Windows UI? Maybe you could share the screenshot?

burnash:
The UI of Google Developers Console keeps changing and it is hard to keep track of it.

Issue #778, opened 2020-05-20
Unable to open sheet by sheetname in a Shared Drive


ryan-pixel:
**Important**: Please do not post usage questions here. 
To get a quick response, please ask a question on Stack Overflow using `gspread` tag.
See existing questions: https://stackoverflow.com/questions/tagged/gspread

---

**Describe the bug**
Getting SpreadSheetNotFound errors when trying to open a sheet inside a Shared Drive by its sheetname. Opening via the url works fine, but I can't open via the sheet name.

**To Reproduce**
1. Create a shared drive that's shared to your service account.
2. Create a Sheet within that shared drive.
3. Try to open via the sheet name.

**Expected behavior**
Trying to open a sheet via the sheetname using gc.open with a spreadsheet that's inside a Shared Drive. The service account has been shared with the service account.

**Code example***
`sheet = gc.open(sheetname)`

**Environment info:**
- Operating System [e.g. Linux, Windows, macOS]: Windows
- Python version: 3.7
- gspread version: latest

yohayonyon:
Was there any progress? I need this feature as well...

burnash:
@yohayonyon could you please check if solution in #756 works for you?

yohayonyon:
Hi @burnash 

It didn't. I had to add a usage f the parameter supportsAllDrives to the method Create.
I will create a PR.

lavigne958:
There is a PR on the way about this. Waiting for some feedback from the PR owner of #865 .
I still don't quiet get it between `supportAllDrives: True` and `corpora: 'allDrives`. but once both are merged you issue should be solved (from what I understand).

lavigne958:
@ryan-pixel after running some test (see latest comment in #865 )
I noticed GSpread is now able to open files in shared drives using default settings.

I recommend you give it second chance using the latest version from the master branch, as for now the latest release is coming but not out yet.

PR #865, opened 2021-05-03
supports_all_drives was added to Client.create - whether to look for …

yohayonyon:
…the folder_id in shared drives.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -174,13 +174,14 @@ class Client(object):
             for x in spreadsheet_files
         ]
 
-    def create(self, title, folder_id=None):
+    def create(self, title, folder_id=None, supports_all_drives=False):
         """Creates a new spreadsheet.
 
         :param str title: A title of a new spreadsheet.
 
-        :param str folder_id: Id of the folder where we want to save
-            the spreadsheet.
+        :param str folder_id: Id of the folder where we want to save the spreadsheet.
+
+        :param boolean supports_all_drives: whether to look for the folder_id in shared drives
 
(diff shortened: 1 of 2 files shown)


burnash:
Thank you for the PR. Could you move the last commit (2d72f12) into a separate PR? It'll be easier to test and also it'll be better documented that way.

yohayonyon:
Not so elegant, but done. :-)

gspread/client.py
```diff
@@ -193,7 +194,13 @@ def create(self, title, folder_id=None):
         if folder_id is not None:
             payload['parents'] = [folder_id]
 
-        r = self.request('post', DRIVE_FILES_API_V3_URL, json=payload)
+        params = None
+        if supports_all_drives:
+            params = {
+                'supportsAllDrives': True
+            }
```
lavigne958, line 201: It works if we set the parameter to `True` even if the user want to save the file in its own drive.

I recommend here, to always create variable `params` and set the value of `'supportsAllDrives'` to `supports_all_drives`

lavigne958:
Thank you @yohayonyon for this contribution, could you please squash all your commits to only include the part that is fixed by your PR please ?
If you need any help doing this operation (as it can be dangerous) let me know I'll guide you.

lavigne958:
closes #778

lavigne958:
Re-created the pr in a simpler way, we can enable the sharedDrive option all the time it has no impact on regular drives and allows to create a spreadhseet in a shared drive.

Closing this PR in favor of #895

Issue #862, opened 2021-04-27
The function insert_row() ignores the value_input_option parameter and forces it to RAW


Targolo:
Hi,

**Steps to reproduce the behavior:**
    mysheet.insert_row(['=HYPERLINK("https://www.google.com","Google")'], value_input_option='USER_ENTERED')

**Expected behavior:**
   the cell content is a link to Google

**Current behavior:**
   the cell content is (with the ' at the beginning)
   `'=HYPERLINK("https://www.google.com","Google")`

We can clearly see in the code, in models.py, that insert_row() overwrites the value of value_input_option and forces it to RAW

**Environment info:**
- Operating System Linux
- Python version 3.8
- gspread version 3.7.0

Thanks

moonzlo:
To work with formulas, you need to use the method .update_acell

Targolo:
Hi,
You mean the _value_input_option_ parameter in insert_row() is useless?

The unique goal of this parameter is to work with formulas, so can you please tell me where it is written that I need to use update_acell(), and how to tell this function that the value is a formula since it does not have the _value_input_option_ parameter?

Thanks

moonzlo:
Hi @Targolo 

the fact is that due to the fact that this is an open source project, its documentation is far from perfect, but the developers have worked very hard, and provided the code with excellent documentation inside the library. All you need is to open the[ file](https://github.com/burnash/gspread/blob/810183bd5d4d441e9f4d797114d3f49b80939969/gspread/models.py) and read the docstring that was carefully left for us. There you can find all the necessary information on working with a particular method. My personal recommendation is how to update the row range immediately:


# Select a range
            cell_list = worksheet.range('A1:C7')

            for cell in cell_list:
                cell.value = 'O_o'

            # Update in batch
            worksheet.update_cells(cell_list,  value_input_option='USER_ENTERED')`

MikeSingularity:
@moonzlo The user @Targolo is attempting to insert rather than update cells and he is correct - there is a bug in the code, exactly as he has described.  I discovered exactly the same issue myself and came to make a bug report about it.

@Targolo as you've probably already discovered from examining the code, you can call insert_rows (wrapping your row in a list) instead and it will accept the value_input_option.

lavigne958:
closed by #873

PR #873, opened 2021-05-17
Fixed a bug with insert_row

elijabesu:
Regardless of what the end user put as `value_input_option`, it always resulted in "RAW".


lavigne958:
Thank you @elijabesu  for your fix, good catch.

Closes: #862

lavigne958 (approved):

Issue #874, opened 2021-05-17
Can we drop support for Python 2?


cclauss:
**Is your feature request related to a problem? Please describe.**
Can we drop [support for legacy Python](https://github.com/burnash/gspread/blob/master/setup.py#L50-L54) which died 500+ days ago?

Also, is the [development state still alpha](https://github.com/burnash/gspread/blob/master/setup.py#L61)?

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
https://devguide.python.org/#status-of-python-branches
https://devguide.python.org/devcycle/#end-of-life-branches

lavigne958:
Hi yes we will, a PR was open some time ago #733  and I tried to contact the owner of the PR if he/she can rebase it and refresh it.
I will wait 15 days (due date is 19th of may) and if no response then I will do it myself.

About the alpha state, I think not but this should be answered by the owner of this repository.

@burnash we should update the `Development Status` field to reflect the release, so far latest release is `3.7.0` so a coming PR should set its values to `3.7.0` and be changed when we release and place a tag.

What do you think ?

lavigne958:
Solved in #879 

Closing this issue.

PR #879, opened 2021-05-21
Remove python2

lavigne958:
The almighty long awaited PR that drops python2 supports is here ! :smile: 

Run the script `2to3` and fix most issues.
Not much to fix as the module already runs using python3.

Added python3.9 to the list of travis CI in case we add it back one day.
Added python3.9 to setup when building the package.


lavigne958:
Closes #874 
Closes #733 
Closes #879

PR #891, opened 2021-07-07
Improvement: Allow users to provide a custom credentials file to oauth()

### LINKED ISSUE #889: Parametrize not only `authorized_user_filename` but also `credentials_filename`
While instantiating a client using `gspread.auth.oauth()`, the latest release `v3.7.0` does not parameterize the filenames. However, in `master`, I see that `authorized_user.json` file is parametrized already. I think it's for the best the file used by auth flow, namely `credentials.json` is parametrized among `oauth()` inputs.

Allow the users to pass `credentials_filename` and use it during Google browser auth flow. Pass the file path to downstream functions that need it.

Not an alternative but maybe while there, it could be nice if `filename`s are renamed to `filepath`s as for instance `DEFAULT_CREDENTIALS_FILENAME` is not actually a file name but a file path.

slmtpz:
closes #889

### DIFF
--- a/docs/oauth2.rst
+++ b/docs/oauth2.rst
@@ -169,9 +169,12 @@ When you run this code, it launches a browser asking you for authentication. Fol
 You only need to do authorization in the browser once, following runs will reuse stored credentials.
 
 .. NOTE::
-    If you want to store the credentials file somewhere else, specify the path to `authorized_user.json` in :meth:`~gspread.oauth`:
+    If you want to store the credentials file somewhere else, specify the path to `credentials.json` and `authorized_user.json` in :meth:`~gspread.oauth`:
     ::
-        gc = gspread.oauth(authorized_user_filename='path/to/the/downloaded/file.json')
+        gc = gspread.oauth(
+            credentials_filename='path/to/the/credentials.json',
+            authorized_user_filename='path/to/the/authorized_user.json'
+        )
 
     Make sure you store the credentials file in a safe place.
 
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -57,13 +57,7 @@ DEFAULT_SERVICE_ACCOUNT_FILENAME = os.path.join(
 )
 
 
-def _create_installed_app_flow(scopes):
-    return InstalledAppFlow.from_client_secrets_file(
-        DEFAULT_CREDENTIALS_FILENAME, scopes
-    )
-
-
-def local_server_flow(scopes, port=0):
+def local_server_flow(scopes, port=0, filename=DEFAULT_CREDENTIALS_FILENAME):
     """Run an OAuth flow using a local server strategy.
 
     Creates an OAuth flow and runs `google_auth_oauthlib.flow.InstalledAppFlow.run_local_server <https://google-auth-oauthlib.readthedocs.io/en/latest/reference/google_auth_oauthlib.flow.html#google_auth_oauthlib.flow.InstalledAppFlow.run_local_server>`_.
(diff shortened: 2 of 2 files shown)


slmtpz:
One thing that bugs me a little is that the filename inputs of `load_credentials` and `store_credentials` have a default value in the function definition even though the default filename values are passed within `oauth` to these functions.
I introduced a new parameter `filename` for the flow functions, namely `local_server_flow` and `console_flow`. It doesn't have a default value in the function definitions as there is no need.

For consistency sake;
1- Should I drop the default values of `filename` from `load_credentials` and `store_credentials`? Are these considered a part of `gspread` API, would this be a breaking change?
2- Or, should I add a default value to `filename` in the flow functions?
3- Or, is this already good as is?

lavigne958:
> One thing that bugs me a little is that the filename inputs of `load_credentials` and `store_credentials` have a default value in the function definition even though the default filename values are passed within `oauth` to these functions.
> I introduced a new parameter `filename` for the flow functions, namely `local_server_flow` and `console_flow`. It doesn't have a default value in the function definitions as there is no need.
> 
> For consistency sake;
> 1- Should I drop the default values of `filename` from `load_credentials` and `store_credentials`? Are these considered a part of `gspread` API, would this be a breaking change?
> 2- Or, should I add a default value to `filename` in the flow functions?
> 3- Or, is this already good as is?

Hi @slmtpz, thank you very much for your contribution.

about the default values in `load_credentials` and `store_credentials` they are here in case someone want to write its own `oauth` scenario and use these functions, this way they are covered with the right default values.

1. to me, yes it could be braking changes, if someone has its own `oauth` function (for some reason, like doing thing in-between load, flow, store) then they assume the default value and use it this way. removing it will brake this code.
2. Again, to me, you should. this is open for discussion though. Like the above example it is the same reason. On the top of that before you PR anyone using GSpread used the default store location for these credentials by using the function `console_flow` or `local_flow` now it should work the same way, so the filename should have a default value again here using `DEFAULT_CREDENTIALS_FILENAME`
3. unfortunately not :slightly_smiling_face: 

Hope this answer your questions and help you move forward :wink: 

Apart from this, your PR is very welcome, I will run the test suite very soon on it.

slmtpz:
> > One thing that bugs me a little is that the filename inputs of `load_credentials` and `store_credentials` have a default value in the function definition even though the default filename values are passed within `oauth` to these functions.
> > I introduced a new parameter `filename` for the flow functions, namely `local_server_flow` and `console_flow`. It doesn't have a default value in the function definitions as there is no need.
> > For consistency sake;
> > 1- Should I drop the default values of `filename` from `load_credentials` and `store_credentials`? Are these considered a part of `gspread` API, would this be a breaking change?
> > 2- Or, should I add a default value to `filename` in the flow functions?
> > 3- Or, is this already good as is?
> 
> Hi @slmtpz, thank you very much for your contribution.
> 
> about the default values in `load_credentials` and `store_credentials` they are here in case someone want to write its own `oauth` scenario and use these functions, this way they are covered with the right default values.
> 
> 1. to me, yes it could be braking changes, if someone has its own `oauth` function (for some reason, like doing thing in-between load, flow, store) then they assume the default value and use it this way. removing it will brake this code.
> 2. Again, to me, you should. this is open for discussion though. Like the above example it is the same reason. On the top of that before you PR anyone using GSpread used the default store location for these credentials by using the function `console_flow` or `local_flow` now it should work the same way, so the filename should have a default value again here using `DEFAULT_CREDENTIALS_FILENAME`
> 3. unfortunately not 🙂
> 
> Hope this answer your questions and help you move forward 😉
> 
> Apart from this, your PR is very welcome, I will run the test suite very soon on it.

Thanks for your elaborate answer @lavigne958 ! I just updated the function definitions so that `filename`s have default value for the backward compatibility.

lavigne958 (approved):

PR #903, opened 2021-08-09
fix docstring and typo

rariyama:
## What I did
- I fixed the following typo:
    - spell of a word in a comment.
    - type of a return value.
        - a return value will be an integer, not a string.


burnash:
Thank you very much for fixing this!

PR #916, opened 2021-08-28
lint_python.yml: Try 'tox -e py' to test current Python

cclauss:
https://docs.github.com/en/actions/guides/building-and-testing-python#running-tests-with-tox

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -14,7 +14,7 @@ jobs:
                     --max-complexity=10 --max-line-length=255 --show-source --statistics .
       - run: isort --check-only --profile black . || true
       - run: pip install -r test-requirements.txt
-      - run: make test || true
+      - run: tox -e py || make test || true
       - run: mypy --ignore-missing-imports . || true
       - run: pytest . || true
       - run: pytest --doctest-modules . || true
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_python.yml
```diff
@@ -14,7 +14,7 @@ jobs:
                     --max-complexity=10 --max-line-length=255 --show-source --statistics .
       - run: isort --check-only --profile black . || true
       - run: pip install -r test-requirements.txt
-      - run: make test || true
+      - run: tox -e py || make test || true
```
lavigne958, line 17: you could remove the `make test` it is useless if we only use `tox`. I was thinking of removing the `Makefile` anyway.

lavigne958 (commented):
Nice improvement!

it would be better if (like the example) we set the python version list using:

```
strategy:
      matrix:
        python: [3.7, 3.8, 3.9]
```

then it would imply updating every mention of `python 3.6` or `python 3.4` to be removed (thinking mostly of setup.py etc)

.github/workflows/lint_python.yml
cclauss: Done.

lavigne958 (approved):
Great job.
Looks much better this way.

PR #917, opened 2021-08-28
Add more flake8 tests

cclauss:
Now that we have run `black` and `isort`, most flake8 tests pass.

This PR makes minor changes so that flake8 tests all pass except one that is incompatible with black.
* https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -15,8 +15,8 @@ jobs:
       - run: bandit --recursive --skip B105,B110,B311,B605,B607 .
       - run: black --check .
       - run: codespell  # --ignore-words-list="" --skip=""
-      - run: flake8 --ignore=E12,E20,E226,E231,E265,E271,E40,E731,F401,W29,W605
-                    --max-complexity=10 --max-line-length=255 --show-source --statistics .
+      - run: flake8 --ignore=E203 --max-complexity=10 --max-line-length=255
+                    --show-source --statistics .
       - run: isort --check-only --profile black .
       - run: pip install -r test-requirements.txt
       - run: tox -e py || true
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1371,9 +1371,14 @@ class Worksheet(object):
         cells = self._list_cells(values, in_row, in_column)
 
         if isinstance(query, str):
-            match = lambda x: x.value == query
+
+            def match(x):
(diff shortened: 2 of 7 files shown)


gspread/worksheet.py
```diff
@@ -1371,9 +1371,14 @@ def _finder(self, func, query, in_row=None, in_column=None):
         cells = self._list_cells(values, in_row, in_column)
 
         if isinstance(query, str):
-            match = lambda x: x.value == query
```
lavigne958, line 1374: Do you know why we can't have a lambda here? 🤔Is this from flake8?

.github/workflows/lint_python.yml
```diff
@@ -15,8 +15,8 @@ jobs:
       - run: bandit --recursive --skip B105,B110,B311,B605,B607 .
       - run: black --check .
       - run: codespell  # --ignore-words-list="" --skip=""
-      - run: flake8 --ignore=E12,E20,E226,E231,E265,E271,E40,E731,F401,W29,W605
-                    --max-complexity=10 --max-line-length=255 --show-source --statistics .
+      - run: flake8 --ignore=E203 --max-complexity=10 --max-line-length=255
```
lavigne958, line 18: Why is the error E203 still ignored?
I read it is the whitespace before `:`
https://www.flake8rules.com/rules/E203.html

lavigne958, line 18: We now use black which has a 88 max character line length. We could set the `max-line-length` to 88 too.

Unless it is applied to comments and doc etc 🤔

lavigne958 (commented):
Thanks for this PR.

I just have a few questions.

.github/workflows/lint_python.yml
cclauss: See the URL in the commit message.

cclauss: Comments.  If you lower the number, you will see which lines.

gspread/worksheet.py
cclauss: E731 | do not assign a lambda expression, use a def

https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes

https://www.python.org/dev/peps/pep-0008/#programming-recommendations

lavigne958 (approved):
Lgtm

PR #918, opened 2021-08-29
Upgrade to modern Python syntax

cclauss:
This pull request was created with the command [`pyupgrade --py3-only **/*.py`](https://github.com/asottile/pyupgrade#readme)

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,5 @@
 six
-nose  # nose2
+nose
 google-auth
 google-auth-oauthlib>=0.4.1
 betamax==0.8.1
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,5 @@
 six
-nose
+nose  # nose2
 google-auth
 google-auth-oauthlib>=0.4.1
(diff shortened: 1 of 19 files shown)


tox.ini
```diff
@@ -6,3 +6,4 @@ skip_missing_interpreters = true
 passenv = GS_RECORD_MODE GS_CREDS_FILENAME
 deps = -rtest-requirements.txt
 commands = nosetests -vv tests/
+# commands = nose2
```
lavigne958, line 9: Instead of migrating to `nose2` I'd rather update the test framework and use pytest.

more generaly speaking if this does not introduce anything we should not have it then.

test-requirements.txt
```diff
@@ -1,5 +1,5 @@
 six
-nose
+nose  # nose2
```
lavigne958, line 2: same here, like comment in tox.ini

cclauss:
`pytest tests/utils_test.py` is now a mandatory test.

lavigne958:
thanks @cclauss .

PR #925, opened 2021-09-07
Create default config directory if not exists

benhoman:
In store_credentials function, create the DEFAULT_CONFIG_DIR defined from the get_config_dir() function if it doesn't currently exist.


lavigne958 (approved):
thank you @benhoman , nice and simple.

PR #934, opened 2021-09-26
Update the link to the Google Developers Console

Croebh:
Currently, the link in the documentation on https://docs.gspread.org/en/latest/oauth2.html#using-signed-credentials links to:

https://console.cloud.google.com/project

However, that link gives you the following error:
> There is a temporary block on your account. This happens when Google detects requests from your network that may have been sent by malicious software, a browser plug-in, or script that sends automated requests.
>Retry in a few minutes.

By removing the `/project` from the URL, the link works and allows users to properly get to the console.


burnash:
Nice catch. Thank you very much!

PR #951, opened 2021-11-13
Implement `open` sheet within Drive folder

DBCerigo:
Enables passing `folder_id` to `open` to allow users to identify and open spreadsheet based on the spreadsheet name and ID of parent folder. This enables users to reuse duplicate spreadsheet names in different Drive folders and still be able to access spreadsheets using name.

(I implemented this as a patch with an open source project as we wanted the functionality already and using it successfully. [This](https://gitlab.com/howtobuildup/phoenix/-/merge_requests/129/diffs) is the PR for it, which includes integration style tests that make live calls to Drive and Sheets API that covers the new functionality.)

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -70,7 +70,7 @@ class Client:
         else:
             raise APIError(response)
 
-    def list_spreadsheet_files(self, title=None):
+    def list_spreadsheet_files(self, title=None, folder_id=None):
         files = []
         page_token = ""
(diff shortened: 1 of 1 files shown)


lavigne958 (approved):
Thank you @DBCerigo for this awesome contribution. 😊

You'll be notified once I release new version of gspread so you'll know when new release comes out.

PR #972, opened 2021-12-15
Аutomatic conversion of a cell letter to uppercase

Burovytskyi:
After upgrading to version 5.0.0, the work in my project broke due to the use of lower case.


lavigne958 (approved):
@Burovytskyi thank you very much for your contribution. 🙂

PR #977, opened 2021-12-25
Client Request Timeout param

Pxeba:
There's not much to say. I work a lot with spreadsheets and am getting error 408 more often, usually when I use col_values to get the last row.

I would like to have this option and ended up choosing to develop inside the library instead of using library + API. I did the tox tests and got only one warning. Added parameter description where I thought it needed.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -55,6 +55,7 @@ class Client:
         json=None,
         files=None,
         headers=None,
+        timeout=None
     ):
         response = getattr(self.session, method)(
             endpoint,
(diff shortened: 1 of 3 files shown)


lavigne958 (commented):
@Pxeba  thank you very much for this contribution !

I am wondering why, in the file `worksheet.py`, did you skip the methods `acell`, `cell`, `get_all_record`, `row_values` `batch_get`, `find` ?
You did not add the `timeout` parameter or these method rely on the `request` method.

PR #980, opened 2022-01-13
Update user-guide.rst

maky-hnou:
`cols` and `rows` shout be int not string in `sh.add_worksheet()`


lavigne958 (approved):
Thank you @maky-hnou  for this contribution.

PR #984, opened 2022-01-23
Feature/add header validation get all records

lavigne958:
1st commit:
Fix issue with tests cassettes.

2nd commit:
Add header validation for method `get_all_records`.
If duplicate headers exists raise exception.

Again, like any changes on test cassettes, skip the first commit to prevent your diff to show 1000s of lines.
Or use github to only show you the diff on `.py` files.


gspread/worksheet.py
```diff
@@ -402,6 +403,15 @@ def get_all_records(
...
         keys = data[idx]
 
+        # Check keys are uniques
+        # do not use 'headers' in return values
+        # it is not ordered
+        headers = set()
+        for key in keys:
+            if key in headers:
+                raise GSpreadException("headers must be uniques")
+            headers.add(key)
```
filip-31, line 413: If headers aren't used later I think this part could be simplified:
```
if len(keys) != len(set(keys)):
    raise GSpreadException("headers must be uniques")
```

lavigne958: thank you very much for you review, I really appreciate.
I update the code to use your proposal.

burnash (approved):

PR #985, opened 2022-01-30
add `in_shape`, `defualt` kwarg at WorkSheet::get method for resolving issue #947, 

### LINKED ISSUE #947: [Proposal]: get, get_values filling empty values response from Google Sheet (with expected cells dimension)
### Point of Contention (1 / 2)

I wrote replies at other issue [get_values does not return all data in some scenarios...](https://github.com/burnash/gspread/issues/910)

But I just want to share some opinions before making a PR.

```get()``` method
https://github.com/burnash/gspread/blob/6d01d7b5bb79601d0eece20255eabed9d13aa5bf/gspread/worksheet.py#L551 

and  ```get_values()``` method 
https://github.com/burnash/gspread/blob/6d01d7b5bb79601d0eece20255eabed9d13aa5bf/gspread/worksheet.py#L254 

has no difference but returning value after wrapping with util function ```fill_gaps()```
https://github.com/burnash/gspread/blob/6d01d7b5bb79601d0eece20255eabed9d13aa5bf/gspread/utils.py#L424

**so I thought "is it really necessary to separate those two methods?"**

Also in Documentation,

[Worksheet.get](https://docs.gspread.org/en/latest/api.html#gspread.worksheet.Worksheet.get)

[Workshhet.get_values](https://docs.gspread.org/en/latest/api.html#gspread.worksheet.Worksheet.get_values)

has no difference in explanation, because the only difference is, as I mentioned above,

```get_values()``` returns value got from ```get()``` after wrapping with ```fill_gaps()```

https://github.com/burnash/gspread/blob/6d01d7b5bb79601d0eece20255eabed9d13aa5bf/gspread/worksheet.py#L320

I think it's better to integrate by separating the slight difference by **kwargs** and it could be more clear

---------------

### Point of Contention (2 / 2)

Just like issue #910, when people use method like

```python
values = MyWorksheet.get('A1:C2')
# or
values = MyWorksheet.get_values('A1:C2')
```

people those who are first to use this method would have expected 
2-D list A1 to C2, even when every cell is empty 

just like this:

```python
# values
[
   [ None, None, None ],
   [ None, None, None ]
]
```

for more clear examples, see my explanation from issue comment :  https://github.com/burnash/gspread/issues/910#issuecomment-952612525 

I know google send just empty value, literally no values,
when cells in range is empty like this

https://github.com/burnash/gspread/blob/6d01d7b5bb79601d0eece20255eabed9d13aa5bf/gspread/worksheet.py#L595

```python
print(response)
# {'range': 'test!A2:C2', 'majorDimension': 'ROWS'}
```

This could be handled like this [from my comment](https://github.com/burnash/gspread/issues/910#issuecomment-953482606)

```python
def getMember(self, name: str, replace_empty=None) -> Member:
        result: Cell = self.memberSheet.find(name, in_column=USER_NAME_COL)

        if not result or not result.value == name:
            return MEMBER_UNKNOWN

        # y, x
        start = rowcol_to_a1(result.row, result.col - 1)
        end = rowcol_to_a1(result.row, result.col - 1 + MEMBER_INFO_width - 1)

        # edit : in this case, it's 1-D list, see that I used [0] indexing below
        # expected column length
        data = [replace_empty  for _ in range(MEMBER_INFO_width + 1)]

        reply = self.memberSheet.get_values(
            f'{start}:{end}',
            value_render_option='UNFORMATTED_VALUE'
        )[0]

        emptyFilled = [val if val else dafault for val, dafault in zip_longest(
            reply, data, fillvalue=replace_empty)]
        return emptyFilled
```

------------------------------

## Conclusion

My proposals are...

1. integrate ```get()``` and ```get_value()``` to ```get()```

add kwargs like **maintain_shape** to get expected dimension from given range

```python
result = worksheet.get('A1:D2')

# result
[
 ['', 123 ] ,
 [ '',   ''  , 42]
]

# result(after fix), default = None
result = worksheet.get('A1:D2', maintain_shape = True)

# result
[
 [ '', 123, '',''],
 [ '', '', 42, '']
]
```

2. add another kwargs option(default value) to fill empty values

```python
result = worksheet.get('A1:D2', maintain_shape = True, default = None)
# result(after fix), default = None
[
 [ None, 123, None ,None],
 [ None, None , 42, None]
]
```

ccppoo:
closes #947

method `_inflate` is for literally inflating json-body value got from Google spreadsheet API

by using `_inflate` it's possible to do the same for `batch_get` method,

but I'm making PR first then some one could do it instead of me ...

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -249,7 +249,7 @@ def a1_to_rowcol(label):
 
         col = 0
         for i, c in enumerate(reversed(column_label)):
-            col += (ord(c) - MAGIC_NUMBER) * (26 ** i)
+            col += (ord(c) - MAGIC_NUMBER) * (26**i)
     else:
         raise IncorrectCellLabel(label)
 
@@ -302,7 +302,7 @@ def _a1_to_rowcol_unbounded(label):
         if column_label:
             col = 0
             for i, c in enumerate(reversed(column_label.upper())):
-                col += (ord(c) - MAGIC_NUMBER) * (26 ** i)
+                col += (ord(c) - MAGIC_NUMBER) * (26**i)
         else:
             col = inf
 
@@ -506,8 +506,8 @@ def absolute_range_name(sheet_name, range_name=None):
 
 
 def absolute_range_to_grid_range(range_name):
-    _, a1_range = range_name.split('!')
-    
+    _, a1_range = range_name.split("!")
+
     return a1_range_to_grid_range(a1_range)
 
 
@@ -608,8 +608,6 @@ def accepted_kwargs(**default_kwargs):
         return wrapper
 
     return decorate
-
-
 
 
 if __name__ == "__main__":
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -505,6 +505,12 @@ def absolute_range_name(sheet_name, range_name=None):
         return sheet_name
 
 
+def absolute_range_to_grid_range(range_name):
+    _, a1_range = range_name.split('!')
+    
+    return a1_range_to_grid_range(a1_range)
+
+
 def is_scalar(x):
     """Return True if the value is scalar.
 
@@ -602,6 +608,8 @@ def accepted_kwargs(**default_kwargs):
         return wrapper
 
     return decorate
+
+
 
 
 if __name__ == "__main__":
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -6,6 +6,8 @@ This module contains common worksheets' models.
 
 """
 
+from itertools import zip_longest
+
 from .cell import Cell
 from .exceptions import GSpreadException
 from .urls import SPREADSHEET_URL, WORKSHEET_DRIVE_URL
@@ -16,6 +18,7 @@ from .utils import (
     a1_range_to_grid_range,
     a1_to_rowcol,
     absolute_range_name,
+    absolute_range_to_grid_range,
     accepted_kwargs,
     cast_to_a1_notation,
     cell_list_to_rect,
@@ -569,6 +572,8 @@ class Worksheet:
         major_dimension=None,
         value_render_option=None,
         date_time_render_option=None,
+        default = '',
+        in_shape = False
     )
     def get(self, range_name=None, **kwargs):
         """Reads values of a single range or a cell of a sheet.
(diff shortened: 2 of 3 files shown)


ccppoo:
I have no idea whats wrong with linting, I'm getting nothing except venv package files and black is telling me All done
Help

gspread/utils.py
```diff
@@ -249,7 +249,7 @@ def a1_to_rowcol(label):
 
         col = 0
         for i, c in enumerate(reversed(column_label)):
-            col += (ord(c) - MAGIC_NUMBER) * (26 ** i)
+            col += (ord(c) - MAGIC_NUMBER) * (26**i)
```
lavigne958, line 252: This is your issue with the linter.

use the following command to format your code (you can refer to the [CONTRIBUTION_GUIDELINES](https://github.com/burnash/gspread/blob/master/.github/CONTRIBUTING.md)):
`tox -e format` it will format your code so it will pass the linter.

gspread/utils.py
```diff
@@ -302,7 +302,7 @@ def _a1_to_rowcol_unbounded(label):
         if column_label:
             col = 0
             for i, c in enumerate(reversed(column_label.upper())):
-                col += (ord(c) - MAGIC_NUMBER) * (26 ** i)
+                col += (ord(c) - MAGIC_NUMBER) * (26**i)
```
lavigne958, line 305: same here for the code format.

lavigne958 (commented):
Thank you @ccppoo for this contribution this is great work but I don't understand why you added a new method `_inflate`.
we already have a function that does the same job in `gspread/utils.py:fill_gaps`.
In that function we don't handle default values, so you simply need to add default values to it and you should be good to go.

remember that we never change what the API returns us, we take the values and create new structures but never change the original structure.

gspread/utils.py
ccppoo: I don't get it, using `black` give me that using `tox -e format` also give me that. No difference doing in venv
I think I have to drop out from this

ccppoo:
can't help myself with linting anymore, sorry

gspread/utils.py
lavigne958: Don't worry mate. we can solve this later.

Keep your PR open and let's focus on the design you choose. I ça handle the format later.

lavigne958:
No worries. Did you think about what I said above ? What's your opinion on updzting the function `fill_gaps` ?

Issue #1005, opened 2022-03-07
Add ability to group columns/rows


juliusdanek:
**Is your feature request related to a problem? Please describe.**
It would be great if you were able to natively group columns and/or rows

This is relatively easy to achieve with batch_update. See below for the request that I am using. 

```python 
    request = {
        "addDimensionGroup": {
            "range": {
                "sheetId": tab_id,
                "dimension": dimension,
                "startIndex": start_index,
                "endIndex": end_index,
            },
        }
    }```

lavigne958:
Hi @juliusdanek 

I just tried it, I did not know about that feature, this is very nice !

I can certainly add this in the next release (will be 5.3.0)

juliusdanek:
Awesome! Thank you

PR #1010, opened 2022-03-09
Add new methods to add/list/delete dimensionGroups

lavigne958:
Add new methods to work over `dimensionGroup`.
Possible actions are:
  - add new group
  - list groups
  - delete a group

 closes https://github.com/burnash/gspread/issues/1005

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -131,6 +131,15 @@ class Worksheet:
         """Number of frozen columns."""
         return self._properties["gridProperties"].get("frozenColumnCount", 0)
 
+    def _get_sheet_property(self, property, default_value):
+        """return a property of this worksheet or default value if not found"""
+        meta = self.spreadsheet.fetch_sheet_metadata()
+        sheet = finditem(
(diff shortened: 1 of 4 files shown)


gspread/worksheet.py
```diff
@@ -1824,3 +1833,129 @@ def delete_named_range(self, named_range_id):
...
+            range can be found here
+
+            https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/request#DeleteDimensionGroupRequest
+
+        :param int start: The start (inclusive) of the group
+        :param int end: The end (exclusive) of the group
+        """
+        return self._delete_dimension_group(start, end, Dimension.cols)
+
+    def delete_grouped_rows(self, start, end):
```
juliusdanek, line 1930: @lavigne958 maybe this could be named this "delete_row_grouping"? If it's named "delete_grouped_rows" I would assume it deletes group **and the rows**, not just the group. 

Similar for columns.

lavigne958: I will go for the `Dimension` term exactly like in the API that will it simpler for users to expect the right behavior.

PR #1012, opened 2022-03-18
Add method to hide rows/columns

### LINKED ISSUE #1006: Add ability to hide columns and/or rows
Add ability to hide columns and/or rows. This is achievable via batch_update using the request body below.

```   request = {
        "updateDimensionProperties": {
            "range": {
                "sheetId": tab_id,
                "dimension": dimension,
                "startIndex": start_index,
                "endIndex": end_index,
            },
            "properties": {
                "hiddenByUser": True,
            },
            "fields": "hiddenByUser",
        }
    }```

lavigne958:
Add new method that allows the user to explicitly hide rows or columns.
Ass new method that allows the user to explicitly unhide rows or
columns.

This feature does not include any _listing_ method, because the listing
is not in the metadata of the `SpreadSheet` or the `Sheet`, it will
require to list all the data in the sheet and list all of them
then check if the flag `hiddenByUser` is set. This is a lot of
processing and with very large data set this can take some time
For a result that is only visible using the UI.

closes #1006

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1959,3 +1959,111 @@ class Worksheet:
         :rtype: list
         """
         return self._get_sheet_property("rowGroups", [])
+
+    def _hide_dimension(self, start, end, dimension):
+        """
+        update this sheet by hiding the given 'dimension'
+
+        Index start from 0.
+
+        :param int start: The (inclusive) start of the dimension to hide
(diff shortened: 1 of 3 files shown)


lavigne958:
@juliusdanek what do you think ?

juliusdanek:
Looks excellent! Thank you sir.

PR #1019, opened 2022-04-05
added 'AppAssertionCredentials' to list of valid oauth2 names

jaceiverson:
It appears that Google Colab has changed how it supplies default credentials using the GoogleCredentials.get_application_default(). Now when you authorize through colab--google.colab.auth--the class name (GoogleCredentials.get_application_default().__class__.__name__) is titled AppAssertionCredentials. This is not a valid type according to gspread. I added this in the valid names for the class so that it would work for Google Colab's base form of authentication. If you try to use colab's current form of auth with the current state of gspread, the following TypeError is raised.

    raise TypeError(
        "Credentials need to be from either oauth2client or from google-auth."
    )

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -50,6 +50,7 @@ def convert_credentials(credentials):
         "OAuth2Credentials",
         "AccessTokenCredentials",
         "GoogleCredentials",
+        "AppAssertionCredentials",
     ):
         return _convert_oauth(credentials)
     elif isinstance(credentials, Credentials):


lavigne958 (approved):
thanks for this PR, this is a good catch I dd not think it could simply come from our restricted list of authorized auth classes.

lavigne958:
unfortunately you must rebase your branch on the laster `master` branch. I just updated it to fix our CI issue.

jaceiverson:
After further testing it appears that the reason this wasn't authenticating is due to an antiquated oauth authentication process. So rather than adapt to the 'AppAssertionCredentials', if you simply use the below method authentication works fine. 
```
from google.colab import auth
auth.authenticate_user()
import gspread
from google.auth import default
creds, _ = default()
gc = gspread.authorize(creds)
```

Feel Free to close this PR out. Appreciate you taking the time to look at it.

PR #1021, opened 2022-04-06
Bugfix/get all record duplicated columns

### LINKED ISSUE #1007: Duplicate header check in 5.2.0 is not backward compatible
**Describe the bug**

A spreadsheet with multiple columns that had a blank header used to load using `get_all_records` before 5.2.0, but it now fails with "headers must be uniques" exception. I presume, but did not confirm, that it is due to this simplification: https://github.com/burnash/gspread/commit/c8a5a7350c40498cf38d3c4a27c748100632804a

**To Reproduce**
1. Run  `get_all_records` on a spreadsheet with multiple columns with a blank header.
2. See error "headers must be uniques".

**Expected behavior**
This should work as it used to without an error. 

**Environment info:**
- Operating System [e.g. Linux, Windows, macOS]: macOS
- Python version: 3.8
- gspread version: 5.2.0

  File "/edx/other/edx-repo-health/repo_health/check_ownership.py", line 79, in check_ownership
    records = find_worksheet(google_creds_file, spreadsheet_url, worksheet_id)
  File "/edx/other/edx-repo-health/repo_health/check_ownership.py", line 44, in find_worksheet
    return worksheet.get_all_records()
  File "/edx/venvs/edx-repo-health/lib/python3.8/site-packages/gspread/worksheet.py", line 408, in get_all_records
    raise GSpreadException("headers must be uniques")
gspread.exceptions.GSpreadException: headers must be uniques

lavigne958:
Add a new argument to `get_all_records` to provide the list of expected
headers.

The given expected headers must:
- be unique
- be part of the complete headers list
- must not contain extra headers

This will provide a way for users to use this method and still
have *some* duplicated headers that are not relevant to pull.

This will ensure the columns that matters have unique headers.

Closes #1007

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -376,6 +376,7 @@ class Worksheet:
         allow_underscores_in_numeric_literals=False,
         numericise_ignore=None,
         value_render_option=None,
+        expected_headers=None,
     ):
         """Returns a list of dictionaries, all of them having the contents of
         the spreadsheet with the head row as keys and each of these
@@ -399,6 +400,7 @@ class Worksheet:
         :param str value_render_option: (optional) Determines how values should
             be rendered in the the output. See `ValueRenderOption`_ in
             the Sheets API.
+        :param list expected_headers: (optional) List of expected headers, they must be unique.
 
         .. _ValueRenderOption: https://developers.google.com/sheets/api/reference/rest/v4/ValueRenderOption
         """
@@ -412,9 +414,27 @@ class Worksheet:
 
         keys = data[idx]
 
-        # Check keys are uniques
-        if len(keys) != len(set(keys)):
-            raise GSpreadException("headers must be uniques")
+        # if no given expected headers, expect all of them
+        if expected_headers is None:
+            expected_headers = keys
+
+        # keys must:
+        # - be uniques
+        # - be part of the complete header list
+        # - not contain extra headers
+        expected = set(expected_headers)
+        headers = set(keys)
+
+        # make sure they are uniques
+        if len(expected) != len(expected_headers):
+            raise GSpreadException("the given 'expected_headers' are not uniques")
+
+        if not expected & headers == expected:
+            raise GSpreadException(
+                "the given 'expected_headers' contains unknown headers: {}".format(
+                    expected & headers
+                )
+            )
 
         if numericise_ignore == ["all"]:
             values = data[idx + 1 :]
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -400,9 +400,17 @@ class Worksheet:
         :param str value_render_option: (optional) Determines how values should
             be rendered in the the output. See `ValueRenderOption`_ in
             the Sheets API.
+
+            .. note::
+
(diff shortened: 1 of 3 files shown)


gspread/worksheet.py
```diff
@@ -399,6 +400,7 @@ def get_all_records(
         :param str value_render_option: (optional) Determines how values should
             be rendered in the the output. See `ValueRenderOption`_ in
             the Sheets API.
+        :param list expected_headers: (optional) List of expected headers, they must be unique.
```
robrap, line 403: Consider adding something like: `Dictionaries will be returned for all headers, even if not included in this list.`

gspread/worksheet.py
```diff
@@ -412,9 +414,27 @@ def get_all_records(
...
+        # make sure they are uniques
+        if len(expected) != len(expected_headers):
+            raise GSpreadException("the given 'expected_headers' are not uniques")
+
+        if not expected & headers == expected:
+            raise GSpreadException(
+                "the given 'expected_headers' contains unknown headers: {}".format(
+                    expected & headers
+                )
+            )
```
robrap, line 437: Consider test coverage to prove the logic is and remains correct.

robrap, line 437: Consider test coverage to prove the logic is and remains correct.

robrap (approved):
Thanks. This would be great. Added some comments for you to consider, but nothing blocking this.

robrap (commented):
I forgot one more comment.

lavigne958:
> Thanks. This would be great. Added some comments for you to consider, but nothing blocking this.

Thank you for your review, it is much appreciated.

You're right my first test scenario was too simple, I improved it to check the new logic and added a note to mention all the headers are returned.

lavigne958:
> I forgot one more comment.

you're right, this was probably an old comment and the code part was copied over and over with time. I cleaned it.

tests/worksheet_test.py
```diff
@@ -629,6 +629,45 @@ def test_get_all_records_duplicate_keys(self):
...
+        rows = [
+            ["A1", "B2", "C3", "D4"],
+            [1, "b2", 1.45, ""],
+            ["", "", "", ""],
+            ["A4", 0.4, "", 4],
+        ]
+        cell_list = self.sheet.range("A1:D4")
+        for cell, value in zip(cell_list, itertools.chain(*rows)):
+            cell.value = value
+        self.sheet.update_cells(cell_list)
```
robrap, line 646: [idea] The tests might be simpler to read if this were moved to a helper method like `update_sheet_cell_values` (or something like that).

robrap (approved):
Nice. Thanks. I added one more idea to consider for the future.

tests/worksheet_test.py
lavigne958: you're right but this will be done during some refactoring of the test suite. with generic helper functions to generate data etc.

thank you for you feedback.

robrap:
Also, in case this helps anyone, this is still a backward incompatible change for some. If you newly run into an issue, you’ll need to update your code to pass `expected_headers` with a list of any of the headers you expect. I’m not sure if sending an empty list would provide the legacy behavior.

PR #1036, opened 2022-04-24
feat: Download PDF from Spreadsheet #1035

### LINKED ISSUE #1035: Download PDF from Spreadsheet
Lovely library.

I was wondering if it could maybe somehow be possible with the Google API to use the "Download -> PDF" function from the file menu in google sheets. This would be simply great to use a spread sheet as a template for table data. You could exchange the template data with real data and generate a PDF document for sending it away. Just dreaming ....

If this is not possible to do with the google API then delete this feature request.

best regards

Chris

100paperkite:
Add feature to export a spreadsheet to a PDF

And I have some questions about...
- Supporting to export a spreadsheet to other types of files (for all file types listed in Google API docs) 
- Writing test for exporting ( How should I write test for export? )

closes #1035

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -213,7 +213,7 @@ class Client:
         spreadsheet_id = r.json()["id"]
         return self.open_by_key(spreadsheet_id)
 
-    def export(self, file_id, format):
+    def export(self, file_id, format=ExportFormat.PDF):
         """Export the spreadsheet in the format.
 
         :param str file_id: A key of a spreadsheet to export
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -207,6 +207,32 @@ class Client:
         spreadsheet_id = r.json()["id"]
         return self.open_by_key(spreadsheet_id)
 
+    def export_to_pdf(self, file_id):
+        """Export a spreadsheet to a PDF.
+
+        :param str file_id: A key of a spreadsheet to export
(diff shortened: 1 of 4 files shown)


gspread/client.py
```diff
@@ -207,31 +213,35 @@ def create(self, title, folder_id=None):
         spreadsheet_id = r.json()["id"]
         return self.open_by_key(spreadsheet_id)
 
-    def export_to_pdf(self, file_id):
-        """Export a spreadsheet to a PDF.
+    def export(self, file_id, format):
```
lavigne958, line 216: please set the default value for `format` to `ExportFormat.PDF`, this is what was requested at first by the user and will be most likely the requested format.

gspread/spreadsheet.py
```diff
@@ -518,12 +518,26 @@ def share(
             with_link=with_link,
         )
 
-    def export_to_pdf(self):
-        """Export the spreadsheet as a pdf.
+    def export(self, format):
```
lavigne958, line 521: please set the default value for `format` to `ExportFormat.PDF`, this is what was requested at first by the user and will be most likely the requested format.

lavigne958 (commented):
Thank you very much for this improvement. 
There is only a minor detail to change then this is ready to merge.

100paperkite:
Thanks for the review! I updated PR

PR #1063, opened 2022-06-15
Implement __eq__ method for `Cell`

### LINKED ISSUE #1061: Implement equality on `Cell` objects
When writing tests for code that makes use of `gspread`, we found a pretty tedious task to compare `Cell` objects, since there's no way to directly compare one instance to another.

Be able to comparte `Cell` objects instances (e.g. `cell == another_cell`). Taking a look at the code, two `Cell` instances should be equal when they are placed in the same `row` and `col` and contain the same `value`, so it should be easy to implement the corresponding `__eq__` method.

Right now, as a workaround, we're doing assertions on each `Cell` instance attributes (e.g. `cell.row == another_cell.row and cell.col == another_cell.col and cell.value == another_cell.value`).

**Additional context**
I've seen that this feature was already requested [here](https://github.com/burnash/gspread/issues/360), but was closed without an answer to this request, therefore I decided to open a specific feature request for it.

chisvi:
As suggested in https://github.com/burnash/gspread/issues/1061, the code included in this PR implements the `__eq__` method for `Cell` class.

This way, it will be possible to compare two `Cell` instances (`cell_1 == cell_2`).

Two cell instances are considered "equal" when they represent the cell at the same `row` and  `col` and contain the same `value`.

This feature is particullary useful for testing purposes.

closes #1061

### DIFF
--- a/tests/cassettes/CellTest.test_equality.json
+++ b/tests/cassettes/CellTest.test_equality.json
@@ -4,7 +4,7 @@
         {
             "request": {
                 "method": "POST",
-                "uri": "https://localhost:3128/drive/v3/files?supportsAllDrives=True",
+                "uri": "https://www.googleapis.com/drive/v3/files?supportsAllDrives=True",
                 "body": "{\"name\": \"Test CellTest test_equality\", \"mimeType\": \"application/vnd.google-apps.spreadsheet\"}",
                 "headers": {
                     "User-Agent": [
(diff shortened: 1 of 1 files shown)


lavigne958 (commented):
Hi thank you for this pr.

I notice you added a test but I don't see the test cassette. Did you record your test ? Does it run successfully?

lavigne958:
> > Hi thank you for this pr.
> > I notice you added a test but I don't see the test cassette. Did you record your test ? Does it run successfully?
> > I notice your PR does not pass the CI, you can follow the contribution guidelines (can be found at the bottom bof the readme of the repo) you'll find all you need to contribute and make your PR pass the CI
> 
> Hi @lavigne958! Yes, I've recorded my test but didn't provide the cassete, sorry. Regarding this I've a couple of questions:
> 

> * Should I include in my PR the test cassete for all tests, or just the one that has been created for my new test?
yes you should, but regarding the situation, you added a new test, so you should only have a single new file for your new test.
When recording what you can do is:
1. discard any changes on your cassettes: `git restore --stagged tests/cassettes/` then `git restore tests/cassettes/`
2. start the test suite with the env variables set to record new episodes: `GS_RECORD_MODE=new_episodes tox -e py`
3. then you should only have a single new file for your new test.

> * Also I see that the cassete contains sensible info like the Bearer token used to authenticate the request, so I'm not sure how to proceed to avoid to push it.
yes but they are removed when recording the cassettes, see [here](https://github.com/burnash/gspread/blob/4fcad2ad3ebea34737f631d53ef3ce0ca719e35e/tests/conftest.py#L59-L61)

> 
> Let me check the CI and fix whatever is broken (weird, since I ran all the checks locally and were ok).

you can find a `tox` command that will format your code: `tox -e format`.

chisvi:
I've just added the cassette, thanks for your help!

Regarding the formatter, after running it with `tox -e format` it left the code in an status the linter was complaining about. I had to set some variable to avoid having multiline comparisons and therefore avoid linting issues.
imagen

gspread/cell.py
```diff
@@ -40,6 +40,12 @@ def __repr__(self):
             repr(self.value),
         )
 
+    def __eq__(self, other):
+        same_row = self.row == other.row
+        same_col = self.col == other.col
+        same_value = self.value == other.value
```
chisvi, line 46: I had to set these variables to avoid linting + format issues ([W503](https://www.flake8rules.com/rules/W503.html))

lavigne958: that's fine.

lavigne958 (approved):

lavigne958:
> I've just added the cassette, thanks for your help!
> 
> Regarding the formatter, after running it with `tox -e format` it left the code in an status the linter was complaining about. I had to set some variable to avoid having multiline comparisons and therefore avoid linting issues. imagen

Did you understand what the error message is about ? because the changes you made are not exactly what the linter is asking you to fix, but because you break it into multiple variables it fixes the warning from the linter (which is better in my opinion).

chisvi:
> > I've just added the cassette, thanks for your help!
> > Regarding the formatter, after running it with `tox -e format` it left the code in an status the linter was complaining about. I had to set some variable to avoid having multiline comparisons and therefore avoid linting issues. imagen
> 
> Did you understand what the error message is about ? because the changes you made are not exactly what the linter is asking you to fix, but because you break it into multiple variables it fixes the warning from the linter (which is better in my opinion).

I think so: it's complaining about putting a line break before binary operator (`and`). My first try was to put it *after* the binary operator, but then the linter was complaining about the format (which was putting it again before the binary operator), so I decided to change it to that final version.

PR #1065, opened 2022-06-19
Change tab color

### LINKED ISSUE #996: add method to change tab color
## Overview

add method to change the color of a tab

## Details

The api provide a way to change the color of a worksheet tab: [sheet API v4](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/sheets#SheetProperties)

expose necessary methods to change that tab color.

cgkoutzigiannis:
Closes #996

Exposed a new ```tab_color_style``` property to retrieve current color  and a new ```update_tab_color``` function to change the color.
I tested locally and it does work, however I am really struggling with writing the test case for it.

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -181,7 +181,7 @@ class Worksheet:
         return self._properties["gridProperties"].get("frozenColumnCount", 0)
 
     @property
-    def tab_color_style(self):
+    def tab_color(self):
         """Tab color style."""
         return self._properties["tabColorStyle"]["rgbColor"]
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1162,6 +1162,41 @@ class Worksheet:
         self._properties["title"] = title
         return response
 
+    def update_tab_color(self, color):
+        """Changes the worksheet's tab color.
+
+        :param dict color: The red, green and blue values of the color, between 0 and 1.
+        """
+        body = {
+            "requests": [
+                {
+                    "updateSheetProperties": {
+                        "properties": {
+                            "sheetId": self.id,
+                            "tabColor": {
+                                "red": color["red"],
+                                "green": color["green"],
+                                "blue": color["blue"],
+                            },
+                            "tabColorStyle": {
+                                "rgbColor": {
+                                    "red": color["red"],
+                                    "green": color["green"],
+                                    "blue": color["blue"],
+                                }
+                            },
+                        },
+                        "fields": "tabColor,tabColorStyle",
+                    }
+                }
+            ]
+        }
+
+        response = self.spreadsheet.batch_update(body)
+        self._properties["tabColorStyle"]["rgbColor"] = color
+        self._properties["tabColor"] = color
+        return response
+
     def update_index(self, index):
         """Updates the ``index`` property for the worksheet.
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -180,6 +180,11 @@ class Worksheet:
         """Number of frozen columns."""
         return self._properties["gridProperties"].get("frozenColumnCount", 0)
 
+    @property
+    def tab_color_style(self):
+        """Tab color style."""
+        return self._properties["tabColorStyle"]["rgbColor"]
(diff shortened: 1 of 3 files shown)


gspread/worksheet.py
```diff
@@ -1157,6 +1162,41 @@ def update_title(self, title):
...
+                        },
+                        "fields": "tabColor,tabColorStyle",
+                    }
+                }
+            ]
+        }
+
+        response = self.spreadsheet.batch_update(body)
+        self._properties["tabColorStyle"]["rgbColor"] = color
+        self._properties["tabColor"] = color
```
lavigne958, line 1197: This is very good. :+1:

gspread/worksheet.py
```diff
@@ -1157,6 +1162,41 @@ def update_title(self, title):
...
+                            },
+                            "tabColorStyle": {
+                                "rgbColor": {
+                                    "red": color["red"],
+                                    "green": color["green"],
+                                    "blue": color["blue"],
+                                }
+                            },
+                        },
+                        "fields": "tabColor,tabColorStyle",
```
lavigne958, line 1189: You said you ran some manual tests and it works ? Last time I tried to do it, here in the `fields`I had to put every single entry that is updated, so `tabColor.red,tabColor.green,tabColor.blue,tabColorStyle.rgbColor.red.....`

If it works likes this then it's better, simpler to read. :slightly_smiling_face:

gspread/worksheet.py
```diff
@@ -180,6 +180,11 @@ def frozen_col_count(self):
         """Number of frozen columns."""
         return self._properties["gridProperties"].get("frozenColumnCount", 0)
 
+    @property
+    def tab_color_style(self):
+        """Tab color style."""
+        return self._properties["tabColorStyle"]["rgbColor"]
```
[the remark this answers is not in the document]
cgkoutzigiannis: I wasn't so sure about the naming myself, so I think you're right. It makes send to keep it simple, since the API is kind of using both atm

lavigne958:
> Regarding the test, I was able to run the existing test suite and nothing is broken. I am not sure how I should go about writing a new test case for the `update_tab_color` method.

About the test I would suggest first to write a test case that ensure the functionality works, so something like this would be a good start:

#### test_get_set_tab_color
1. get the current tab color (we don't know what the API returns if no color is set, this is good thing to check)
2. set the tab to a different color
3. get the color again (we will need here to update the `(spread)?sheet` metadata I suppose)
4. assert that both colors are different

cgkoutzigiannis:
You're right, the API doesn't return the tab color if no tab color is set. However I think we should handle this case by writing some code to retrieve it, if someone calls the ```tab_color``` property. I am not so sure on how we could retrieve the tab color from the API.

lavigne958:
I noticed I did not completely reply to your question.

If the API does return z color we can't do anything.
What we can do is check of we have other properties in the `self._properties` dict. If so then we can't do much, if we don't then it means we might need to rework how we retrieve the properties (which I can handle if you need).

lavigne958:
@cgkoutzigiannis do you wish to write the test scenario or should I do it ? either ways are fine to me just let me know.

cgkoutzigiannis:
I am sorry for the delay, I have been trying to get the test to work, but it's throwing an error:
```gspread.exceptions.APIError: {'code': 400, 'message': 'Invalid requests[0].updateSheetProperties: The sheet name cannot be empty.', 'status': 'INVALID_ARGUMENT'}```
I can push the code, if you want  to take a look yourself.

lavigne958:
Yes please do so, and we'll look at it here in the pull request.

lavigne958 (commented):
This looks good to me. You just have to format your code and then we can merge it.

As mentioned in contributing guidelines you can use the command `tox -e format`.

lavigne958:
Hi would like me to fix the code format and merge this for you ? or would you like ot fix it and push your last commit with the fix then we can mere this PR ? :thinking:

lavigne958:
closing this PR in favor of #1095  due to the original PR owner who has left.

PR #1080, opened 2022-07-10
Fix grammar in docstring for client.export

dgilman:
Fix use of English articles


lavigne958 (approved):
Hey thanks again or this one.

We have `codespell` running but it can't catch everything.

PR #1081, opened 2022-07-10
Fix docstring name of named_range() param

dgilman:


lavigne958 (approved):
Hi @dgilman  thank you for this contribution ! :+1:

PR #1083, opened 2022-07-10
Many typo fixes to worksheet docstrings

dgilman:


gspread/worksheet.py
```diff
@@ -2095,8 +2095,8 @@ def add_dimension_group_columns(self, start, end):
 
         .. note::
 
-            API behavior with nested groups and non matching ``[start:end[``
-            range can be found here `Add Dimension Group Request`_
+            API behavior with nested groups and non-matching ``[start:end]``
+            range can be found here: `Add Dimension Group Request`_
```
lavigne958, line 2099: the open square-brackets at the end is a mathematical notation to mention that the end index is not included.

I learned at school to write it like this but it seems from what I find online the general way to write it would be: `[start:end)`
If you have time may be you could update it in your PR using round-brackets on each `end` occurence please, or I'll do later otherwise.

gspread/worksheet.py
```diff
@@ -2111,7 +2111,7 @@ def add_dimension_group_rows(self, start, end):
 
         .. note::
 
-            API behavior with nested groups and non matching ``[start:end[``
+            API behavior with nested groups and non-matching ``[start:end]``
```
lavigne958, line 2114: same here for the range.

gspread/worksheet.py
```diff
@@ -2140,11 +2140,11 @@ def _delete_dimension_group(self, start, end, dimension):
 
     def delete_dimension_group_columns(self, start, end):
         """
-        Delete a grouped columns
+        Remove the grouping of a set of columns.
 
         .. note::
 
-            API behavior with nested groups and non matching ``[start:end[``
+            API behavior with nested groups and non-matching ``[start:end]``
```
lavigne958, line 2147: same here for the range.

gspread/worksheet.py
```diff
@@ -2156,10 +2156,10 @@ def delete_dimension_group_columns(self, start, end):
 
     def delete_dimension_group_rows(self, start, end):
         """
-        Delete a grouped rows
+        Remove the grouping of a set of rows.
 
         .. note::
-            API behavior with nested groups and non matching ``[start:end[``
+            API behavior with nested groups and non-matching ``[start:end]``
```
lavigne958, line 2162: same here for the range.

lavigne958 (commented):
thank you for this contribution.

there is one detail about the math interval notation currently used that is not right, and the fix is not right too.

appart from that it looks good to me :slightly_smiling_face:

lavigne958 (approved):
Thank you for your contribution 🙂

PR #1079, opened 2022-07-09
add support for inheritFromBefore option when inserting rows or cols

p-doyle:
This is a really small update to add support for the inheritFromBefore option for insertDimensions.  https://developers.google.com/sheets/api/samples/rowcolumn#insert_an_empty_row_or_column

PR #1092, opened 2022-07-20
inherit_from_before option for insert_row/insert_rows

### LINKED ISSUE #1091: `insert_row`/`insert_rows` methods cannot add data to end of sheet (inheritFromBefore option)
**Describe the bug**

The [inheritFromBefore field](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/request#insertdimensionrequest) in the Sheets API is unavailable in `gspread`. This parameter allows formatting / properties to be inherited from the previous row rather than the subsequent row (which is the default).

Because of this, the `insert_row` and `insert_rows` `Worksheet` methods don't work when adding a row at the bottom of a spreadsheet, because there is no subsequent row to inherit formatting from.

**To Reproduce**
1. Make a spreadsheet with N rows
2. Open the sheet as usual and call `worksheet.insert_row([1, 2, 3, 4, 5], index=N+1)`; since `gspread` indices are is 1-indexed, `N+1` corresponds to a new row at the bottom of the sheet
3. Observe the following traceback:

```
  File "/Users/yongrenjie/progs/gspread/gspread/worksheet.py", line 1355, in insert_row
    return self.insert_rows([values], index, value_input_option=value_input_option)
  File "/Users/yongrenjie/progs/gspread/gspread/worksheet.py", line 1393, in insert_rows
    self.spreadsheet.batch_update(body)
  File "/Users/yongrenjie/progs/gspread/gspread/spreadsheet.py", line 131, in batch_update
    r = self.client.request(
  File "/Users/yongrenjie/progs/gspread/gspread/client.py", line 92, in request
    raise APIError(response)
gspread.exceptions.APIError: {'code': 400, 'message': 'Invalid requests[0].insertDimension: range.startIndex must be less than the grid size (N) if inheritFromBefore is false.', 'status': 'INVALID_ARGUMENT'}
```

Partially fixed by #1079 which adds an `inherit_from_before` parameter for `insert_rows`. (I'm working on a PR based on that!)

yongrenjie:
(see also: pr #1079, closes #1091)

Hi! I've gone ahead and done up the code.

As for tests, I'm not entirely sure what to do, since I understand there's no way to access the properties of a sheet, which means here's no way to check whether the properties were correctly inherited from the desired row (correct me if I'm wrong). So I simply added some tests to check the behaviour when called with different indices (i.e. check that an exception is being thrown with invalid combos of `inherit_from_before` and row index).

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1355,17 +1355,23 @@ class Worksheet:
             should be interpreted. Possible values are ``ValueInputOption.raw``
             or ``ValueInputOption.user_entered``.
             See `ValueInputOption`_ in the Sheets API.
-
-        .. _ValueInputOption: https://developers.google.com/sheets/api/reference/rest/v4/ValueInputOption
         :param bool inherit_from_before: (optional) If True, the new row will
             inherit its properties from the previous row. Defaults to False,
             meaning that the new row acquires the properties of the row
-            immediately after it. This cannot be True if the row being added is
-            at the top of the sheet (i.e. index=1).
+            immediately after it.
+            .. warning::
+               `inherit_from_before` must be False when adding a row to the top
+               of a spreadsheet (`index=1`), and must be True when adding to
+               the bottom of the spreadsheet.
+
+        .. _ValueInputOption: https://developers.google.com/sheets/api/reference/rest/v4/ValueInputOption
         """
-        return self.insert_rows([values], index,
-                                value_input_option=value_input_option,
-                                inherit_from_before=inherit_from_before)
+        return self.insert_rows(
+            [values],
+            index,
+            value_input_option=value_input_option,
+            inherit_from_before=inherit_from_before,
+        )
 
     def insert_rows(
         self,
@@ -1388,8 +1394,11 @@ class Worksheet:
         :param bool inherit_from_before: (optional) If true, new rows will
             inherit their properties from the previous row. Defaults to False,
             meaning that new rows acquire the properties of the row immediately
-            after them. This cannot be True if the rows being added are at the
-            top of the sheet (i.e. row=1).
+            after them.
+            .. warning::
+               `inherit_from_before` must be False when adding rows to the top
+               of a spreadsheet (`row=1`), and must be True when adding to
+               the bottom of the spreadsheet.
         """
 
         # can't insert row on sheet with colon ':'
@@ -1402,10 +1411,6 @@ class Worksheet:
         if inherit_from_before and row == 1:
             raise GSpreadException(
                 "inherit_from_before cannot be used when inserting row(s) at the top of a spreadsheet"
-            )
-        if not inherit_from_before and row == self.row_count + 1:
-            raise GSpreadException(
-                "inherit_from_before must be set to True when inserting row(s) at the bottom of a spreadsheet"
             )
 
         body = {
@@ -1455,8 +1460,12 @@ class Worksheet:
         :param bool inherit_from_before: (optional) If True, new columns will
             inherit their properties from the previous column. Defaults to
             False, meaning that new columns acquire the properties of the
-            column immediately after them. This cannot be True if the columns
-            being added are at the left edge of the sheet (i.e. col=1).
+            column immediately after them.
+
+            .. warning::
+               `inherit_from_before` must be False if adding at the left edge
+               of a spreadsheet (`col=1`), and must be True if adding at the
+               right edge of the spreadsheet.
         """
 
         if inherit_from_before and col == 1:
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1359,7 +1359,9 @@ class Worksheet:
             inherit its properties from the previous row. Defaults to False,
             meaning that the new row acquires the properties of the row
             immediately after it.
+
             .. warning::
+
                `inherit_from_before` must be False when adding a row to the top
                of a spreadsheet (`index=1`), and must be True when adding to
                the bottom of the spreadsheet.
@@ -1395,7 +1397,9 @@ class Worksheet:
             inherit their properties from the previous row. Defaults to False,
             meaning that new rows acquire the properties of the row immediately
             after them.
+
             .. warning::
+
                `inherit_from_before` must be False when adding rows to the top
                of a spreadsheet (`row=1`), and must be True when adding to
                the bottom of the spreadsheet.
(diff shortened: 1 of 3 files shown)


yongrenjie:
Uh, I realise I haven't done the columns yet. I'll do that later today / tomorrow, the code should all be similar.

gspread/worksheet.py
```diff
@@ -1351,10 +1357,22 @@ def insert_row(self, values, index=1, value_input_option=ValueInputOption.raw):
             See `ValueInputOption`_ in the Sheets API.
 
         .. _ValueInputOption: https://developers.google.com/sheets/api/reference/rest/v4/ValueInputOption
+        :param bool inherit_from_before: (optional) If true, the new row will
+            inherit its properties from the previous row. Defaults to False,
+            meaning that the new row acquires the properties of the row
+            immediately after it.
         """
```
lavigne958, line 1364: this is a nice description, we should mention that we can't use this option for the row 1 (first row).

gspread/worksheet.py
```diff
@@ -1366,6 +1384,10 @@ def insert_rows(self, values, row=1, value_input_option=ValueInputOption.raw):
             should be interpreted. Possible values are ``ValueInputOption.raw``
             or ``ValueInputOption.user_entered``.
             See `ValueInputOption`_ in the Sheets API.
+        :param bool inherit_from_before: (optional) If true, new rows will
+            inherit their properties from the previous row. Defaults to False,
+            meaning that new rows acquire the properties of the row immediately
+            after them.
```
lavigne958, line 1390: same here.

gspread/worksheet.py
```diff
@@ -1375,6 +1397,15 @@ def insert_rows(self, values, row=1, value_input_option=ValueInputOption.raw):
...
             )
 
+        if inherit_from_before and row == 1:
+            raise GSpreadException(
+                "inherit_from_before cannot be used when inserting row(s) at the top of a spreadsheet"
+            )
+        if not inherit_from_before and row == self.row_count + 1:
+            raise GSpreadException(
+                "inherit_from_before must be set to True when inserting row(s) at the bottom of a spreadsheet"
+            )
```
lavigne958, line 1407: this is a very nice catch but we will be facing an old gspread issue:
the value of `self.row_count` might be outdated, this is because when adding rows we don't update the sheet properties. so adding new row will not increment that value, meaning the first row the user can use the flag (`inherit_from_before=True`, then second call  to add a row at the same index, which is not the last one anymore, will fail or it should not :slightly_frowning_face: )

I am still looking for a solution to that issue

lavigne958 (commented):
thank you for this PR this is nice work :+1: 

I can't merge it now I still need to find a solution to the properties issue I mention in one of the comments :thinking:

yongrenjie:
@lavigne958 Thanks for taking a look! 😄 

Regarding the `row_count` property, I looked through some other issues and I see what you mean, it doesn't look like there's an easy way to handle concurrent modifications in particular.

I guess for now, we just won't check for the bottom-of-spreadsheet case. Do you think it's worth catching the resulting `APIError` (e.g. to re-throw a more understandable error message), or should we just let the user handle it as-is?

```
gspread.exceptions.APIError: {'code': 400, 'message': 'Invalid requests[0].insertDimension: range.startIndex must be less than the grid size (N) if inheritFromBefore is false.', 'status': 'INVALID_ARGUMENT'}
```

lavigne958:
Yep that's my biggest issue with gspread.

For the bottom case just leave it and let the user handle the API error. The error message is quiet explicit but still it's better if we add a warning in the docstring.

Do you know how to add one ?

lavigne958:
Btw you can use the following command to format your code: `tox -e format`

yongrenjie:
> The error message is quiet explicit but still it's better if we add a warning in the docstring. Do you know how to add one ?

Yup! Let me know if that latest commit works.

>  Btw you can use the following command to format your code: `tox -e format`

Thanks!

gspread/worksheet.py
```diff
@@ -1349,12 +1355,31 @@ def insert_row(self, values, index=1, value_input_option=ValueInputOption.raw):
...
             or ``ValueInputOption.user_entered``.
             See `ValueInputOption`_ in the Sheets API.
+        :param bool inherit_from_before: (optional) If True, the new row will
+            inherit its properties from the previous row. Defaults to False,
+            meaning that the new row acquires the properties of the row
+            immediately after it.
+            .. warning::
+               `inherit_from_before` must be False when adding a row to the top
+               of a spreadsheet (`index=1`), and must be True when adding to
+               the bottom of the spreadsheet.
```
lavigne958, line 1365: I believe you need to have a blank line before "warning" and immediately after too.

gspread/worksheet.py
```diff
@@ -1366,6 +1391,14 @@ def insert_rows(self, values, row=1, value_input_option=ValueInputOption.raw):
...
             or ``ValueInputOption.user_entered``.
             See `ValueInputOption`_ in the Sheets API.
+        :param bool inherit_from_before: (optional) If true, new rows will
+            inherit their properties from the previous row. Defaults to False,
+            meaning that new rows acquire the properties of the row immediately
+            after them.
+            .. warning::
+               `inherit_from_before` must be False when adding rows to the top
+               of a spreadsheet (`row=1`), and must be True when adding to
+               the bottom of the spreadsheet.
```
lavigne958, line 1401: Same here.

gspread/worksheet.py
```diff
@@ -1412,7 +1457,22 @@ def insert_cols(self, values, col=1, value_input_option=ValueInputOption.raw):
...
             See `ValueInputOption`_ in the Sheets API.
+        :param bool inherit_from_before: (optional) If True, new columns will
+            inherit their properties from the previous column. Defaults to
+            False, meaning that new columns acquire the properties of the
+            column immediately after them.
+
+            .. warning::
+               `inherit_from_before` must be False if adding at the left edge
+               of a spreadsheet (`col=1`), and must be True if adding at the
+               right edge of the spreadsheet.
```
lavigne958, line 1468: Same here you need a blank line immediately after "warning"

yongrenjie:
That's odd, I remember successfully building the docs locally. Well I guess the only explanation is that I built it and then removed the newlines, which is downright silly.

That should hopefully be all good on the CI now, I've tested everything locally.

PR #1114, opened 2022-09-20
Fix sendNotificationEmail API parameter

aviv:
## Bug description:

The spreadsheet `share` [method](https://docs.gspread.org/en/latest/api/models/spreadsheet.html?highlight=share#gspread.spreadsheet.Spreadsheet.share) sends notification emails even when the `notify` parameter is passed as `False`.

## Solution:

In the API call, `sendNotificationEmail` is the parameter name currently being used. The correct parameter name is `sendNotificationEmails` (plural). See https://developers.google.com/drive/api/v2/reference/permissions/insert#parameters

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -500,7 +500,7 @@ class Client:
         }
 
         params = {
-            "sendNotificationEmail": notify,
+            "sendNotificationEmails": notify,
             "emailMessage": email_message,
             "supportsAllDrives": "true",
         }


lavigne958:
Hi thank you for you contribution but we don't use the DRIVE API V2 anymore. We now use the V3 API. So the current parameter is correct.

See the reference documentation: https://developers.google.com/drive/api/v3/reference/permissions/create#parameters

Closing this PR.

aviv:
Thanks a bunch for taking a look @lavigne958! That's interesting -- I'm wondering why we're seeing the `notify=False` param on the `share` method be disregarded then: we're seeing email notifications go out every time we share a sheet. Would it be helpful for me to create an issue for that?

lavigne958:
Sure it will be, create an issue and from there we'll investigate what is happening. 😁

PR #1115, opened 2022-09-23
(fixed #1113) Supply correct Google API v3 permission for domains

### LINKED ISSUE #1113: Regression on API V3 Migration (#1060)
**Describe the bug**
Since our last `gspread` upgrade (from `5.4.0` to `5.5.0`), we started experiencing file sharing errors when using `perm_type='domain'`.

**To Reproduce**
1. Create a new spreadsheet
2. Share the spreadsheet using `perm_type='domain'`
3. An exception is raised. It reads as follows:

```python
gspread.exceptions.APIError: {'code': 400, 'message': "The domain field is required for permissions of type 'domain'.", 'errors': [{'message': "The domain field is required for permissions of type 'domain'.", 'domain': 'global', 'reason': 'required', 'location': 'permission.domain', 'locationType': 'other'}]}
```

**Expected behavior**
As in `gspread==5.4.0`, we expect the spreadsheet to be successfully shared with no exception raised.

**Code example**
```python
client = gspread.authorize(credentials)
sh = client.create(title)
sh.share('domain.es', perm_type='domain', role='writer')  # <- raises gspread.exceptions.APIError
```

**Environment info:**
- Operating System [e.g. Linux, Windows, macOS]: Debian GNU/Linux 10 (buster)
- Python version: 3.8.13 (Docker image `python:3.8-slim-buster`)
- gspread version: 5.5.0

```python
  File "/app/src/spreadsheet.py", line 38, in get_or_create
    spreadsheet.share("domain.com", perm_type="domain", role="writer")
  File "/usr/local/lib/python3.8/site-packages/gspread/spreadsheet.py", line 512, in share
    self.client.insert_permission(
  File "/usr/local/lib/python3.8/site-packages/gspread/client.py", line 508, in insert_permission
    return self.request("post", url, json=payload, params=params)
  File "/usr/local/lib/python3.8/site-packages/gspread/client.py", line 92, in request
    raise APIError(response)
gspread.exceptions.APIError: {'code': 400, 'message': "The domain field is required for permissions of type 'domain'.", 'errors': [{'message': "The domain field is required for permissions of type 'domain'.", 'domain': 'global', 'reason': 'required', 'location': 'permission.domain', 'locationType': 'other'}]}
```

**Additional context**

As per [API V3 documentation](https://developers.google.com/drive/api/v3/reference/permissions/create#request-body), the `domain` property is expected to be provided when granting permissions with `type=domain`. In the [V2 to V3](https://github.com/burnash/gspread/pull/1060) migration, the `value` property was renamed to `emailAddress`, which only applies to `user` and `group` types.

NickCrews:
closes #1113

I didn't add any tests as I took a look at the current test code and it looked like more than I wanted to get into. Probably they should get added, but if someone feels more qualified then they can start off of this commit.

Without this change, my client code of `sheet.share(DOMAIN, perm_type="domain", role="writer")` fails. With this change, my client code works.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -491,19 +491,25 @@ class Client:
         """
 
         url = "{}/{}/permissions".format(DRIVE_FILES_API_V3_URL, file_id)
-
         payload = {
-            "emailAddress": value,
             "type": perm_type,
             "role": role,
             "withLink": with_link,
         }
-
         params = {
-            "sendNotificationEmail": notify,
-            "emailMessage": email_message,
             "supportsAllDrives": "true",
         }
+
+        if perm_type == "domain":
+            payload["domain"] = value
+        elif perm_type in {"user", "group"}:
+            payload["emailAddress"] = value
+            params["sendNotificationEmail"] = notify
+            params["emailMessage"] = email_message
+        elif perm_type == "anyone":
+            pass
+        else:
+            raise ValueError("Invalid permission type: {}".format(perm_type))
 
         return self.request("post", url, json=payload, params=params)
 


gspread/client.py
```diff
@@ -491,20 +491,26 @@ def insert_permission(
...
+            payload["domain"] = value
+        elif perm_type in {"user", "group"}:
+            payload["emailAddress"] = value
+            params["sendNotificationEmail"] = notify
+            params["emailMessage"] = email_message
+        elif perm_type == "anyone":
+            pass
+        else:
+            raise ValueError("Invalid permission type: {}".format(perm_type))
+
```
NickCrews, line 513: IF this gets much more complicated then it should probably be refactored to somehow be delegated to a _get_user_group_permissions(), _get_domain_permissions(), _get_anyone_permissions() or similar

lavigne958:
Hi thank you for this PR I went through the API doc and yes there is an issue here. I need some time to check the whole API to create new permission and review your code.  It must be " general" and work for any permission creation.

Once I understand all possible options and their constraints I'll review your code and we'll try to merge something that works for everyone 😉

gspread/client.py
```diff
@@ -491,20 +491,26 @@ def insert_permission(
...
-            "emailMessage": email_message,
             "supportsAllDrives": "true",
         }
 
+        if perm_type == "domain":
+            payload["domain"] = value
+        elif perm_type in {"user", "group"}:
+            payload["emailAddress"] = value
+            params["sendNotificationEmail"] = notify
+            params["emailMessage"] = email_message
```
lavigne958, line 508: after reading the documentation of the drive API V3, the `emailMessage` is not restricted to only `user` or `groupe` type.
We can keep it in the general parameters. I don't know who will receive the email in case of `domain` :thinking: so keeping it like this make more sense to me.

gspread/client.py
lavigne958: I agree but so far from what I see in the drive API documentation that's all.

lavigne958 (approved):
Thank you for taking some time to solve this issue. It's all good to me.
There the detail about the email message param but it can be merge like this for now and we can adjust it later.

NickCrews:
@lavigne958 Is there anything you need me to do here? Or are you just waiting on @burnash (or some other owner) to merge?

gspread/client.py
NickCrews: I had the same thought. If someone wants this, would be trivial to adjust later.

lavigne958:
> @lavigne958 Is there anything you need me to do here? Or are you just waiting on @burnash (or some other owner) to merge?

Yes I was waiting for you opinion on the "email notification" thing. That's it.

Lets go for it !

PR #1144, opened 2022-11-14
Update dependencies

cbini:
The pinned older version of `google-auth` is starting to cause some conflicts in my projects that rely on `gspread`, and I'm unable to update to its latest version. I used [PDM](https://pdm.fming.dev/latest/) to resolve the minimal-supported versions while keeping 3.6 support. I also refactored the dev requirements to `setup.py` and updated `tox.ini` to reference that instead of the additional requirements files.

The only potentially-breaking change is that 3.6.2 is now the minimally-required version of Python. `black` requires at least that to work, even though they don't officially support 3.6 anymore.

I ran all `tox` tests locally, and everything passed.


cbini:
Thanks for considering this anyway! FWIW you should be able to update to `google-auth-oauthlib>=0.7.1` and `google-auth>=2.14.1` while keeping 3.6.0+ support. It's just the dev dependencies, mainly `black` I think, that are preventing it.

lavigne958:
I suppose yes I can, but I can only change the dev dependencies.
The dependencies set in `setup.py` are the public dependencies, I cannot change them until I change the major version of gspread. This is because if anyone out there still uses an old version of google auth lib, I cannot force this person from upgrading. 

I plan to do so in the next major release (version 6.0.0).

In the mean time I can review your PR, when we plan to release version 6.0.0 we can start merging your PR.

lavigne958:
Hi I can merge your changes now that I solves some issues with the dependencies.

would you be keen on rebasing your PR so you use the current version of google auth libraries as dependencies and keep a dedicated file for doc dependencies please ?

lavigne958:
I cannot introduce these changes. Now that we use dependabot we use a dedicated requirements.txt file and for readthedoc we need a dedicated file too.

closing this PR

Issue #1002, opened 2022-02-24
get_all_worksheet_values feature request


JPGarCar:
**Is your feature request related to a problem? Please describe.**
API limits is the problem. When wanting to read values from all or most of the worksheets in a google sheets I have to get the values of each worksheet one at a time, that is one API call per worksheet. This can make the program reach the API limit very quickly if enough users access the program.

**Describe the solution you'd like**
a `get_all_worksheet_values()` function that uses the batchGet endpoint to retrieve multiple worksheet data in one API call.
If a list of worksheet names are given, then use those, else grab all the worksheet names and retrieve all their values. I know the latter would involve making two API calls, but it would be worth on most cases.

I know the library has a `values_batch_get()` method. What I am requesting is an abstraction to this method.

I would prefer the function to return a dict of all the worksheet values, mapped by the worksheet name/title. 

**Describe alternatives you've considered**
Only alternatives are calling for each worksheet data at a time, or just using `values_batch_get()`.

**Additional context**
I am happy to tackle this feature, would just like to get some input from the devs on if the approach is correct and get any feedback or improvements.

JPGarCar:
I was able to make a demo, it works!

```
        def get_all_data(self, skip_worksheet_titles: list[str] = None):
        """ Grabs all the data from all the worksheets in one API call. Skips any worksheets that were named in the
        skip_worksheet_title parm.
        :returns Dict of worksheet data with worksheet title as key
        """

        if skip_worksheet_titles is None:
            skip_worksheet_titles = []

        def colnum_to_string(n: int) -> str:
            """  Converts an int to a sheet column string index.
            Example: 0 -> A, 27 -> AA
            """
            string = ""
            while n > 0:
                n, remainder = divmod(n - 1, 26)
                string = chr(65 + remainder) + string
            return string

        def extract_title_from_range(range_string: str) -> str:
            """ Will extract the sheet title from a range.
            Example: "'Volunteer Portal'!A1:Z1005" -> "Volunteer Portal"
            """
            first_appearance = range_string.index("'")
            return range_string[first_appearance + 1:range_string.index("'", first_appearance + 1)]

        ranges = []

        for worksheet in self.google_sheet.worksheets():
            if worksheet.title not in skip_worksheet_titles:
                ranges.append('{}!A1:{}'.format(worksheet.title, colnum_to_string(worksheet.col_count)))

        values = self.google_sheet.values_batch_get(
            ranges=ranges
        )

        return_data = {}

[...]

PR #1180, opened 2023-03-18
get_all_worksheet_values feature

swimninja247:
Addresses #1002.

The function returns a dict of all the worksheet values, mapped by the worksheet name/title.
I have not written tests for this function yet.

### DIFF
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -20,8 +20,14 @@ from .urls import (
     SPREADSHEET_VALUES_CLEAR_URL,
     SPREADSHEET_VALUES_URL,
 )
-from .utils import ExportFormat, finditem, quote
-from .worksheet import Worksheet
+from .utils import (
+    ExportFormat,
+    finditem, quote,
+    column_index_to_letter,
+    extract_title_from_range,
+    fill_gaps,
+)
+from .worksheet import Worksheet, ValueRange
 
 
 class Spreadsheet:
@@ -739,3 +745,30 @@ class Spreadsheet:
             raise WorksheetNotFound("worksheet id {} not found".format(sheetid))
 
         return sheet.get("protectedRanges", [])
+
+    def get_all_worksheet_values(self, skip_worksheet_titles: list[str] = None):
+        """ Grabs all the data from all the worksheets in one API call. Skips any worksheets that were named in the
+        skip_worksheet_title parm.
+        :returns Dict of worksheet data with worksheet title as key
+        """
+
+        if skip_worksheet_titles is None:
+            skip_worksheet_titles = []
+
+        ranges = []
+
+        for worksheet in self.google_sheet.worksheets():
+            if worksheet.title not in skip_worksheet_titles:
+                ranges.append(f'{worksheet.title}!A1:{column_index_to_letter(worksheet.col_count)}')
+
+        values = self.google_sheet.values_batch_get(
+            ranges=ranges
+        )
+
+        return_data = {}
+
+        for values in values['valueRanges']:
+            value_range = ValueRange.from_json(values)
+            return_data[extract_title_from_range(value_range.range)] = fill_gaps(value_range)
+
+        return return_data
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -22,12 +22,13 @@ from .urls import (
 )
 from .utils import (
     ExportFormat,
-    finditem, quote,
     column_index_to_letter,
     extract_title_from_range,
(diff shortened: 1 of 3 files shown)


lavigne958:
Hi thank you for your contribution, I'll have a proper look at the code when I have some free time (very soon).

I looked at it quickly and I see some changes to be done. I'll come back with a complete review.

gspread/spreadsheet.py
```diff
@@ -739,3 +746,32 @@ def list_protected_ranges(self, sheetid):
...
+        if skip_worksheet_titles is None:
+            skip_worksheet_titles = []
+
+        ranges = []
+
+        for worksheet in self.google_sheet.worksheets():
+            if worksheet.title not in skip_worksheet_titles:
+                ranges.append(
+                    f"{worksheet.title}!A1:{column_index_to_letter(worksheet.col_count)}"
+                )
```
lavigne958, line 765: We have utils function to compute a full range, here you don't need to give it a range. you can give an empty range the API will return the whole sheet with all data.

gspread/spreadsheet.py
```diff
@@ -739,3 +746,32 @@ def list_protected_ranges(self, sheetid):
...
+
+        values = self.google_sheet.values_batch_get(ranges=ranges)
```
lavigne958, line 767: again `google_sheet` object does not exists

gspread/utils.py
```diff
@@ -482,6 +484,31 @@ def column_letter_to_index(column):
...
+    >>> a1_to_rowcol(1)
+    'A'
+    """
+
+    if not isinstance(column, int) or column < 1:
+        raise InvalidInputValue(f"invalid value: {column}, must be an int >= 1")
+
+    return rowcol_to_a1(1, column).strip("1")
+
+
```
lavigne958, line 511: we don't need a specific function to do what `rowcol_to_a1` already does if you give it 1 as second arg.

swimninja247:
Thanks for the review.  I resolved your comments and fixed the docstrings for the code I added.  Cheers

gspread/spreadsheet.py
```diff
@@ -739,3 +739,33 @@ def list_protected_ranges(self, sheetid):
...
+        skip_worksheet_title param.
+
+        :param list skip_worksheet_titles: A list of worksheet titles to skip.
+        :returns: a dict of worksheet data with worksheet title as key
+        :rtype: dict
+        """
+
+        if skip_worksheet_titles is None:
+            skip_worksheet_titles = []
+
```
lavigne958, line 754: if you set it to `[]` empty list when it's not set, then just set the default value to `[]` in the argument in the method definition.

gspread/spreadsheet.py
```diff
@@ -739,3 +739,33 @@ def list_protected_ranges(self, sheetid):
...
+        ranges = []
+
+        for worksheet in self.worksheets().worksheets():
```
lavigne958, line 757: that won't work, it should be only `self.worksheets()`

gspread/spreadsheet.py
```diff
@@ -739,3 +739,33 @@ def list_protected_ranges(self, sheetid):
...
+            if worksheet.title not in skip_worksheet_titles:
+                ranges.append(worksheet.title)
```
lavigne958, line 759: Here we should protect the worksheet title using the util function `absolute_range_name`

gspread/utils.py
```diff
@@ -27,6 +27,8 @@
 URL_KEY_V1_RE = re.compile(r"key=([^&#]+)")
 URL_KEY_V2_RE = re.compile(r"/spreadsheets/d/([a-zA-Z0-9-_]+)")
 
+TITLE_RANGE_RE = re.compile(r"'(.*?)'!.*")
```
lavigne958, line 30: This won't always work, here we can have 2 cases:
1. names with a ` ` blank space or that starts with `'` so they start with a `'` but they may have multiple `'` surrounding the actual name
2. no ` ` blank space, no `'` at starts or end. so it's a single string with only characters.

The regex should be improved to match all possible titles.

lavigne958 (commented):
took me some time but I finally found the time to review your PR.

They are some small adjustments that need to be addressed.

gspread/spreadsheet.py
butvinm: I suppose the initial solution was right, because usage of a mutable object as default argument can follow wrong behaviour. You can see example of such problem [there](https://realpython.com/python-mutable-vs-immutable-types/#using-mutable-default-values)

Issue #966, opened 2021-12-08
Add type hints


lavigne958:
Add type hints to gspread.
Add mypy to check types in the CI.

This allows developer to get type hints from their editors when developing.
This allows users to get the type hints from their editors when using gspread and have a clear hint on the return value of each function, the type of each expected argument.
It should make the use of gspread much simpler and the development of gspread easier.

lavigne958:
remove that issue from milestone 5.2.0.

It is much longer than expected.
Tried the _bottom up_ approach and it's not right, tried having a strict typing and it does not work well with JSON (see issue python/typing#182)

Will do it again but using a _top down_ approach by typing
1. every response from the library as a `typedDict` see [PEP 589](https://www.python.org/dev/peps/pep-0589/)
2. from there type every method/function manipulating the response from Spreadsheet API
3. Will not use strict typing and let `mypy` bring up only what needs to be typed.

Postone that issue for later, will work on it a background task.

Any help is welcome :slightly_smiling_face:

Congee:
What about gradually typing this library? We can write `Json = typing.Any` to skip typing json for now and type other stuff as much as possible. It is better than nothing for users.

lavigne958:
I tried that in my previous attempt and it does not help. We access keys of z dict and the values may be an `int` or a `str` and it does not help us identifying the exact type for the next functions.

I will see may be i can simply add types to some user facing functions and leave the rest for later, that is an other option too 🤔

lavigne958:
I found the bottleneck of this feature:

the `spreadsheet` class can return instance of the `worksheet` class.
The `worksheet` class has a link to the `spreadsheet` parent object to call every low level methods (`batch_get`, `batch_update`, ...)

This brings circular dependency.

In order to simply type the `gspread` library we'll need to: 
- move any low level method to the `worksheet` class
- break the circular dependency between `spreadsheet` and `worksheet`.

From there we should be able to add some basic typing.

Will plan this for release 6.0.0

OskarBrzeski:
How much progress has been made for this issue? Is it possible to see this progess at the current moment?

lavigne958:
Hi it's starting, I have a branch where I add commits for the 6.0.0 release.

You can follow it here: https://github.com/burnash/gspread/commits/feature/release_6_0_0

OskarBrzeski:
I could try helping towards this. If there are any particular files you'd like me to type hint, I'll gladly give it a go.

lavigne958:
[...]

PR #1196, opened 2023-05-28
#966  Add sketch typing for utils.py

butvinm:
I add sketch types for `utils.py` but have following issues:

### Optional  imports

OAuth2Credentials, AccessTokenCredentials, GoogleCredentials, ServiceAccountCredentials are optional and only needed for oauth2client, so cannot be resolved as a type.

Currently, I have to use the following code to make it work:

```python
def _convert_oauth(credentials: Any) -> Credentials:
    ...

def _convert_service_account(credentials: Any) -> Credentials:
    ...
```

But it's not ideal, because it's not type safe. I think, we can add try-except imports to functions body:

```python
def _convert_oauth(credentials: Any) -> Credentials:
    try:
        from oauth2client.client import OAuth2Credentials, AccessTokenCredentials, GoogleCredentials
    except ImportError:
        pass
    ...
```

### NamedTuples used as enums

NamedTuples are used for options, but it's not proper way to use them. It also produce type checking issues.

I think, we can use Enum instead of NamedTuple, if there is no reasons for backward compatibility.

### Some unclear types

I'm not sure if I have set the correct types for some functions because they are not explained anywhere. Therefore, I need some validation of my propositions.

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -158,7 +158,7 @@ def finditem(func: Callable[[T], bool], seq: Iterable[T]) -> T:
 def numericise(
     value: Optional[AnyStr],
     empty2zero: bool = False,
-    default_blank: Optional[str] = "",
+    default_blank: Optional[AnyStr] = "",
     allow_underscores_in_numeric_literals: bool = False,
 ) -> Optional[Union[int, float, AnyStr]]:
     """Returns a value that depends on the input:
@@ -233,7 +233,7 @@ def numericise(
 def numericise_all(
     values: List[Optional[AnyStr]],
     empty2zero: bool = False,
-    default_blank: Union[AnyStr, int, float] = "",
+    default_blank: Optional[AnyStr] = "",
     allow_underscores_in_numeric_literals: bool = False,
     ignore: List[int] = [],
 ) -> List[Optional[Union[int, float, AnyStr]]]:
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -581,7 +581,7 @@ def fill_gaps(
         return [[]]
 
 
-def cell_list_to_rect(cell_list: List["Cell"]) -> List[List[Optional[str]]]:
+def cell_list_to_rect(cell_list: List[Any]) -> List[List[Optional[str]]]:
     if not cell_list:
         return []
 
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -11,15 +11,17 @@ from collections import defaultdict, namedtuple
 from collections.abc import Sequence
 from functools import wraps
 from itertools import chain
-from math import inf
-from typing import AnyStr, Optional, Tuple, Union
+from typing import (Any, AnyStr, Callable, Dict, Iterable, List, NewType,
+                    Optional, Tuple, TypedDict, TypeVar, Union)
 from urllib.parse import quote as uquote
 
 from google.auth.credentials import Credentials as Credentials
 from google.oauth2.credentials import Credentials as UserCredentials
-from google.oauth2.service_account import Credentials as ServiceAccountCredentials
+from google.oauth2.service_account import \
+    Credentials as ServiceAccountCredentials
 
-from .exceptions import IncorrectCellLabel, InvalidInputValue, NoValidUrlKeyFound
+from .exceptions import (IncorrectCellLabel, InvalidInputValue,
+                         NoValidUrlKeyFound)
 
 MAGIC_NUMBER = 64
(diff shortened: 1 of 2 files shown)


lavigne958:
Hi I'll have a look at the PR.

speaking of the credentials:
- I know it's an issue
- I'll have a look at the code later, we might not need it anymore, we are on the next major release if it's not necessary we should be able to remove that entire function, but first I need to check

speaking of the Namedtuple:
- yes I agree, we can change that, but that will impact other files (like `worksheet.py`)
- I propose we first type what we can
- In a second PR we makes changes to use Enums instead of NamedTuple

speaking of functions:
- It's a nightmare to type, I managed in the past to type some of the wrappers but using mostly some `Any` parameters are they are generics and can be anything.
- I'll have a look and make some proposals if I find anything better but I don't mind so much if it's ot perfect

Thank you for this contribution, the major goal here is to move forward and to type what we can so we can have a new 6.0.0 release the sooner the better :upside_down_face:

gspread/utils.py
```diff
@@ -519,19 +549,23 @@ def extract_id_from_url(url):
...
 
-def rightpad(row, max_len):
+def rightpad(row: List[Any], max_len: int) -> List[Any]:
     pad_len = max_len - len(row)
     return row + ([""] * pad_len) if pad_len != 0 else row
 
 
-def fill_gaps(L, rows=None, cols=None):
+def fill_gaps(
+    L: List[Any],
```
lavigne958, line 565: This exactly is expected to be a matrix so it should be: `List[List[Any]]`

gspread/utils.py
```diff
@@ -519,19 +549,23 @@ def extract_id_from_url(url):
...
+    rows: Optional[int] = None,
+    cols: Optional[int] = None
+) -> List[Any]:
```
lavigne958, line 568: looking at the code I see that it should return too `List[List[Any]]` and we should update the return value in case of value error to: `[[]]` please.

lavigne958 (commented):
Nice work thank you for this.

I have only a few comments regarding some details.

- About the callable, that's fine to me, I might rework it anyway
- could you please format your code, by running: `tox -e format` in order to make it pass the CI, thanks.

butvinm:
I updated the PR

gspread/utils.py
```diff
@@ -581,7 +581,7 @@ def fill_gaps(
         return [[]]
 
 
-def cell_list_to_rect(cell_list: List["Cell"]) -> List[List[Optional[str]]]:
+def cell_list_to_rect(cell_list: List[Any]) -> List[List[Optional[str]]]:
```
butvinm, line 572: Usage of literal "Cell" type is recognized as error by linter. But importing of Cell class  produce circular imports. I think we can refactor utils.py in later and place utilities that depends on specific classes in same modules. Now I replace it Cell with Any for CI passing

lavigne958: Hi yes I noticed similar issues when doing Worksheet.py, to solve this you can use the following trick:

- at the top of the file check for the value of `TYPE_CHECKING` 
- if true, then import what you need regardless of circular dependencies
- keep your code as it was by putting double quotes around the class name

The check at the top of the file is only valid when mypy is running, that ensures the code runs and allows you to check the types.

You can check on my branch `feature/add_type_hints_worksheets` i do that for the http client

gspread/utils.py
```diff
@@ -196,29 +216,27 @@ def numericise(
...
 def numericise_all(
-    values,
-    empty2zero=False,
-    default_blank="",
-    allow_underscores_in_numeric_literals=False,
-    ignore=[],
-):
+    values: List[Optional[AnyStr]],
+    empty2zero: bool = False,
+    default_blank: Union[AnyStr, int, float] = "",
```
lavigne958, line 236: the type here should be `Optional[AnyStr]`

lavigne958 (commented):
sorry I was wrong before, you were right about the type for the `default_blank` it should be as mentioned in the comment.  Users are allowed to pass `None` as value for default blank.

apart from that I see issues with the named tuples but that will go away anyway so no worries bout them.

I see a couple of issues with the `float("inf")` which has only be a pain to handle so let's merge it like this (after you revert the default blank type) then I'll have a look if I keep the `Inf` value of not.

thanks again for your work.

butvinm:
I also changed type of `default_blank` at `numericise` method, because it has same behaviour as `numericise_all`

lavigne958 (approved):
looks good to me.

thank you for your contribution.

Issue #1090, opened 2022-07-16
rework `sort` method signature


lavigne958:
## Overview
rework the signature of the method `Worksheet.sort` to expose the expected parameters.

## Details

The method `Worksheet.sort` uses some signature style that hides the possible argument for the method.
This is not needed and not useful for the user.

the current signature is `sort(*specs, **kwargs)` and expects the following arguments:
1. the list of tuples with `(the column, the sort order)`
2. the range where to apply this sort, if provided

the signature should be: `sort(*sorting_spec: List[Tuple[int, str]], range: Optional[str])`

**this update should not introduce any breaking change**.  the first argument must remain the list of positional arguments in order to maintain backward compatibility. The other key/value arguments can be updated to only accept `range`. This is the best we can do

OskarBrzeski:
Arbitrary argument lists use the type of the individual arguments, so the typing there should be:
`*sorting_spec: Tuple[int, str]`.

With Python 3.8, we could go further and specify that the string in each tuple conforms to one of two options like so:
`*sorting_spec: Tuple[int, Literal["asc", "des"]]`.

lavigne958:
Thanks for the tip, I'll use that ! Bad name the kwarg `range` so it's explicit

PR #1198, opened 2023-05-30
Add_deprecated_warning_sort_method

lavigne958:
The function signature will change in the coming release 6.0.0

Add a warning for all users.

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -6,10 +6,13 @@ This module contains common worksheets' models.
 
 """
 
+import warnings
+
 from .cell import Cell
 from .exceptions import GSpreadException
 from .urls import SPREADSHEET_URL, WORKSHEET_DRIVE_URL
 from .utils import (
+    DEPRECATION_WARNING_TEMPLATE,
     Dimension,
     PasteOrientation,
     PasteType,
@@ -27,6 +30,7 @@ from .utils import (
     is_scalar,
     numericise_all,
     rowcol_to_a1,
+    
 )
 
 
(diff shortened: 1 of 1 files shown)


alifeee:
This looks good.

The signature will be changed, as per #1090.

A similar warning should be added for #1209?

Merge-time 👍

alifeee (changes requested):
Look good.

`sort(range: str, specs: List[Tuple[int, Union['asc', 'desc']]])` is quite hard to read if you don't use types in python that often.

Might be nice to have a simple "...arguments will swap", or a simple description of the change

lavigne958:
> Look good.
> 
> `sort(range: str, specs: List[Tuple[int, Union['asc', 'desc']]])` is quite hard to read if you don't use types in python that often.
> 
> Might be nice to have a simple "...arguments will swap", or a simple description of the change

that's true, I've been doing python static typing for too long :-p 

I can propose the following method signature (without the types just the arguments names) it's explicit and short: 
`sort(range, specs)` in addition with your comment it should look like:

`This function signature will change, arguments will swap places:  sort(range, specs)`

what do you think ?

lavigne958:
> This looks good.
> 
> The signature will be changed, as per #1090.
> 
> A similar warning should be added for #1209?
> 
> Merge-time +1

yes a similar one will be added very soon, but I put it in a different PR as planned it after this one.

alifeee:
> I can propose the following method signature (without the types just the arguments names) it's explicit and short: `sort(range, specs)` in addition with your comment it should look like:
> 
> `This function signature will change, arguments will swap places: sort(range, specs)`
> 
> what do you think ?

This is good to me 👍

lavigne958:
I updated the message, waiting for you to change your "request for change" to "approve" before merging in case you see anything else.

alifeee (approved):

PR #1200, opened 2023-05-31
CLEANUP: cleanup tox.ini, and ignore ./env

alifeee:
tox was taking a long time to run.  This was because it was checking all the files in my python virtual environment, in `./env`.

The [.gitignore](https://github.com/burnash/gspread/blob/master/.gitignore#L9) suggests that a python virtual environment should be called `env`, which is what mine is.

I have added a skip/ignore to the lint/format commands, so they no longer look in `./env`. 

This also has the benefit that they are much faster to run (around 8 secs to lint, 5 secs to format).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,6 +22,7 @@ statistics = True
 
 [isort]
 skip=.tox,./env
+profile=black
 
 # Used by the CI to check code format/security
 [testenv:lint]
@@ -30,14 +31,14 @@ deps = -r lint-requirements.txt
 commands = black --check --diff .
         codespell --skip=".tox,.git,./docs/build,./env" .
         flake8 .
-        isort --check-only --profile black .
+        isort --check-only .
 
 # Used by developers to format code, best advised to be run before commit
 [testenv:format]
 description = Format code
 deps = -r lint-requirements.txt
 commands = black .
-        isort --profile black .
+        isort .
 
 
 [testenv:build]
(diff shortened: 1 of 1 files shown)


lavigne958:
Thanks that's great ! 😃

I noticed it too.

I thought about the following changes: using the command `git ls-files *.py` in order to only format/lint tracked files.

I mostly faced issues when it tries to lint my dev files that are not formatted at all.

We can mix your proposal with mine in this PR.

lavigne958:
I let you do the changes in your PR?

Remove the exclude list in lint/format. Then instead of `.` for the folder to scan, replace it with `$(git ls-files *.py)`

Hope this works but I'm not sure if tow runs the commands in a shell or not.

lavigne958:
In fact I was wrong we don't have access to a shell we should keep it like this for now (in terms of exclude/include).

We may introduce exclude list to `black`, like the others: .tox, env, docs, ...

alifeee:
> In fact I was wrong we don't have access to a shell we should keep it like this for now (in terms of exclude/include).

Ah, ok. Sounds good.

> I let you do the changes in your PR?

In theory, you should be able to make changes in my repository, since you are a maintainer.

image

However, I will make this small change.

alifeee:
I believe this can be merged 👍

lavigne958 (approved):

PR #1206, opened 2023-06-05
Refactor/update-contributing-guide

alifeee:
Update contributing guide

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -32,7 +32,7 @@ This will use the currently recorded HTTP requests + responses. It does not make
 **Note:** the CI runs that command, if it fails you won't be able to merge
 your changes in gspread.
 
-```
+```bash
 tox -e py
 ```
 
@@ -40,7 +40,7 @@ tox -e py
 
 Example:
 
-```
+```python
 tox -e py -- -k test_find -v -s
 ```
 
@@ -51,7 +51,7 @@ You must in that case provide a service account credentials in order to make the
 You can control vcrpy's [Record Mode](https://vcrpy.readthedocs.io/en/latest/usage.html#record-modes) using `GS_RECORD_MODE` environment variable.
 
 The following command will run the entire test suite and record every HTTP request.
-```
+```python
 GS_RECORD_MODE=all GS_CREDS_FILENAME=<YOUR_CREDS.json> tox -e py
 ```
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -27,10 +27,10 @@
 2. Run tests offline:
 
 Run the test suite using your current python version, in offline mode.
-This will use the curently recorded HTTP requests + responses. It does not make any HTTP call, does not require an active internet connection.
+This will use the currently recorded HTTP requests + responses. It does not make any HTTP call, does not require an active internet connection.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -176,16 +176,16 @@ worksheet.batch_update([{
 
 ## [Documentation](https://gspread.readthedocs.io/en/latest/)
```
lavigne958, line 177: I think as we are currently updating that part we can change something too here:

- I always get confused when I see a link on the header of a markdown file.
- We could leave the header with `Documentation`
- and move the actual link to a basic paragraph bellow like "Documentation can be found here" with the link on "here"

What do you think ?

.github/CONTRIBUTING.md
```diff
@@ -1,121 +1,120 @@
...
-```
+### `GS_CREDS_FILENAME` environment variable
 
-Then run the tests in offline mode to make sure you have recorded everything.
+You must provide service account credentials using the `GS_CREDS_FILENAME` environment variable in order to make HTTP requests to the Sheets API.
 
-```
-tox -e py
-```
+[Obtain OAuth2 credentials from Google Developers Console](http://gspread.readthedocs.org/en/latest/oauth2.html).
```
lavigne958, line 81: In the record mode we expect a service account in order to run the test onlie.

The correct link is: `https://docs.gspread.org/en/latest/oauth2.html#for-bots-using-service-account`

.github/CONTRIBUTING.md
```diff
@@ -1,121 +1,120 @@
...
 
-## Render Documentation
+### Run test, capturing *only new* HTTP requests
 
-The documentation uses [reStructuredText](http://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html#rst-index) markup and is rendered by [Sphinx](http://www.sphinx-doc.org/).
+To record new HTTP requests:
 
-To build the documentation locally, use the following command:
+1. Remove the file holding the recorded HTTP requests of the test(s).
+  (e.g.: for the file `tests/cell_test.py` delete `tests/cassettes/CellTest.json`)
```
lavigne958, line 110: we create a single cassette file for a single test, so in the `cell_test.py` file we have multiple tests, so it produces multiple cassettes.

Here in order to guide the new user perfectly we should be more prices and mention:
- for the file `tests/cell_test.py`
- for the test `test_a1_value`
- remove the file: `tests/cassettes/CellTest.test_a1_value.json`

.github/CONTRIBUTING.md
```diff
@@ -1,121 +1,120 @@
...
 
-**Note:** the CI runs that command, if it fail you won't be able to merge
-your changes in GSpread.
+Installing all the `*.txt` files is optional (tox installs dependencies per-run).
 
-```
-tox -e py
+```bash
+pip install tox
+pip install -r ./test-requirements.txt -r ./lint-requirements.txt -r ./docs/requirements.txt
```
lavigne958, line 32: Interestingly enough we don't require to install the dependency files. Tox will install them in a virtual env :smiley: 

The user is only *required* to installed `tox` then the rest is just useful for your editor to find the libraries installed on your computer for development purpose (mostly google dependencies)

lavigne958 (commented):
Thank you for this awesome contribution ! 

This is much easier to understand, I only have a few comments then we after fixing them we can merge it !

alifeee:
ready to merge! click tha button! 🚀

PR #1207, opened 2023-06-06
457-fetch-without-hidden-worksheets

### LINKED ISSUE #457: Ability to distinguish between normal worksheets and hidden worksheets?
There doesn't seem to be any way to fetch only the non-hidden worksheets in a particular workbook. (It's unclear to me from reading Google's documentation whether that is even possible using their API.)

alifeee:
add `exclude_hidden` kwarg to gspread.Spreadsheet.worksheets(), so you can fetch only non-hidden sheets

closes #457

### DIFF
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -295,12 +295,22 @@ class Spreadsheet:
         except (StopIteration, KeyError):
             raise WorksheetNotFound("id {} not found".format(id))
 
-    def worksheets(self):
+    def worksheets(self, exclude_hidden: bool = False):
         """Returns a list of all :class:`worksheets <gspread.worksheet.Worksheet>`
         in a spreadsheet.
+
+        :param exclude_hidden: (optional) If set to ``True`` will only return
+                                 visible worksheets. Default is ``False``.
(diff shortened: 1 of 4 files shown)


alifeee:
Marked as draft. Might change `exclude_hidden` to `include_hidden` (with opposite behaviour), as I think it is more clear(?)

lavigne958:
Thank you for this awesome feature ! 

Regarding the extra parameter: that is a good question. In both cases when users will update their code to skip hidden sheets they will need to change the default value anyway, so no option is better than the other to me. 

- in `exlude_hidden`, as a user I now don't want them so I will update my code to use the new parameter set to `True`
- in `include_hidden`, as a user I now don't want them so I will update my code to use the new parameter set to `False`
Both cases the interest of the new feature will require update from users, so I would go for the first one you propose `exclude_hidden` just because usually in the computer :computer:  world :earth_africa: we tend to exclude thing :thinking:

alifeee:
Hm, yes. I will leave it as is. Plus, then I don't have to make more commits...

lavigne958 (approved):
Agreed, let's go for merge.

PR #1205, opened 2023-06-05
add legacy option for DateTimeOption

alifeee:
now, someone can use the correct (new) spelling, or the incorrect (old) spelling
this is thus not longer a breaking change

See #1187 for further discussion

PR #1210, opened 2023-06-06
Spelling fix (with legacy option)

alifeee:
Spelling fix: #1187
Legacy fix: #1205

This is a combination of the two, because pull requests and git are confusing.


lavigne958 (approved):
Great thanks. I understand now, that's much better. Thanks

PR #1215, opened 2023-06-08
700-fill-merged-cells

### LINKED ISSUE #700: Same value for all merged cells

alifeee:
Implement feature suggestion from #700:

" Same value for all merged cells "

This is added as an optional (default off) feature of `Worksheet.get_values`

I would have edited #741 but it was very old so hard to edit without a git mess.

closes #700
closes #741

this is probably not done in the best way. I use an extra API request where I should perhaps just use the internal Worksheet._properties object. I can change this if requested.

note: this could also be implemented for `Worksheet.row_values` and `Worksheet_get`, but I did not do so, because I did not want to play with `ValueRange`...

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -733,7 +733,7 @@ def combined_merge_values(worksheet_metadata, values):
 
     for merge in merges:
         start_row, end_row = merge["startRowIndex"], merge["endRowIndex"]
-        start_col, end_col = (merge["startColumnIndex"], merge["endColumnIndex"])
+        start_col, end_col = merge["startColumnIndex"], merge["endColumnIndex"]
         top_left_value = values[start_row][start_col]
         row_indices = range(start_row, end_row)
         col_indices = range(start_col, end_col)
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -711,31 +711,21 @@ def accepted_kwargs(**default_kwargs):
 
 def combined_merge_values(worksheet_metadata, values):
     """For each merged region, replace all values with the value of the top-left cell of the region.
-    `worksheet_metadata` is the metadata returned by the Google API for the worksheet.
-      i.e.,
-        {
-            "properties": { ... }
-            "merges": [
-                {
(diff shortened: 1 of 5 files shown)


gspread/worksheet.py
```diff
@@ -434,7 +446,15 @@ def get_values(self, range_name=None, **kwargs):
             worksheet.get_values('A2:B4', value_render_option=ValueRenderOption.formula)
         """
         try:
-            return fill_gaps(self.get(range_name, **kwargs))
+            vals = fill_gaps(self.get(range_name, **kwargs))
+            if combine_merged_cells:
```
lavigne958, line 450: please compare to boolean value: `if combine_merged_cells is True`, otherwise someone can just pass any random object instead and it will pass the condition.

gspread/utils.py
```diff
@@ -709,6 +709,41 @@ def wrapper(*args, **kwargs):
...
+
+    :param values: The values returned by the Google API for the worksheet. 2D array.
+    """
+    merges = worksheet_metadata.get("merges", [])
+    # each merge has "startRowIndex", "endRowIndex", "startColumnIndex", "endColumnIndex
+    new_values = [[v for v in row] for row in values]
+
+    for merge in merges:
+        start_row, end_row = merge["startRowIndex"], merge["endRowIndex"]
+        start_col, end_col = (merge["startColumnIndex"], merge["endColumnIndex"])
```
lavigne958, line 736: why use a tuple here ? just a typo may be ?

lavigne958 (commented):
thanks for the feature it has been waiting for so long. 

I only have a couple of comments

lavigne958 (approved):
Good job :partying_face:

PR #1217, opened 2023-06-08
Refresh-test-cassettes

alifeee:
re-generate test cassettes

note: I could generate them all in one go. I guess my API key hasn't been that rate-limited yet heh...

## test_get_all_values_date_time_render_options

One test broke: the one added by #1187

Looks like Google might have changed the API, so it doesn't error any more? At the time, it did seem weird that it would return an error. Dunno, can't find much reference of it online...

Anyway, changed the test, and documentation to reflect this.

(documentation is back to the state it was before #1187)


lavigne958 (approved):
I am pretty sure they considered this as a bug and fixed it. Makes more sense that the option for dates I'd ignored in case we ask for a formatted output.

You managed to renew everything in one go ?? 😱😎

I never could. Sounds like I need to renew my service account 😆

Thanks again that's a good catch.

Issue #881, opened 2021-05-24
Update local properties when updating spreadsheet/worksheet


lavigne958:
**Is your feature request related to a problem? Please describe.**
Updating spreadsheet property call does not update local properties.

ex: call to `append_row` does not increase property `row_count`

**Describe the solution you'd like**
Update local properties when necessary.

**Describe alternatives you've considered**
One can make the explicit call to `sht.fetch_sheet_metadata()` to update the properties with latest known values from the API.

**Additional context**
**Warning** this only solves the issue for SpreadSheets accessed via a single script. If an other script or a human using a web browser accesses the SpreadSheet and updates it, then the values known localy are still invalid. 

If you are in that case then you must call `sht.fetch_sheet_metadata()` to prevent concurent accesses etc.

relates to #545

lavigne958:
# Issue
## This issue raises the following problems:

- update the spreadsheet (and underlying worksheet) is not reflected locally
- not having the latests state of the spreadsheet locally brings some unexpected behavior
  - updating the row or column count brings column/row insertion to the wrong place
  - updating the title can break any function that tries to get a value using a range (because ranges are build using sheet title)
- the spreadsheet can be updated online using the UI and the running script can't be notified about the changes.

# Requirements
## To solve we must ensure the following requirements:

- The use of gspread can be limited by the number of API calls, this should not be raised extensively otherwise users will be blocked and won't be able to send any more requests for some time.
- The local spreadsheet properties should reflect what has been update online.

# Possible solutions
## update local properties after successful update call
- for any method that updates the spreadsheet properties:
  - update the local properties

- solves the basics cases and bug pointed by issues linked to this issue.
- does not solves concurrent access to the spreadsheet (multiple scripts accessing the same spreadsheet)
- after an update, getting the same spreadsheet can result in different properties (so different behaviors) if spreadhseet has been updated in between

## fetch latest spreadsheet properties after every update call
- for any method that updates the spreadsheet properties
  - fetch the spreadsheet properties

- solves the basics cases and cases when someone else update the spreadsheet online (bypasses the running script).
- **increases the number of API** calls by a lot.
  - This is not suitable for production for heavy usage of google API

# Proposal
## Update local properties and provide convenient method to fetch properties

- When a method updates a properties online, on successful call update local property too
  - append a new column will increase the column count, etc...
- Provide a new method that will fetch the latest spreadsheet properties and sheet properties.
[...]

PR #1225, opened 2023-06-16
Fix/update-internal-properties

alifeee:
Changes discussed in #881

`gspread.Worksheet._properties` (and thus the `gspread.Worksheet` `@property` fields) are now updated whenever the following are called (as well as previous changes):

- `append_rows`
- `insert_rows`
- `insert_cols`
- `delete_dimension`

(as you can see, the leftover was mostly `gspread.Worksheet.row_count` and `[...].col_count`

also new tests added for `insert_cols` and `delete_cols` (before, only row tests existed)

### DIFF
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -698,7 +698,9 @@ class Spreadsheet:
             ]
         }
 
-        return self.batch_update(body)
+        res = self.batch_update(body)
+        self._properties["timeZone"] = timezone
+        return res
 
     def update_locale(self, locale):
         """Update the locale of the spreadsheet.
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -677,9 +677,9 @@ class Spreadsheet:
             ]
         }
 
-        response = self.batch_update(body)
+        res = self.batch_update(body)
         self._properties["title"] = title
-        return response
+        return res
 
     def update_timezone(self, timezone):
         """Updates the current spreadsheet timezone.
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -724,7 +724,9 @@ class Spreadsheet:
             ]
         }
 
-        return self.batch_update(body)
+        res = self.batch_update(body)
+        self._properties["locale"] = locale
+        return res
 
     def list_protected_ranges(self, sheetid):
         """Lists the spreadsheet's protected named ranges"""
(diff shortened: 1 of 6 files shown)


alifeee:
Final thoughts: we should consider if `gspread.Spreadsheet` has any internal properties that are not updated, in a similar vein. On a brief look, there are:

- `id`
- `title`
- `url`
- `creationTime`
- `lastUpdateTime`
- `updated`
- `timeZone`
- `locale`
- `sheet1`

These are not updated on property change. I see:

- `id`, `creationTime` are never updated (or shouldn't be?)
- `title`, `timeZone`, `locale` should be easy fixes (I will do these in #1225)
- I have not looked at the implementation of `lastUpdateTime`, but it seems like it would be complex to update. Perhaps an extra API call makes sense for this one (but then as a method rather than a `@property` - deprecate the property?)
- `sheet1`, `url`, `updated` do not directly depend on `_properties` so can be ignored

alifeee:
(based on comment https://github.com/burnash/gspread/issues/881#issuecomment-1591166832)

closes #881

alifeee:
looks the only methods we may have to edit are

- `gspread.Spreadsheet.update_title`
- `gspread.Spreadsheet.update_timezone`
- `gspread.Spreadsheet.update_locale`

I will do this soon.

and then the discussion of `lastUpdateTime` above - require @lavigne958 thoughts.

alifeee:
ok, done `title`, `timezone`, & `locale`

it is ready to merge now, just after wondering whether `lastUpdatedTime` should change. Current implementation:

https://github.com/burnash/gspread/blob/20a2209308f01eb12855307c58a7630923c9d537/gspread/spreadsheet.py#L62-L70

So it runs an API request if it is not already in `_properties`. Due to the nature of the property, it might be nice to run the API request every time, so that it is always up to date. However, this may be better suited to a function, i.e. `getLastUpdatedTime()`, so it's more obvious the API is used. Thoughts?

lavigne958 (approved):

lavigne958:
> 

That's hard to say, currently we request it if not present but as you mention it's not updated.

To me we could add a new method like you said `getLastUpdatedTime` but instead of `get` we coudl make it actually update the last update time, kind of like retrieve the last update time. this makes it obvious that the method makes an API call then the user still uses the property to get it. 

In a way we provide a way to update the property.

does that makes sense ? this way we can add it to the next release using the current PR without breaking anything.

PR #1229, opened 2023-06-18
Remove accepted_kwargs decorator

### LINKED ISSUE #1228: Remove decorator accepted_kwargs
## Overview
Remove the decorator `accepted_kwargs`. We don't need it and it prevent us from correctly typing the library and using `Enums`.

## Details

The decorator `accepted_kwargs` will filter given `kwargs`.
It will raise an esception is any unknown kwargs is given.
It will set a default value for the kwargs that are not set.

We don't need that, we can simply put them on the command line with a default value.  ( I wonder if that was here for python2 :thinking: )

Then this will allows us to properly use `Enum`s for the arguments like `ValueRenderOption` etc. 

This will make the function `filter_dict_value` obsolete too now as we can pass `None` values as params they won't have any effects.

**This changes should not impact the user**, after the changes the test suite must run like before.
Merging this in 6.0.0 with the rest of the major changes anyway, it is not necessary for the user in v5.Y.Z

## To remove

- function `filter_dict_value`
- decorator `accepted_kwargs`

This issue blocks: #1222

lavigne958:
Remove the function decorator accepted_kwargs.
We can simply put all the args on the method signature.
The function filter_dict_values is useless too now, we can set
the params with some None values that does not break the API calls.

This:
- simplifies the method signature
- allows rich editors to find and display the right method signature
  and not just `wrapper(*args, **kwargs)`.
- allows us to better type arguments
- allows us to set default value like any other method using `param=Defaultvalue`
- simplifies the code in general

closes #1228

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -19,7 +19,6 @@ from typing import (
     Dict,
     Iterable,
     List,
-    NewType,
     Optional,
     Tuple,
     TypeVar,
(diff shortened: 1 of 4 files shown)


alifeee:
This sounds great to me. Those function decorators were certainly a lot. As you say, were they for Python 2?

This is a great change to me ♥

I added a test, because I wasn't fully sure how python 3 dealt with kwarguments used as arguments. I thought we may as well leave it in.

alifeee (approved):

lavigne958:
> This sounds great to me. Those function decorators were certainly a lot. As you say, were they for Python 2?
> 
> This is a great change to me hearts
> 
> I added a test, because I wasn't fully sure how python 3 dealt with kwarguments used as arguments. I thought we may as well leave it in.

thanks for the test ! :star_struck: I did not think about it when removing code but that's a great addition.

yes arguments can be passed in the right order or named but in the order you want.

sure we'll leave it in, if we change the order of the parameters in the methods `Worksheet.update` or `Worksheet.get` we'll catch it with the test.

alifeee:
sounds good!

❤

PR #1231, opened 2023-06-19
Update-build-tools

### LINKED ISSUE #1218: Update `tox -e build` script
`tox -e build` runs [`setup.py`]

https://github.com/burnash/gspread/blob/fde332b9d59c4825207b5d4c56dfe41d299780e6/tox.ini#L44-L47

Some things in [`setup.py`] are outdated

[`setup.py`]: https://github.com/burnash/gspread/blob/master/setup.py

Here we can gather a list of things to update:

- [x] #1230
- [x] update the commands
- [x] update [required python versions](https://github.com/burnash/gspread/blob/fde332b9d59c4825207b5d4c56dfe41d299780e6/setup.py#L55) (automate? - linked to #1214)
- [ ] make sure we deliver the same generated packages for PyPi repository
- [x] add [maintainer](https://github.com/burnash/gspread/blob/fde332b9d59c4825207b5d4c56dfe41d299780e6/setup.py#L50-L51) (me) :)
- [x] [distutils.core](https://github.com/burnash/gspread/blob/fde332b9d59c4825207b5d4c56dfe41d299780e6/setup.py#L11) is deprecated

n.b., some items in the above list are not so clear. you should be able to edit this comment(?)

alifeee:
closes #1218

Move from using `setup.py` to `pyproject.toml`


alifeee:
I have replaced `setup.py` with `pyproject.toml`.

Changes:

- Build now uses [Flit](https://flit.pypa.io/en/latest/rationale.html)
- Build command is now `python -m build` via the [build package](https://pypa-build.readthedocs.io/en/stable/mission.html) (as per https://github.com/burnash/gspread/issues/1218#issuecomment-1584229136)
- Project metadata is now in `pyproject.toml`, as described by [Python documentation](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata)
- Project metadata is changed
  - New maintainer added (me)
  - Python requirements changed & simplified (see https://github.com/burnash/gspread/pull/1231#issuecomment-1597251441)
  - Long description removed -> replaced with readme.md

alifeee:
# PyPi

To see the comparison of old project metadata vs new project metadata practically, see the new project on test PyPi (note, package name is `gspreadalifeee` because I did not have permission to push to `gspread`)

old

https://test.pypi.org/project/gspread/

new

https://test.pypi.org/project/gspreadalifeee/

alifeee:
# Updated build files

To compare the new `sdist` files, you could download and compare from the test.pypi links above, or see this diff

https://github.com/alifeee/gspread-new-build/compare/old-build...main

alifeee:
This seems good to me. I cannot think of much more to check, it would be good to hear your thoughts. Then, I will check anything you think might cause an issue.

tox.ini
```diff
@@ -14,7 +14,7 @@ commands = pytest {posargs} tests/
 
 [flake8]
 extend-ignore = E203
-extend-exclude = .tox,./env
+extend-exclude = .tox,./env,./gspread/__init__.py
```
lavigne958, line 17: why do we skip `__init__.py` from flake8 ? :thinking:

lavigne958 (approved):
looks good to me, I Just have a question about linting and skipping files.

tox.ini
alifeee: Flit takes the `pyproject.toml`'s `project.description` from the docstring of the `__init__.py` file of the module. Thus, I changed this in https://github.com/burnash/gspread/commit/35639014f87c7b0977f1521d06a848cfb46fcebc.

You will see that at the top of the file was `#flake8: noqa` which skips the file when using flake8.

https://github.com/burnash/gspread/blob/97ab607bc3212e47157e706129743c586d342c95/gspread/__init__.py#L1

Thus, my change is just to move where this ignore is specified. The file was already ignored.

However, as for *why*, it is because `flake8` complains about unused imports in the `__init__.py` file:

```bash
.\dist\gspread\gspread\__init__.py:7:1: F401 '.auth.authorize' imported but unused
from .auth import (
^
```

etc., as many things are imported (so that users can do `import * from gspread`, and other reasons.

https://github.com/burnash/gspread/blob/a4789287ed7bbb0db5c42d0ba3c69c735fe08211/gspread/__init__.py#L7-L25

lavigne958: Oh ok that makes sense then to ignore it with the rest. 

It's better this way.

I see for the import. Let's skip it.

It's all good to me.

lavigne958 (approved):

Issue #1214, opened 2023-06-08
Update supported python versions (remove Python 3.7 Support)


alifeee:
Python 3.7 End of life is [2023-06-27](https://devguide.python.org/versions/)

Here we can gather the things that must be done for removing Python 3.7 support.

- [x] Remove Python 3.7 from [Workflows](https://github.com/burnash/gspread/tree/master/.github/workflows)
- [x] Approve #1150 (now #1236)
- [x] Approve #1165 
- [x] Update " Requires: Python >=3.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.* " on PyPi (automate?)
- [x] Update " Programming Language " on https://pypi.org/project/gspread/ (automate?)
- [x] bump sphinx to latest version (#1235 #1239)

alifeee:
@lavigne958 added to 6.0.0 because I believe this is a "breaking change" that you often mention...

alifeee:
If Python 3.7 is removed from the workflow I think it is definitely breaking

alifeee:
not sure what you mean by "tox classifier", do you mean this?

https://github.com/burnash/gspread/blob/20a2209308f01eb12855307c58a7630923c9d537/tox.ini#L1-L2

> everyone using gpsread on python 3.7 won't be able to use it.

This is the intention, no? For the 6.0.0 release at least, you will not be able to use Python 3.7. If you want to, you must specify gspread v5.

lavigne958:
> not sure what you mean by "tox classifier", do you mean this?
> 
> https://github.com/burnash/gspread/blob/20a2209308f01eb12855307c58a7630923c9d537/tox.ini#L1-L2
> 
> > everyone using gpsread on python 3.7 won't be able to use it.
> 
> This is the intention, no? For the 6.0.0 release at least, you will not be able to use Python 3.7. If you want to, you must specify gspread v5.

sorry my mistake, I meant: the classifier in `setup.py`

which is metadata for pypi to understand thinkgs like which version is compatible with this package.

lavigne958:
Add a new task:

- [ ] bump sphinx to latest version

alifeee:
> bump sphinx to latest version

done in https://github.com/burnash/gspread/commit/d7e968a4b48bdf6815146fe2dfdd91346b8c89da

alifeee:
somehow we should remember to put this in the changelog?

lavigne958:
[...]

PR #1235, opened 2023-06-24
add read the doc configuration file

### LINKED ISSUE #1220: Update read the doc configuration
## Overview
Read the doc webiste will enforce all repos to have a configuration file

Update the repository with the appropriate configuration file for read the doc.

See: https://blog.readthedocs.com/migrate-configuration-v2/

## Details

Actually we are very late, we don't have any configuration file and they now enforce a configurationV2 file.

Need to be done by the end of the summer.

lavigne958:
Fix sphinx version to pass read the doc

closes #1220

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,25 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  fail_on_warning: true
+  configuration: docs/conf.py
+
+# We recommend specifying your dependencies to enable reproducible builds:
+# https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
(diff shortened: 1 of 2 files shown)


lavigne958:
I solved the issue with the build failing for python-3.7

I propose we merge it like this.

Because the version is fixed, dependabot will make a PR to update it. 

It will fails until we remove python-3.7 then we can merge a compatible version woch should be the latest one.

alifeee:
okay, sounds good to me.

I don't know how dependabot works properly, but if it will automatically see when 3.7 is removed and suggest a PR, then it sounds good to me.

Otherwise we ought to include our own issue to track the removal if it is not automatic.

Good stuff 👍

alifeee (approved):

lavigne958:
> okay, sounds good to me.
> 
> I don't know how dependabot works properly, but if it will automatically see when 3.7 is removed and suggest a PR, then it sounds good to me.
> 
> Otherwise we ought to include our own issue to track the removal if it is not automatic.
> 
> Good stuff 👍

It's not that clever.

It sees any pin version and offers a PR if a newer version is available.

When we merge this PR, it will offer a new PR with a bump in sphinx build to the latest version.

This will fail (due to python-3.7)

So we leave it until we remove python-3.7 and then we just ask dependabot to rebase it's pr and it should pass and we merge it 

I updated #1214 to keep track of it.

PR #1239, opened 2023-06-25
merge read the doc configuration file from master

alifeee:
(cherry picked from commit a659f88c9bfdd080ee4f30ca95b30ca94aeb7f6f)

this way we will not have to remember to update sphinx dependency, but it will be updated from the get-go

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,25 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  fail_on_warning: true
+  configuration: docs/conf.py
+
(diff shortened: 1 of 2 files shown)


alifeee:
I also make this instead of waiting for @dependabot, because it is the last item in #1214 so we can close that after.

lavigne958:
Hey that's a good idea to update that branch too. 

To le we should simply merge master branch into feature/6.0.0 that way we keep them both on track and the feature brench receives everything like the update on the license etc.

alifeee:
> we should simply merge master branch into feature/6.0.0

I did not do this because it would not be simple. Many merge conflicts since `master` has changed a fair amount from when `6.0.0` was forked off it.

However, the changes must be made eventually. However however, does it make more sense to merge `6.0.0` branch into `master` rather than the other way round? (and naturally we would do this later on, when the `6.0.0` release is imminent)

lavigne958:
you should not have too many changes to apply. I tend to merge master branch from time to time into `6.0.0`. 

I'll make a PR with the integration of the latest changes in `master` into `6.0.0`.

It makes sense as you mention it the day we actually release the version `6.0.0`. before that both branches will live their life along side each others. with the detail that `6.0.0` will receive all new changes from time time.

lavigne958:
Done in #1241 
We can close this PR.

PR #1241, opened 2023-06-26
merge master into 6.0.0

lavigne958:
- remove MANIFEST.in file It is supposed to declare files to include/exclude when built (with setuptools) I rebuilt the project without the file Files in MANIFEST.in remained in the source sdist It is no longer needed https://stackoverflow.com/questions/24727709/do-python-projects-need-a-manifest-in-and-what-should-be-in-it
- remove Makefile
- rewrite short description (__init__ docstring)
- update gitignore (formatting)
- add pyproject.toml
- ignore __init__.py with flake8 previously this was done in the file with `# flake8: noqa`
- remove setup.py and change build command
- (change tox build command)
- remove index.txt was used for old build, no longer needed
- Bump codespell from 2.2.4 to 2.2.5
- add method `refresh_lastUpdateTime`
- add test `test_refresh_lastUpdateTime`
- tox -e format
- update test cassette
- add read the doc configuration file
- Update requirements.txt
- update licence year
- remove deprecated methods from tests


docs/requirements.txt
```diff
@@ -1,3 +1,3 @@
-sphinx
+sphinx==5.3.0
```
alifeee, line 1: #1239 was created to update this to 6.2.1

lint-requirements.txt
```diff
@@ -1,5 +1,5 @@
 bandit==1.7.5
 black==23.3.0
-codespell==2.2.4
-flake8==6.0.0
-isort==5.12.0
+codespell==2.2.5
+flake8==5.0.4
+isort==5.11.4
```
alifeee, line 5: these should remain the newest versions, not be downgraded

alifeee (changes requested):

docs/requirements.txt
lavigne958: true I mixed up the conflicts. thanks

lint-requirements.txt
lavigne958: oh right I saw the version in `isort` I though that would keep the latest one but no I should keep the latest on those version.
That's why I put it into review :laughing:

lavigne958:
@alifeee if everything is ok for you i let you put your approval. Unless there are other things that need to be changed. So we can merge it.

alifeee (approved):

PR #1245, opened 2023-06-28
Refactor/spreadsheet-open-metadata

### LINKED ISSUE #1244: Unexpected Exception Thrown when Opening Inaccessible/Non-existent Spreadsheet using 'open_by_url' Method
**Description**
I encountered an unexpected behavior while working with the gspread (5.9.0) library, specifically when attempting to open an inaccessible or non-existent spreadsheet using the `open_by_url()` method. I expected the library to raise a `SpreadsheetNotFound` error in such cases, as it does when using the `open()` method. However, instead of the expected exception, the library throws an `APIError`.

**To Reproduce**
1. Import the gspread library and the necessary dependencies.
2. Use the open_by_url method with the URL of an inaccessible or non-existent spreadsheet.
3. Observe the exception thrown by the library.

**Expected behavior**
I expected the open_by_url method to raise a SpreadsheetNotFound exception when attempting to open an inaccessible or non-existent spreadsheet.

Actual Behavior:
Contrary to expectations, the open_by_url method throws an APIError when provided with the URL of an inaccessible or non-existent spreadsheet.

**Code example***
```
>>> gc = gspread.service_account_from_dict(service_account_key)
>>> sh = gc.open_by_url("https://docs.google.com/spreadsheets/d/non-existent-url/edit")
>>> sh.worksheets()
gspread.exceptions.APIError: {'code': 404, 'message': 'Requested entity was not found.', 'status': 'NOT_FOUND'}
```

**Environment info:**
- Operating System: MacOS
- Python version: 3.11
- gspread version: 5.9.0

alifeee:
closes #1244

- `open_by_key` (and thus `open_by_url`) now raises `SpreadsheetNotFound` exception instead of `APIError` if spreadsheet does not exist (i.e., API error 404)
- spreadsheet metadata is loaded *before* spreadsheet creation. not sure if you want to keep this change. let me know.*

*: if we keep the old style, then there will be 2 API calls when you open a spreadsheet, instead of one.


alifeee:
I am not sure why tests fail. They work if `GS_CREDS_FILENAME` is set. I think it is something to do with the new code I added here:

https://github.com/burnash/gspread/blob/3278e541a11ebf3003aba6085c208fe4fe498d85/tests/client_test.py#L13-L21

Would appreciate help @lavigne958, have you seen this before?

lavigne958:
I would suggest we keep the call to `fetch_sheet_metadata` in the init of a `Spreadsheet`. so we always do it.

and we place the `try` / `except` either in the `Spreadsheet.__init__()` method or around the `Spreadsheet(...)` creation.

for sure we should not call `fetch_sheet_metadata` outside of the spreadsheet init.

About the Permission denied: we can create a personal spreadsheet, then run the test and register the HTTP request + response. then delete the created spreadsheet so anyone reading the cassettes will only get the ID of a spreadsheet that does not exists. 

And we should remember to create a new one when recording the scripts.

lavigne958:
> I am not sure why tests fail. They work if `GS_CREDS_FILENAME` is set. I think it is something to do with the new code I added here:
> 
> https://github.com/burnash/gspread/blob/3278e541a11ebf3003aba6085c208fe4fe498d85/tests/client_test.py#L13-L21
> 
> Would appreciate help @lavigne958, have you seen this before?

no never, let me check :thinking:  :eyes: :mag:  :bug:

lavigne958:
and bTW I think we can merge this in the master branch it's too late for 5.10 to me, unless you feel like it can fit before tomorrow.

alifeee:
> I would suggest we keep the call to `fetch_sheet_metadata` in the init of a `Spreadsheet`. so we always do it.

Okay, yes. I will refactor it thusly now.

lavigne958:
> I am not sure why tests fail. They work if `GS_CREDS_FILENAME` is set. I think it is something to do with the new code I added here:
> 
> https://github.com/burnash/gspread/blob/3278e541a11ebf3003aba6085c208fe4fe498d85/tests/client_test.py#L13-L21
> 
> Would appreciate help @lavigne958, have you seen this before?

from what I see we don't have a cassette for these API call (create the file) so it tries to make the real request and it fails.

alifeee:
> and we place the try / except either in the Spreadsheet.__init__() method or around the Spreadsheet(...) creation.

Okay. I have refactored code now to work differently. Now, I only change the code in `gspread.client.open_by_key` method.

> About the Permission denied: we can create a personal spreadsheet, then run the test and register the HTTP request + response. then delete the created spreadsheet so anyone reading the cassettes will only get the ID of a spreadsheet that does not exists.

This makes sense. It may be forgotten when we update the test cassettes. I can try and add the functionality, and add a comment in the test case saying you have to create a personal spreadsheet.

> and bTW I think we can merge this in the master branch it's too late for 5.10 to me, unless you feel like it can fit before tomorrow.

I think it is a breaking change. People may be relying on `open_by_key` throwing an `APIError`, not a `SpreadsheetNotFound` error. This is why I added it to the next major version. Do you disagree?

> from what I see we don't have a cassette for these API call (create the file) so it tries to make the real request and it fails.

I am not sure how to fix this. I took the [`init` code](https://github.com/burnash/gspread/blob/27a97433018748ba3a02d4372df54aa4035d1d14/tests/client_test.py#L12C1-L21) from [`spreadsheet_test.py`'s `init`](https://github.com/burnash/gspread/blob/27a97433018748ba3a02d4372df54aa4035d1d14/tests/spreadsheet_test.py#L15-L22). So I am not sure why it works in `spreadsheet_test` but not `client_test`. The only different I see is that in `client_test`, the client is attached to `ClientTest` object, for use in tests. Maybe this messes with things?

I tried deleting all cassettes and reloading them (see 27a97433018748ba3a02d4372df54aa4035d1d14). This did not fix the problem. Not sure how to fix :/

alifeee:
Fixed the tests with 2c4216835020db178735271c6fb431c29583d0b9. Not sure of the details of why it works, but it does now.

Permission denied test is implemented, but the test is bad. It takes 260 seconds to complete. Not sure why - seems to hit the request timeout. It passes though. I have skipped it for now.

alifeee:
It is tricky to decide whether this should be in `5.11.0` or `6.0.0`.

The documentation for [`open`] says it throws a `SpreadsheetNotFoun` exception, so if *that* changed, that is a breaking change.

However, `open_by_key` does not say anything about exceptions, so from a "public API" standpoint, changing the exception would not "*change*" the documentation-as-API.

[`open`]: https://docs.gspread.org/en/latest/api/client.html#gspread.Client.open

lavigne958:
Hi I'll have a look at this PR today. 

We might be able to merge it in v5.11 in fact ( I need to check 😁) as you said, contract wise speaking we could add the new exception as it is not documented.

I'll have a look at the tests too, why it takes so long and why we don't get a 403  right away when trying to open a private spreadsheet.

lavigne958 (commented):
I ran some tests, and I noticed:

- the exception `APIError` derives from `GSpreadException`
- the exception `SpreadSheetNotFound` derives from `GSPreadException`

we can't make it land in v5.11 :-( 

the exception `SpreadSheetnotFound` does *not* derive from `APIError` so any user catching `APIError` will have the exception raising to the top level now that we raise `SpreadSheetNotFound` instead and it does not derive from `APIError`....

I thought we could move the exception `SpreadSheetNotFound` to derive form `APIError` but the `APIError` requires a response object as argument as it is a direct result from a HTTP API call, and in the location where we already use the exception `SpreadSheetNotFound` we don't have an API error as it is when we did not find the spreadsheet in the users folder.

We could make it happen by making the `repsonse` argument optional :thinking:  
or in the `SpreadSheetNotFound` could have some init function setting the response to some `None` object or some empty object. 

What do you think ?

alifeee:
> ...the exception SpreadSheetnotFound does not derive from APIError...

I changed the code so that now `SpreadsheetNotFound` *does* derive from `APIError`. Since `SpreadsheetNotFound` is only raised when an `APIError` is encountered, we can just use the response from that 

https://github.com/burnash/gspread/blob/64e066c74f64a06eb4c3303984f4074632f349fd/gspread/exceptions.py#L62-L67

https://github.com/burnash/gspread/blob/64e066c74f64a06eb4c3303984f4074632f349fd/gspread/client.py#L122-L124

This also made me think... `PermissionDenied`, `CellNotFound`, etc... should really probably *also* derive from `APIError` (I guess only the ones that are a direct result of an API failure, like `PermissionDenied`)? What do you think?

alifeee:
Test failure... because SpreadsheetNotFound in `Client.open(name)` doesn't derive from an APIError now... I see the issue

https://github.com/burnash/gspread/blob/64e066c74f64a06eb4c3303984f4074632f349fd/gspread/client.py#L99-L110

Does it make sense here to find a response object? It could be the response object containing the files found from Drive. That would make sense.

lavigne958:
> > ...the exception SpreadSheetnotFound does not derive from APIError...
> 
> I changed the code so that now `SpreadsheetNotFound` _does_ derive from `APIError`. Since `SpreadsheetNotFound` is only raised when an `APIError` is encountered, we can just use the response from that
> 
> https://github.com/burnash/gspread/blob/64e066c74f64a06eb4c3303984f4074632f349fd/gspread/exceptions.py#L62-L67
> 
> https://github.com/burnash/gspread/blob/64e066c74f64a06eb4c3303984f4074632f349fd/gspread/client.py#L122-L124
> 
> This also made me think... `PermissionDenied`, `CellNotFound`, etc... should really probably _also_ derive from `APIError` (I guess only the ones that are a direct result of an API failure, like `PermissionDenied`)? What do you think?

good job that's actually better this way.

yep the others should, I noticed `CellNotFound` is not used in the product :laughing:  can be removed without notice I believe lol

lavigne958:
> Test failure... because SpreadsheetNotFound in `Client.open(name)` doesn't derive from an APIError now... I see the issue
> 
> https://github.com/burnash/gspread/blob/64e066c74f64a06eb4c3303984f4074632f349fd/gspread/client.py#L99-L110
> 
> Does it make sense here to find a response object? It could be the response object containing the files found from Drive. That would make sense.
> 
> Only, `list_spreadsheet_files` would have to return the API response

that works for me but we still need to provide some `Response` object, do we create one emtpty ? we manually create one with some data ? I have not clue which is best so far.

I think when we catch the `StopIteration` it's good to create a new response object with some: 404, not found response (if we can).

alifeee:
okay, super. fixed. The changes now are:

- remove `CellNotFound` exception
- `open` and `open_by_key` now return `SpreadsheetNotFound` exception (if spreadsheet not found), which extends `APIError`
  - request is attached to SpreadshetNotFound as payload. Technically, `open` can make multiple requests, but only the latest is attached
- `open_by_key` returns `PermissionError` if client is not allowed to open the spreadsheet. This test is currently broken (see related issue)

gspread/exceptions.py
```diff
@@ -61,3 +56,11 @@ def _text_from_detail(
...
         except (AttributeError, KeyError, ValueError):
             return None
+
+
+class SpreadsheetNotFound(APIError):
+    """Trying to open non-existent or inaccessible spreadsheet."""
+
+    def __init__(self, response: Response):
+        super().__init__(response)
+        self.response: Response = response
```
lavigne958, line 66: you don't need to set the `self.response` object it's already set by the parent class `APIError`.

lavigne958 (commented):
great changes, it's very _chirurgical_ ! it adds the new exception but does not break a thing, well done.

lavigne958:
don't forget to change the destination branch to `master` so we merge it to master. you might need to merge master branch onto your work branch or rebase in order to update again the test cassettes

alifeee:
> don't forget to change the destination branch to master so we merge it to master. you might need to merge master branch onto your work branch or rebase in order to update again the test cassettes

I am getting very confused trying to do this, as master and release6.0.0 branches change how the `Client` object works a lot. I will leave this as it is for now, until I can stomach trying to attempt the rebase.

alifeee:
In line with the above, I replaced this with #1277. rebasing was too confusing as v5 and v6 have diverged a fair bit.

PR #1250, opened 2023-07-05
Remplace named tuples with enums

### LINKED ISSUE #1222: Use Enums instead of NamedTuple
## Overview

Currently we use `NamedTuple` for the static parameters to send to the API.
Replace with `Enum` instances, much better to deal with.

## Details

The named tuples are more complexe to type and to work with.

With the enums it's much easier and faster.

## Mention

Add warnings again in the latest release for the coming breaking change :grimacing:

lavigne958:
Most of the parameters for the API are `ENUM`s

Replace the named tuples that are not fitted for that with enums.

Each enum hold one of the choices.
The actual string value must be extracted in order to pass strings only to the API using the function `extract_enum_value` using the `enum.value` field if the enum is not `None`.

this forces us to filter all arguments that are passed when they can possibly be `None`.

thought about setting the default value for each instead but this changes the API call made to the API and that could lead to unwanted behavior for the users.

Possible way was to add a new value to each enum: `default = None` which will result in a `None` value so it won't be part of the API call, but this requires all method to have that value as default value, and if a user (for some reason) sets one of the parameter to `param=None` then the code breaks and requires that user to change to `param=MyEnum.none` which is not user friendly. To me any function with default empty param should be either `None` or some default container like `param=[]` etc.

so far this is the less intrusive for users and keeps the API calls as is and use the default behavior from google.

closes #1222

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -9,7 +9,6 @@ This module contains utility functions.
 import re
 from collections import defaultdict
 from collections.abc import Sequence
-from enum import Enum
 from functools import wraps
 from itertools import chain
 from typing import (
@@ -20,7 +19,6 @@ from typing import (
     Dict,
     Iterable,
     List,
-    MutableMapping,
     Optional,
     Tuple,
     TypeVar,
@@ -31,6 +29,7 @@ from urllib.parse import quote as uquote
 from google.auth.credentials import Credentials as Credentials
 from google.oauth2.credentials import Credentials as UserCredentials
 from google.oauth2.service_account import Credentials as ServiceAccountCredentials
+from strenum import StrEnum
 
 from .exceptions import IncorrectCellLabel, InvalidInputValue, NoValidUrlKeyFound
 
@@ -46,28 +45,28 @@ URL_KEY_V1_RE = re.compile(r"key=([^&#]+)")
 URL_KEY_V2_RE = re.compile(r"/spreadsheets/d/([a-zA-Z0-9-_]+)")
 
 
-class Dimension(Enum):
+class Dimension(StrEnum):
     rows = "ROWS"
     cols = "COLUMNS"
 
 
-class ValueRenderOption(Enum):
+class ValueRenderOption(StrEnum):
     formatted = "FORMATTED_VALUE"
     unformatted = "UNFORMATTED_VALUE"
     formula = "FORMULA"
 
 
-class ValueInputOption(Enum):
+class ValueInputOption(StrEnum):
     raw = "RAW"
     user_entered = "USER_ENTERED"
 
 
-class DateTimeOption(Enum):
+class DateTimeOption(StrEnum):
     serial_number = "SERIAL_NUMBER"
     formatted_string = "FORMATTED_STRING"
 
 
-class MimeType(Enum):
+class MimeType(StrEnum):
     google_sheets = "application/vnd.google-apps.spreadsheet"
     pdf = "application/pdf"
     excel = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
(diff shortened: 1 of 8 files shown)


gspread/utils.py
```diff
@@ -43,69 +45,60 @@
...
+
+
+class ValueInputOption(Enum):
+    raw = "RAW"
+    user_entered = "USER_ENTERED"
+
+
+class DateTimeOption(Enum):
+    serial_number = "SERIAL_NUMBER"
+    formatted_string = "FORMATTED_STRING"
```
alifeee, line 67: should we consider including `formated_string` for backwards compatibility? probably not.

gspread/worksheet.py
```diff
@@ -1062,18 +1079,25 @@ def update(
...
+                "valueInputOption": value_input_option,
+                "includeValuesInResponse": include_values_in_response,
+                "responseValueRenderOption": response_value_render_option,
+                "responseDateTimeRenderOption": response_date_time_render_option,
+            }
+        )
+
+        # get actual value only if not None
+        major_dimension_value = (
+            major_dimension.value if major_dimension is not None else None
```
alifeee, line 1093: why not use extract_enum_values here?

alifeee (approved):

alifeee:
I cannot find the comment we discussed it originally, but why not use strEnums, so that you do not have to use `.value`?

gspread/worksheet.py
lavigne958: True I could, I will

lavigne958:
> I cannot find the comment we discussed it originally, but why not use strEnums, so that you do not have to use `.value`?

We can't use `StrEnum` they were introduced in python3.11 only :-( I though about it, it would make life much easier !

gspread/worksheet.py
lavigne958: I remember why I did not use the function here:
- the function `extract_enum_values` only takes dict as argument
- So I need to create a dict just to pass the single value
- then get the resulting dict
- if a value put it, if no value put `None`

I felt like this was easier to put

lavigne958:
> > We can't use StrEnum they were introduced in python3.11 only :-( I though about it, it would make life much easier !
> 
> What about with StrEnum: https://pypi.org/project/StrEnum/
> 
> It supports Python 3.7+

wow awesome ! :star_struck:  I'll update the code to use it then it makes like so much easier (If everything goes according to plan lol)

gspread/worksheet.py
```diff
@@ -1186,12 +1198,13 @@ def batch_update(
...
         body = {
             "valueInputOption": value_input_option,
-            "includeValuesInResponse": include_values_in_response,
             "responseValueRenderOption": response_value_render_option,
             "responseDateTimeRenderOption": response_date_time_render_option,
-            "data": data,
         }
 
+        body["includeValuesInResponse"] = include_values_in_response
+        body["data"] = data
```
alifeee, line 1206: these two `body` modifications can go in the object now, instead of afterwards (they were not before because of parse_enum function)

gspread/worksheet.py
```diff
@@ -1681,9 +1695,11 @@ def append_rows(
         params = {
             "valueInputOption": value_input_option,
             "insertDataOption": insert_data_option,
-            "includeValuesInResponse": include_values_in_response,
         }
 
+        # Not an Enum, does not pass `extract_enum_values`
+        params["includeValuesInResponse"] = include_values_in_response
```
alifeee, line 1701: can this go back in the `params` object?

lavigne958: Yes or can. Good catch.

gspread/worksheet.py
lavigne958: Yes it can too. I forgot with  all the back and forth

gspread/worksheet.py
lavigne958: done

gspread/worksheet.py
lavigne958: done

lavigne958:
you already approved but as I made changes I wait for your review before merging.

alifeee (approved):
lgtm!

it's nice that worksheet only has docs changes now that the enums can be used where strings are used.

question: why all the whitespace change?

image

ready to merge! good work! ❤

lavigne958:
> lgtm!
> 
> it's nice that worksheet only has docs changes now that the enums can be used where strings are used.

Yes that's right, so much easier.

> 
> question: why all the whitespace change?

Because the RST documentation format works with indentation. The length (number of whitespaces) need to create a block, a list, a warning etc is aligned with the parent block. All of those where not aligned so I aligned them to properly render the documentation.
> 
> image
> 
> ready to merge! good work! ❤

PR #1254, opened 2023-07-15
Feature/add type hints worksheets

lavigne958:
Add type hints for `Worksheet` object.

It's complete, but the majority's of it is done. It could be improved with time and when all files have been typed.

It passez the test suite, it passes mypy (some errors are still showing due to missing type annotations on some files).

Later on some of the types here will be improved when I type `Spreadsheet` object and once merged I will be able to better type the functions in `utils.py` etc.

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -3021,8 +3021,8 @@ class Worksheet:
         self,
         source: str,
         dest: str,
-        paste_type:PasteType=PasteType.normal,
-        paste_orientation:PasteOrientation=PasteOrientation.normal,
+        paste_type: PasteType = PasteType.normal,
+        paste_orientation: PasteOrientation = PasteOrientation.normal,
     ) -> JSONAnyType:
         """Copies a range of data from source to dest
 
@@ -3066,7 +3066,7 @@ class Worksheet:
         self,
         source: str,
         dest: str,
-        paste_type:PasteType=PasteType.normal,
+        paste_type: PasteType = PasteType.normal,
     ) -> JSONAnyType:
         """Moves a range of data form source to dest
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -6,6 +6,8 @@ This module contains common worksheets' models.
 
 """
 
+from typing import Any, Mapping, Optional, TypeVar, Type, MutableMapping
+
 from .cell import Cell
 from .exceptions import GSpreadException
 from .urls import WORKSHEET_DRIVE_URL
@@ -27,6 +29,8 @@ from .utils import (
     numericise_all,
     rowcol_to_a1,
 )
+
+T = TypeVar("T", bound="ValueRange")
 
 
 class ValueRange(list):
@@ -65,8 +69,10 @@ class ValueRange(list):
        It will be instantiated using the response from the sheet API.
     """
 
+    _json: MutableMapping[str, Any] = {}
+
     @classmethod
-    def from_json(cls, json):
+    def from_json(cls: Type[T], json: Mapping[str, Any]) -> T:
         values = json.get("values", [])
         new_obj = cls(values)
         new_obj._json = {
@@ -77,12 +83,12 @@ class ValueRange(list):
         return new_obj
 
     @property
-    def range(self):
+    def range(self) -> str:
         """The range of the values"""
         return self._json["range"]
 
     @property
-    def major_dimension(self):
+    def major_dimension(self) -> str:
         """The major dimension of this range
 
         Can be one of:
@@ -92,7 +98,7 @@ class ValueRange(list):
         """
         return self._json["majorDimension"]
 
-    def first(self, default=None):
+    def first(self, default: Optional[str] = None) -> Optional[str]:
         """Returns the value of a first cell in a range.
 
         If the range is empty, return the default value.
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -61,7 +61,7 @@ CellFormat = TypedDict(
 
 BatchData = TypedDict("BatchData", {"range": str, "values": List[List[Any]]})
 
-JSONAnyType = Mapping[str, Any]
+JSONResponse = MutableMapping[str, Any]
 ValueRangeType = TypeVar("ValueRangeType", bound="ValueRange")
 
 
@@ -734,7 +734,7 @@ class Worksheet:
         except KeyError:
             return []
 
-    def update_acell(self, label: str, value: Union[int, float, str]) -> JSONAnyType:
+    def update_acell(self, label: str, value: Union[int, float, str]) -> JSONResponse:
         """Updates the value of a cell.
 
         :param str label: Cell label in A1 notation.
@@ -748,7 +748,7 @@ class Worksheet:
 
     def update_cell(
         self, row: int, col: int, value: Union[int, float, str]
-    ) -> JSONAnyType:
+    ) -> JSONResponse:
         """Updates the value of a cell.
 
         :param int row: Row number.
@@ -1018,7 +1018,7 @@ class Worksheet:
         include_values_in_response: Optional[bool] = None,
         response_value_render_option: Optional[ValueRenderOption] = None,
         response_date_time_render_option: Optional[DateTimeOption] = None,
-    ) -> JSONAnyType:
+    ) -> JSONResponse:
         """Sets values in a cell range of the sheet.
 
         :param list values: The data to be written in a matrix format.
@@ -1153,7 +1153,7 @@ class Worksheet:
         include_values_in_response: Optional[bool] = None,
         response_value_render_option: Optional[ValueRenderOption] = None,
         response_date_time_render_option: Optional[DateTimeOption] = None,
-    ) -> JSONAnyType:
+    ) -> JSONResponse:
         """Sets values in one or more cell ranges of the sheet at once.
 
         :param list data: List of dictionaries in the form of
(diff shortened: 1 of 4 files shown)


gspread/worksheet.py
```diff
@@ -195,7 +238,9 @@ def _get_sheet_property(self, property, default_value):
 
         return sheet.get(property, default_value)
 
-    def acell(self, label, value_render_option=ValueRenderOption.formatted):
+    def acell(
+        self, label: str, value_render_option=ValueRenderOption.formatted
```
alifeee, line 242: value_render_option is not Typed

gspread/worksheet.py
```diff
@@ -217,7 +262,9 @@ def acell(self, label, value_render_option=ValueRenderOption.formatted):
             *(a1_to_rowcol(label)), value_render_option=value_render_option
         )
 
-    def cell(self, row, col, value_render_option=ValueRenderOption.formatted):
+    def cell(
+        self, row: int, col: int, value_render_option=ValueRenderOption.formatted
```
alifeee, line 266: value_render_option is not Typed

gspread/worksheet.py
```diff
@@ -244,14 +291,14 @@ def cell(self, row, col, value_render_option=ValueRenderOption.formatted):
                 rowcol_to_a1(row, col), value_render_option=value_render_option
             )
 
-            value = data.first()
+            value = data.first() or ""
```
alifeee, line 294: why this change?

gspread/worksheet.py
```diff
@@ -948,20 +1005,20 @@ def batch_get(
...
-        values=None,
-        raw=True,
-        major_dimension=None,
-        value_input_option=None,
-        include_values_in_response=None,
-        response_value_render_option=None,
-        response_date_time_render_option=None,
-    ):
+        values: Iterable[Iterable[Any]],
+        range_name: Optional[str] = None,
```
alifeee, line 1009: the first two args switch. is there a depracation warning in this code? I can't see it

(is it just not in this branch?)

gspread/worksheet.py
```diff
@@ -1620,12 +1678,12 @@ def add_cols(self, cols):
 
     def append_row(
         self,
-        values,
+        values: Sequence[Union[str, int, float]],
         value_input_option=ValueInputOption.raw,
```
alifeee, line 1682: value_input_option is not Typed

gspread/worksheet.py
```diff
@@ -1659,12 +1717,12 @@ def append_row(
 
     def append_rows(
         self,
-        values,
+        values: Sequence[Sequence[Union[str, int, float]]],
         value_input_option=ValueInputOption.raw,
```
alifeee, line 1721: value_input_option is not Typed

gspread/worksheet.py
```diff
@@ -1706,11 +1764,11 @@ def append_rows(
 
     def insert_row(
         self,
-        values,
-        index=1,
+        values: Sequence[Union[str, int, float]],
+        index: int = 1,
         value_input_option=ValueInputOption.raw,
```
alifeee, line 1769: value_input_option is not Typed

gspread/worksheet.py
```diff
@@ -1745,11 +1803,11 @@ def insert_row(
 
     def insert_rows(
         self,
-        values,
-        row=1,
+        values: Sequence[Sequence[Union[str, int, float]]],
+        row: int = 1,
         value_input_option=ValueInputOption.raw,
```
alifeee, line 1808: value_input_option is not Typed

gspread/worksheet.py
```diff
@@ -1817,11 +1875,11 @@ def insert_rows(
 
     def insert_cols(
         self,
-        values,
-        col=1,
+        values: Sequence[Sequence[Union[str, int, float]]],
+        col: int = 1,
         value_input_option=ValueInputOption.raw,
```
alifeee, line 1880: value_input_option is not Typed

gspread/worksheet.py
```diff
@@ -1883,13 +1941,13 @@ def insert_cols(
...
         self,
-        name,
-        editor_users_emails,
-        editor_groups_emails=[],
-        description=None,
-        warning_only=False,
-        requesting_user_can_edit=False,
-    ):
+        name: str,
+        editor_users_emails: str,
```
alifeee, line 1945: it seems to me this should be Sequence[str]

gspread/worksheet.py
```diff
@@ -2244,13 +2345,13 @@ def clear_basic_filter(self):
...
-        new_sheet_id=None,
-        new_sheet_name=None,
-    ):
+        client: HTTPClient,
+        spreadsheet_id: str,
+        sheet_id: int,
+        insert_sheet_index: Optional[int] = None,
+        new_sheet_id: Optional[int] = None,
+        new_sheet_name: Optional[str] = None,
+    ) -> "Worksheet":
```
alifeee, line 2354: unneccessary quotation marks?

gspread/worksheet.py
```diff
@@ -2330,10 +2434,12 @@ def copy_to(
...
         :rtype: dict
         """
-        return self.client.spreadsheets_sheets_copy_to(spreadsheet_id, self.id)
+        return self.client.spreadsheets_sheets_copy_to(
+            self.spreadsheet_id, self.id, destination_spreadsheet_id
+        )
 
     @cast_to_a1_notation
-    def merge_cells(self, name, merge_type="MERGE_ALL"):
+    def merge_cells(self, name: str, merge_type: str = "MERGE_ALL") -> JSONAnyType:
```
alifeee, line 2442: it seems as if merge_type could be an Enum, if there are only 3 options.

gspread/worksheet.py
```diff
@@ -2886,21 +3004,21 @@ def _set_gridlines_hidden_flag(self, hidden):
...
         return self._set_gridlines_hidden_flag(False)
 
     def copy_range(
         self,
-        source,
-        dest,
+        source: str,
+        dest: str,
         paste_type=PasteType.normal,
         paste_orientation=PasteOrientation.normal,
```
alifeee, line 3020: paste_type and paste_orientation are not typed

gspread/worksheet.py
```diff
@@ -2941,10 +3059,10 @@ def copy_range(
 
     def cut_range(
         self,
-        source,
-        dest,
+        source: str,
+        dest: str,
         paste_type=PasteType.normal,
```
alifeee, line 3064: paste_type is not typed

alifeee (changes requested):
This is a big change! You've done great work with it ❤

I have added a few comments.

I also wonder: `JSONAnyType` is returned by a lot of the functions like `update_title`. This seems a bit vague. I am not sure where else it is used, but when we `return response` (which we do a lot), perhaps we could change the name to `HTTPResponseType`, or the default `import requests` response type? In my opinion this would make it more obvious when a method is returning the (raw) HTTP response, rather than a "useful" JSON object.

tests/cell_test.py
```diff
@@ -82,7 +82,7 @@ def test_a1_value(self):
 
     @pytest.mark.vcr()
     def test_merge_cells(self):
-        self.sheet.update("A1:B2", [[42, 43], [43, 44]])
+        self.sheet.update([[42, 43], [43, 44]], "A1:B2")
```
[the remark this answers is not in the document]
lavigne958: yes this is the method.

lavigne958 (commented):
thanks for the review that's a good catch. I added the missing types and updated the wrong types.

gspread/worksheet.py
alifeee: I see. It makes sense to me to cast the value to a string with `str(value)`. Otherwise, the `or ""` exists purely for type-checking purposes, and doesn't actually do anything.

However, the return type of `first` is "str" or "None", so with that it makes sense. But... *is* the return type of `first` *always* "str" or "None"?

gspread/worksheet.py
alifeee: sounds good :) we can resolve this conversation when issue created

alifeee:
> I thought may be we want to be exaustif and create a new type with all possible values inside from the API but that means we need to maintain it ! and we don't look at the response very often we simply return it to the user so no point to me but it's open for discussion 😆

This sounds like a lot to me, and probably not worth it...! Just having the JSON Any type makes sense, but perhaps with a different name like JSONResponse :)

lavigne958:
> > I thought may be we want to be exaustif and create a new type with all possible values inside from the API but that means we need to maintain it ! and we don't look at the response very often we simply return it to the user so no point to me but it's open for discussion laughing
> 
> This sounds like a lot to me, and probably not worth it...! Just having the JSON Any type makes sense, but perhaps with a different name like JSONResponse :)

agreed we can rename it `JSONResponse`. 
We can merge this PR, then after it I can start working on better typing between files now that they will be almost all typed.
I'll may be able to add some typed-dict etc too. We'll see :upside_down_face:

gspread/worksheet.py
lavigne958: I need to check that, let's cast it to string first. The `value` of a `Cell` object is set to `str` for now I'll update it if I can find some code path that can set it to any value (like `int` etc..)

gspread/worksheet.py
lavigne958: done see #1259

alifeee (approved):

lavigne958:
Alright this is a go for merge ! :partying_face:

PR #1255, opened 2023-07-15
Remove API calls from `creationTime`/`lastUpdateTime`

### LINKED ISSUE #1253: Avoid doing i/o in properties
**Describe the bug**
There are a handful of `@property` methods on the Spreadsheet object that do i/o under the hood. This isn't a great idea for all the same reasons that were listed in #600. There are a bunch of ways forward here but it seems sensible to introduce new methods as replacements, depreciate the current properties, and eventually remove the properties in gspread 6.x.

Properties:
- creationTime
- lastUpdateTime
- sheet1

alifeee:
closes #1253 

`creationTime` and `lastUpdateTime` were `@property`s. But, the first time they were used they called the API

https://github.com/burnash/gspread/blob/d0f1380a78bf12b342897cd5771cd84c4581b43b/gspread/spreadsheet.py#L52-L60

This is because they are both **Drive metadata**, and not **Sheets metadata**. Thus, the drive metadata was fetched the first time the `@property`s were accessed.

This PR changes the behaviour of this. In my opinion, there are several options to fix the problem of "Avoid doing i/o in properties" (issue #1253):

1. Remove `@property`, i.e., change them to methods
  I don't particularly like this since the API call is only made one time, the first time. After that, they act as properties
1. Return `None` if they do not exist, and expect the user to use `refresh_lastUpdateTime` or `get_lastUpdateTime`
  This is not very user-friendly, imo.
1. Throw an `Exception` if they do not exist, and tell the user to use `refresh_lastUpdateTime` or `get_lastUpdateTime`
  Again, not very user friendly
1. Fetch the properties in `Spreadsheet.__init__`
  This means they can be `@property`s just like the other ones (`id`, `title`, etc.). However, it means another API call when opening a spreadsheet

I went with option 4 for this PR. Please convince me otherwise if you think another idea is better, or have a different idea.

This solution introduces another API call, but otherwise does not change the public API, as we do not have to deprecate anything*. 

- It turns 1 API calls on spreadsheet initialisation to 2
  If using the properties anyway, it would be 2 as soon as you try to access the properties, or 3 if using both.
- I am not sure how auth works properly, but it could introduce problems if a user/key is allowed to access the Sheets API but ***not*** the Drive API (!!!)

*: however, I do suggest deprecating `lastUpdateTime` in favour of `get_lastUpdateTime()` (which calls the API), since this is something that a user probably always wants the latest version of

todo:

- [x] deprecate lastUpdateTime (add warning/another issue)
- [x] fetch_sheet_metadata should fetch drive api too???


alifeee:
tests currently broken because of new API request in `Spreadsheet.__init__`. cassettes need refreshing

alifeee:
@lavigne958 Before I continue with this I'd like your thoughts on the above, no rusg :)

gspread/spreadsheet.py
```diff
@@ -743,7 +745,11 @@ def list_protected_ranges(self, sheetid):
...
-        """Refresh the lastUpdateTime property of the spreadsheet."""
-        metadata = self.client._get_file_drive_metadata(self.id)
+    def get_lastUpdateTime(self):
+        """Get the lastUpdateTime metadata from the Drive API.
+        Also updates the cached value in the _properties dict.
+        """
+        metadata = self.client.get_file_drive_metadata(self.id)
+        # remove this upon deprecation of lastUpdateTime @property
         self._properties["modifiedTime"] = metadata["modifiedTime"]
+        return metadata["modifiedTime"]
```
lavigne958, line 755: if this is just a matter of API, may be we can keep both options os users already using the `refresh_lastUpdateTime` will keep using it as it is and can keep using the property. Users interedted in the new method can use it too along side the refresh ethod.

this way I suppose we could be able to merge it in v5 right ? :thinking:

alifeee:
Moving this to here instead of in a review comment...?

> Then, yes... we could merge it into v5... if you are happy to introduce an extra API request on Spreadsheet init :)

alifeee:
Two questions:

should I add a fetch of the Drive metadata to `fetch_sheet_metadata`?

https://github.com/burnash/gspread/blob/ba162326543dd487986808f18f46562f52211e48/gspread/spreadsheet.py#L239-L247

should I turn this warning into a deprecation warning?

https://github.com/burnash/gspread/blob/ba162326543dd487986808f18f46562f52211e48/gspread/spreadsheet.py#L62-L73

lavigne958:
> Two questions for @lavigne958:
> 
> should I add a fetch of the Drive metadata to `fetch_sheet_metadata`?

I would go for: keep it next to it, it's ok to have both method next ot each others. we only call them in the init method.

> 
> https://github.com/burnash/gspread/blob/ba162326543dd487986808f18f46562f52211e48/gspread/spreadsheet.py#L239-L247
> 
> should I turn this warning into a deprecation warning?
I would say YES, but I feel like we are removing so much code :see_no_evil: 
> 
> https://github.com/burnash/gspread/blob/ba162326543dd487986808f18f46562f52211e48/gspread/spreadsheet.py#L62-L73

alifeee:
> I would go for: keep it next to it, it's ok to have both method next ot each others. we only call them in the init method.

I do not understand this. The options are:

- fetch drive metadata in spreadsheet init
- fetch drive metadata in spreadsheet.fetch_sheet_metadata

I do the first. I will stick with that for now, because to do otherwise would mean refreshing the cassettes again...

> I would say YES, but I feel like we are removing so much code 🙈

This is fair. Perhaps we just leave in a warning

lavigne958:
> > I would go for: keep it next to it, it's ok to have both method next ot each others. we only call them in the init method.
> 
> I do not understand this. The options are:
> 
> * fetch drive metadata in spreadsheet init
> * fetch drive metadata in spreadsheet.fetch_sheet_metadata
> 
> I do the first. I will stick with that for now, because to do otherwise would mean refreshing the cassettes again...
Agreed, to me it looks good like this. the first option you suggest is fine.

> 
> > I would say YES, but I feel like we are removing so much code see_no_evil
> 
> This is fair. Perhaps we just leave in a warning
yep that's fine like that, it does not break a thing, people can use the new function, sounds nice for everyone.

alifeee:
okay, super

Changes are:

- rename `Client._get_file_drive_metadata` to `Client.get_file_drive_metadata` (change to public method, as it is required to be used in `Spreadsheet.__init__()`.
- Fetch Drive metadata on `Spreadsheet.__init__()`. This adds an API call, but means that `creationTime` and `lastUpdateTime` properties of `Spreadsheet` are now static methods (this is the original desire from the issue #1253)
- make `Spreadsheet.creationTime` and `Spreadsheet.lastUpdateTime` static properties
- add warning for `Spreadsheet.lastUpdateTime`. Users should instead use `Spreadsheet.get_lastUpdateTime()`
- add `Spreadsheet.get_lastUpdateTime()`
- add tests

lavigne958 (approved):

lavigne958:
12k lines of diff, nothing to see here, please keep going 😂

Great job l'été get this merge. So it start adding content for next release. 😁

PR #1269, opened 2023-08-08
Fix Worksheet ID Type Inconsistencies

### LINKED ISSUE #1267: Worksheet ID Type Inconsistencies
**Describe the bug**
On a `Spreadsheet` instance when operating on the underlying worksheet objects sometimes a the worksheet id can be a string (`del_worksheet_by_id`), other times it must be an int (`get_worksheet_by_id`).

**Expected behavior**
Ability to get a worksheet off a spreadsheet with any valid integer or integer-like string.  I frequently store Worksheet indexes in environment variables so if I forget to wrap them in `int()` I get a needless error.

**Code example**
```python
sht = client.open('My fancy spreadsheet')
worksheet = sht.get_worksheet(1234) # this works

worksheet = sht.get_worksheet("1234") # throws gspread.exceptions.WorksheetNotFound
```

**Environment info:**
- Operating System: macOS
- Python: 3.11.4
- gspread: 5.10.0

**Proposed Solution**
https://github.com/burnash/gspread/blob/739b8fca7e91d798d71a446e8849f42b950c2840/gspread/spreadsheet.py#L271
```python
def get_worksheet_by_id(self, id):
        # ...
        sheet_data = self.fetch_sheet_metadata()

        try:
            item = finditem(
                lambda x: x["properties"]["sheetId"] == int(id), # wrap id argument with int converter
                sheet_data["sheets"],
            )
            return Worksheet(self, item["properties"])
        except (StopIteration, KeyError, ValueError): # add Value Error in case int conversion fails
            raise WorksheetNotFound("id {} not found".format(id))
```
It might be clearer to break out the ValueError into it's own exception handler but I don't know what the ideal error to raise would be.  Happy to open a PR for this, just looking for some guidance first.

FlantasticDan:
Implements #1267 

I'm not super familiar with VCR tests so I hope adding `allow_playback_repeats=True` is acceptable in this case.

closes #1267


alifeee:
thanks for the PR! Great work :)

Some automated tests fail because of old Python versions. You need to replace `int | str` with `Union[int, str]` and `from typing import Union` (`|` was added Python 3.10, so earlier versions fail)

For the tests, try this:

- add your credentials to `tests/creds.json`
- run in the console
  ```bash
  GS_CREDS_FILENAME="tests/creds.json"
  GS_RECORD_MODE="all"
  tox -e py -- -k test_get_worksheet_by_id -v -s
  ```

This will run the test online (with real API requests).

When you are happy - or the test passes - either clear the env variables or start a fresh console and run the test command(s) again. Let me know if you need any more help with testing.

Again thanks for the PR! I have left some other comments as code comments

gspread/spreadsheet.py
```diff
@@ -287,11 +287,11 @@ def get_worksheet_by_id(self, id):
 
         try:
             item = finditem(
-                lambda x: x["properties"]["sheetId"] == id,
+                lambda x: x["properties"]["sheetId"] == int(id),
```
alifeee, line 290: as discussed, I think this should be `str(id)`

gspread/spreadsheet.py
```diff
@@ -287,11 +287,11 @@ def get_worksheet_by_id(self, id):
...
                 sheet_data["sheets"],
             )
             return Worksheet(self, item["properties"])
-        except (StopIteration, KeyError):
+        except (StopIteration, KeyError, ValueError):
             raise WorksheetNotFound("id {} not found".format(id))
```
alifeee, line 295: This should be a separate except clause. However, `str(id)` should not throw a ValueError so could also be removed.

gspread/spreadsheet.py
```diff
@@ -437,12 +437,12 @@ def del_worksheet(self, worksheet):
 
         return self.batch_update(body)
 
-    def del_worksheet_by_id(self, worksheet_id: str):
+    def del_worksheet_by_id(self, worksheet_id: int | str):
```
alifeee, line 440: see above comment about Union

gspread/spreadsheet.py
```diff
@@ -437,12 +437,12 @@ def del_worksheet(self, worksheet):
...
         """
         Deletes a Worksheet by id
         """
 
-        body = {"requests": [{"deleteSheet": {"sheetId": worksheet_id}}]}
+        body = {"requests": [{"deleteSheet": {"sheetId": str(worksheet_id)}}]}
```
alifeee, line 445: you could also extract this beforehand as

```python
worksheet_id_str = str(worksheet_id)
```

just so it is obvious what is happening to it. But this is fine too.

tests/spreadsheet_test.py
```diff
@@ -36,10 +36,12 @@ def test_get_worksheet(self):
         sheet1 = self.spreadsheet.get_worksheet(0)
         self.assertTrue(isinstance(sheet1, gspread.Worksheet))
 
-    @pytest.mark.vcr()
+    @pytest.mark.vcr(allow_playback_repeats=True)
```
alifeee, line 39: with my advice (see [comment](https://github.com/burnash/gspread/pull/1269#issuecomment-1671494535)) you should not need this

tests/spreadsheet_test.py
```diff
@@ -36,10 +36,12 @@ def test_get_worksheet(self):
...
     def test_get_worksheet_by_id(self):
-        sheet1 = self.spreadsheet.get_worksheet_by_id(0)
-        self.assertTrue(isinstance(sheet1, gspread.Worksheet))
+        sheet1_by_int = self.spreadsheet.get_worksheet_by_id(0)
+        sheet1_by_str = self.spreadsheet.get_worksheet_by_id(0)
+        self.assertTrue(isinstance(sheet1_by_int, gspread.Worksheet))
+        self.assertTrue(isinstance(sheet1_by_str, gspread.Worksheet))
```
alifeee, line 44: `sheet1_by_str` does not use a string? It should be `"0"`?

alifeee, line 44: it may also be nice to set up a test for del_worksheet_by_id, to see that it works with both `str` and `int` id. This should be a quick addition.

alifeee (changes requested):
thanks for the changes :)

it is great that you write tests

gspread/spreadsheet.py
FlantasticDan: The `sheet_data` object in this function stores `sheet_data["properties"]["sheetId"]` as an int, passing a string into the lambda fails every time.  Seems like the upstream `fetch_sheet_metadata()` call returns the G Sheet API response unmodified and [according to the Google Documentation](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/sheets#SheetProperties), the `sheetId` value will always be an integer so I think string -> int is the way to go.

alifeee: you're right, I wasn't sure of Google's stance. in that case what you had is perfect! and perhaps we should change delete to send the data as an int.

good investigating!

gspread/spreadsheet.py
FlantasticDan: Because of [above](https://github.com/burnash/gspread/pull/1269#discussion_r1288756690) I believe the `ValueError` is still necessary, the question then is leave it as one exception handler or create a separate one?  Maybe something like this could help point someone in the right direction:
```python
try:
  # ...
except ValueError:
  raise WorksheetNotFound("id {} not found, id must be of type int".format(id))
```

tests/spreadsheet_test.py
FlantasticDan: oops good catch 😅

gspread/spreadsheet.py
FlantasticDan: Sound's good, the [API declares](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/request#DeleteSheetRequest) that a `DeleteSheetRequest` should also provide `sheetId` as an int so perhaps it's just been luck it's been working all this time.

alifeee: I think so, yes. Let's switch to int

tests/spreadsheet_test.py
alifeee: I personally prefer tests to be seperate, but that's fine too :)

alifeee:
This looks good. You have a merge conflict with your the cassettes. I would perhaps just try deleting the test cassette (json file) and re-running the tests, rather than trying to fix it manually.

After that's fixed, I will have another look :)

FlantasticDan:
So I deleted the .json's, re-ran tests, and pushed the commit with the new ones and now it looks like there's still merge commits but they're subtly different...  It seems to me like the conflicts are just the headers coming in a different order in the file than the existing versions which maybe doesn't matter?

alifeee:
the merge conflict was my fault haha. I merged a seperate PR earlier today, but after you opened this.

I have fixed it for you. It all looks good now!

FlantasticDan:
good timing lol, I was just about to go through and do some copy paste hijinks to resolve it.

alifeee (approved):

alifeee:
thanks so much for this PR. It should be implemented in the next version (v10) in a few weeks.

I consider it good, so once @lavigne958 reviews this we will merge it.

alifeee (approved):

FlantasticDan:
Awesome, thanks for your help @alifeee

lavigne958 (approved):
thank you for this contribution :+1:

Issue #1213, opened 2023-06-07
REFACTOR: Change how colors are handled internally (& externally) with update_tab_color


alifeee:
There are many problems with how colors are currently set/read. See #1199.

The main issue is that if you send a color of `0.5` to Google, the color is set to `0.49803922` (closest 8-bit value).

This issue proposes that we change how colors are implemented to align to this 8-bit-ness.

## old

```python
color = {
  "red": 0,
  "green": 0.5,
  "blue": 1
}
```

## new

unsure which of these is best

```python
color = "#FF8000"
color = [255, 128, 0]
color = {
  "red": 255,
  "green": 128,
  "blue": 0
}
```

More discussion in #1199 

> Thinking about it:
>
>  - the API uses float values
>  - but colors are always set using integer between [0:255]
>  - why do we have to ask the user for a float and then change it 🧐
>
> Can't we use whol numbers from 0 to 255 do the computation for the API and then return the appropriate value for the user ?
>
> You see what I mean?
>
> Btw: that's a breaking change 😛

> 
> 1. a new method to set the color of a tab, something like: set_tab_color(hex_color) that takes only a color as hex value
> 1. a new function in utils.py that can convert a color from 3 arguments to a hex value: convert_colors_to_hex_value(red, gree, blue) this > way a user can use it 2 in ways seamlessly:
>     1. using direct argument assignment: ..._hex_value(123,43,12)
>     1. using named arguments: ..._hex_value(red=123, green=43, blue=12)

lavigne958:
closed by #1270

PR #1270, opened 2023-08-12
Implement hex color conversion

idonec:
Added utility functions for conversion from hex color value to API-accepted color dicts and converting 0-255 RGB colors to hex color values.

Added the `set_tab_color` worksheet function to update tabs colors using hex values.

Included relevant documentation, tests, doctests, and cassette.

As mentioned in #1199 and #1213 
I'm interested in the maintainers' perspective on non-breaking ways of incorporating this change
into other related sheet formatting functionality.

Thank you for your work on this great library.


alifeee (changes requested):
thank you for this PR! :) your code is nice and your tests nicer

I have submitted a few comments, and also have these remarks:

A question we must ask is if we want to make this a breaking change (force everyone to use hex from v6.0.0) or an addition (new methods/updated arguments for using hex)

In this context, we can either

1. case 1 - breaking change

    1. change `Worksheet.update_tab_color` to use **only** hex colors (we must provide a deprecation warning and guidance for which function to use to convert dict -> hex to ease migration).
    1. change `Worksheet.tab_color` to return hex instead of dict 

1. case 2 - addition

    1. change `update_tab_color` to *also* accept hex colors (@lavigne958 I do not know how complex you wish the typing to be), and we can write a code fork for `isinstance(dict)` and `isinstance(hex)` (convert to dict) before sending the API
    1. leave `Worksheet.tab_color` as is and inform the user (somewhere?) that there are functions in `utils` for converting this to a hex*

*one could also add an enum when creating the Worksheet object which specifies color type. This sounds too complex.

I prefer option 1 as I believe it is simpler in the long run. Using hex codes for colors is very standard, and it would be nice if that were the case everywhere, without users having to think about dictionaries. A major version release is also a good time to introduce such a breaking change like this.

We would love to know your thoughts, as a contributor! and also @lavigne958's thoughts.

thanks again for this great PR 💗

alifeee:
Additional note:

The "hex -> dict" function represents colors as 0.0-1.0. The "dict -> hex" represents them with 0-255. It seems to me like they should work the same to reduce confusion.

idonec:
I'm also leaning towards the first option. While it would introduce a breaking change, handling colors as hex values seems to be more intuitive and user-friendly.
Unless we account for the ability to do some sick color arithmetics I don't think this would result in significant losses.
Definitely would need a warning though.

gspread/utils.py
```diff
@@ -8,7 +8,7 @@
 
 import re
 from collections import defaultdict
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
```
alifeee, line 11: you should import `Mapping` from `typing` instead. This is why the Python 3.8 test fails.

tests/utils_test.py
```diff
@@ -295,3 +295,18 @@ def test_convert_hex_to_color(self):
         # raise ValueError on invalid hex characters
         with self.assertRaises(ValueError):
             utils.convert_hex_to_colors_dict("axbcde")
+
+    def test_fill_gaps(self):
```
alifeee, line 299: this has snuck in from elsewhere. it is not relevant to this PR

alifeee (changes requested):

alifeee:
> I'm also leaning towards the first option. While it would introduce a breaking change, handling colors as hex values seems to be more intuitive and user-friendly.

Agreed. For now, you can continue with this PR, and we will deal with any Deprecation warnings that need adding :)

tests/utils_test.py
```diff
@@ -260,6 +260,42 @@ def test_combine_merge_values(self):
...
+        hex = utils.convert_colors_to_hex_value(green=1)
+        self.assertEqual(hex, "#00FF00")
+
+        # throw ValueError on color values out of range (0-1)
+        with self.assertRaises(ValueError):
+            utils.convert_colors_to_hex_value(1.23, 0, -50)
+
+    def test_convert_hex_to_color(self):
+        hex = "#FF8000"
+        expected_color = {"red": 1, "green": 128 / 255, "blue": 0}
```
lavigne958, line 281: for the green value please use some random value with some decimals, just in case and not some rounded value if possible, something like: `0.49803922` which is the return value for the green field of one of my sheets.

lavigne958 (commented):
thank you for this contribution this is very nice ! good work

I agree with the breaking change, this is now or in a far future.... no uses the dict decimal value like the API uses :-( 

@alifeee  do you agree we must first merge the PR that introduces the warning for breaking change then we merge this PR ? just in case :thinking:

alifeee:
> @alifeee do you agree we must first merge the PR that introduces the warning for breaking change then we merge this PR ? just in case 🤔

I think we are fine to merge this PR first. Deprecation warnings are tracked by the issue #1272.

lavigne958 (approved):

alifeee:
personally I would stay away from rounding, as

as mentioned by https://github.com/burnash/gspread/pull/1199#issuecomment-1579546594, Google inconsistently rounds the returned value. Thus, you would desire a test for each of the 256 possible values (perhaps with [Pytest.Parametrize](https://docs.pytest.org/en/stable/how-to/parametrize.html)) to ensure that equality can be maintained.

However, this would be quite frivolous for a number the user never gets. However, it might be worth testing that equality is conserved in hex-space, ie, that setting "#343434" is not incorrectly rounded to "#353535", for example (I am not sure of the logic, whether it uses floor, ceil, or round).

lavigne958:
> Having looked, it looks like rounding errors are fine as `def convert_colors_to_hex_value` rounds decimals to the closest fraction of `255`.
> 
> I still think we should not use `round` in `convert_hex_to_colors_dict` for the reasons above.
> 
> Really, the tests should stay in hex only, as that is the only color format that we support (after this merge), so the only one we should test.

I understand that's what I thought too when I saw that. What do you want to test: only the conversation given a hex value I always get the same dict values ?
Or the other day around? 🤔

alifeee (changes requested):

alifeee (approved):
this is lovely now. thank you very much for your work :))

I am very happy with the code here. I will let @lavigne958 merge this.

lavigne958 (approved):

PR #1271, opened 2023-08-13
Add required kwargs with no default value

### LINKED ISSUE #1264: Feature request: Support named parameters to permit easier transition to 6.x API
Since around v5.10, we have started seeing `DeprecationWarning` notices. For example, a call such as `worksheet.update('A1', [['foo']])` results in the warning:

> `method signature will change to: 'Worksheet.update(value = [[]], range_name=) arguments 'range_name' and 'values' will swap, values will be mandatory of type: 'list(list(...))`

We want to update our 5.x code to support the new 6.x API when it arrives. However it seems that it is currently _not possible_ to use named parameters for some functions. If we specify named parameters like `worksheet.update(range_name='A1', values=[['foo']])` then we get an error:

> `update got unexpected keyword arguments: ['range_name', 'values']`

It appears that this is because of the `@accepted_kwargs` decorator on many functions that implements defaults. As far as I understand it, these only define *optional* parameters (and their defaults), and not *required* parameters. So if we use named required parameters, they won't pass the `@accepted_kwargs` decorator check.

The goal is to allow us to support named parameters in a 5.x release, so that we can easily transition to the 6.x API. E.g. we should be able to use `worksheet.update(range_name='A1', values=[['foo']])` in a 5.x release

To implement this, my suggestion is to add the required params to the `@accepted_kwargs` decorator with a sentinel value, e.g.

```python
    @accepted_kwargs(
        range_name=REQUIRED,
        values=None,
        raw=True,
        major_dimension=None,
        ...)
     def update(self, range_name, **kwargs):
```

And update the `@accepted_kwargs` decorator to handle defaults a little differently:

```python
              for k, v in default_kwargs.items():
                   if v != REQUIRED:  # Ignore setting defaults for REQUIRED kwargs.
                      kwargs.setdefault(k, v)
```

lavigne958:
In the next major release we will update some method signature. In order to help users in the transition, add this new special value which allows a user to specify an argument as a regular arg and as a kwarg.

In the case of the breaking change of the method `Worksheet.update()` it will allow users to speficy the argument names and keep their code running when the update arrives with no changes.

closes #1264

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -89,6 +89,8 @@ DEPRECATION_WARNING_TEMPLATE = (
     "[Deprecated][in version {v_deprecated}]: {msg_deprecated}"
 )
 
+REQUIRED_KWARGS = "required"
+
 
 def convert_credentials(credentials):
     module = credentials.__module__
@@ -723,7 +725,8 @@ def accepted_kwargs(**default_kwargs):
                 raise TypeError(err % (f.__name__, list(unexpected_kwargs)))
 
             for k, v in default_kwargs.items():
-                kwargs.setdefault(k, v)
+                if v != REQUIRED_KWARGS:
+                    kwargs.setdefault(k, v)
 
             return f(*args, **kwargs)
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -13,6 +13,7 @@ from .exceptions import GSpreadException
 from .urls import SPREADSHEET_URL, WORKSHEET_DRIVE_URL
 from .utils import (
     DEPRECATION_WARNING_TEMPLATE,
+    REQUIRED_KWARGS,
     Dimension,
     PasteOrientation,
     PasteType,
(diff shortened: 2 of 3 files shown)


alifeee:
code looks good

I am not fully understanding, so... it would be nice to have a "before" and "after" example of one or two of the functions which are changing (which we can also use as part of the 6.0.0 release notes, i.e., a mini "migration guide")

alifeee (approved):

lavigne958:
> code looks good
> 
> I am not fully understanding, so... it would be nice to have a "before" and "after" example of one or two of the functions which are changing (which we can also use as part of the 6.0.0 release notes, i.e., a mini "migration guide")

I see, you think I should update the commit message or post some details here ?

basicaly:
## Currently
- any argument with no default valut is mandatory
- any argument with a default value is mandatory
- if a function accepts a special argument `**kwargs` and the function has the decorator `accepted_kwargs` then when being used all kwargs must be in in this list

## Now
- any argument given to `accepted_kwargs` with the special value can be given or not to the function, can have a value or not it's free.
- this is so users can use this argument like before  without naming it (ex: `update("B1")`) or using named argument and so the argument goes in the `**kwargs` list and it woks (ex: `update(range="B1")`)

alifeee:
Neat, so it's something like

### old (v5) with deprecation warning

`update("A4:E4", [4, 5, 6, 7, 8])`

### old (v5) without deprecation warning

`update(range_name="A4:E4", values=[4, 5, 6, 7, 8])`

### new (v6)

`update([[4, 5, 6, 7, 8]], "A4:E4")`

I think something like this would be good to put in some release notes? Not sure.

alifeee:
That's lovely. With this and other things to do with v6.0.0, is it worth collecting these somewhere for a mini migration guide?

lavigne958:
> 

we could update the main readme with some kind of "important" or "V6.0.0 migration guide" that's a good idea. I'll create the issue to keep track of it

Issue #1208, opened 2023-06-06
Suggestions for Wrapper and Convenient Methods (add protect/unprotect all)


PFython:
**Is your feature request related to a problem? Please describe.**
First of all, thanks for all the effort you've put into this package already - it's really helped me on my latest project.

I created the following general purpose 'wrapper' for quickly creating a Worksheet/Tab based on key, title, sheet name, sheet index, or account path.  It also includes convenience methods for:

- `.view()` : Opening a worksheet in your browser directly
- `.protect()` : Protecting a whole sheet (except for the service account running the code)
- `.unprotect()` : Unprotecting a whole sheet
- `.df` : Converting directly to a Pandas DataFrame
- `pandas.DataFrame.to_gspread()` : Saving a Pandas DataFrame (created using `.df`) back to its associated worksheet

I've also added convenience attributes to the worksheet object:

`.email` : the service account email (needs to be 'invited' to share a Google Sheet before having access)
`.account_path` : path to the service account JSON file

It wasn't immediately obvious to me where to add these in your structure otherwise I'd have created a pull request but the basic code is below if you think this would be a helpful utility to include?

**Describe the solution you'd like**
```
from pathlib import Path
import webbrowser

import gspread

def _view(self: gspread.worksheet.Worksheet):
    """Open this Google Sheet in the default web browser"""
    webbrowser.open(self.url)

def _protect(self):
    body = {
        "requests": [{
            "addProtectedRange": {"protectedRange": {
                "range": {"sheetId": self._properties['sheetId'],},
                "warningOnly": False,
                "description": f"LOCKED by: {self.email}",
                "editors": {
                        "domainUsersCanEdit": False,
                        "users": [self.email]
        },}}}]}
    try:
        return self.spreadsheet.batch_update(body)
    except gspread.exceptions.APIError as api:
        print(f"'{self._properties['title']}' is already protected")
        return api

def _unprotect(self):
    range_id = self.spreadsheet.list_protected_ranges(self.id).pop()
    range_id = range_id['protectedRangeId']
    body = {
        "requests":
            [{"deleteProtectedRange": {"protectedRangeId": range_id}}]
    }
    try:
        return self.spreadsheet.batch_update(body)
    except gspread.exceptions.APIError as api:
        print(f"'{self._properties['title']}' is already protected")
        return api

def _df(self):
    """
    Return a Pandas Dataframe from the specified Google Sheet object.

    Add attribute:
[...]

PR #1273, opened 2023-08-15
Protect/unprotect

PFython:
Relates to Issue #1208

- Added `Worksheet.protect` and `Worksheet.unprotect`.  I've suggested keeping the names 'protect' and 'unprotect' instead of 'protect_all' and 'unprotect_all' because the scope is already clear ie. the whole Worksheet, and this avoids extra typing.  The lack of arguments should also make it abundantly clear this doesn't protect/unprotect specific ranges.
- Still need a way of fetching `client.auth._service_account_email` to send to `Worksheet.add_protected_range()`.  This attribute seems to only be created on initialisation and I wasn't able to work out how to fetch it eg. from `auth`.  One approach might be to add `.email` attribute to `Spreadsheet` when object is first initialised?
- Added `domain_users_can_edit=False` as parameter to `Worksheet.add_protected_range()`.  Without it, a range is created but other (non system) users can still edit ie. not "View only" mode.
- Added `Worksheet.list_protected_ranges` as a helper method to match `Spreadsheet.list_protected_ranges` and avoid people having to know or guess code for the "round trip" ie. up to the parent spreadsheet, then back down to the original worksheet using id.
- Added `Worksheet.column_count` as alias for `.col_count`.  A pet peeve for many non-English speakers are abbreviations where an explicit (full English) variable name would avoid them having to guess.
- Not sure what tests to add - my test was visual inspection of the sample Worksheet: Is there a padlock icon on the specified sheet? Is the protected range shown in the right hand panel (Data - Protect sheets and ranges)?  This was what alerted me to things not working without 1domain_users_can_edit=False` (above).
- Updated README

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1910,6 +1910,7 @@ class Worksheet:
         description=None,
         warning_only=False,
         requesting_user_can_edit=False,
+        domain_users_can_edit=False,
     ):
         """Add protected range to the sheet. Only the editors can edit
         the protected range.
@@ -1962,6 +1963,7 @@ class Worksheet:
                             "editors": {
                                 "users": editor_users_emails,
                                 "groups": editor_groups_emails,
+                                "domainUsersCanEdit": domain_users_can_edit,
                             },
                         }
                     }
@@ -2964,3 +2966,22 @@ class Worksheet:
         }
 
         return self.spreadsheet.batch_update(body)
+
+    def column_count(self):
+        """Full English alias for .col_count"""
+        return self.col_count()
+
+    def list_protected_ranges(self):
+        """List protected ranges in current Worksheet"""
+        return self.spreadsheet.list_protected_ranges(self.id)
+
+    def protect(self):
+        """Protect all ranges in current Worksheet"""
+        email = get_email_from_somewhere_tbd()  # TODO ?
+        last_cell = rowcol_to_a1(self.row_count, self.col_count)
+        self.add_protected_range(f"A1:{last_cell}",email,description=f"LOCKED by {email}",)
+
+    def unprotect(self):
+        """Unprotect all ranges in current Worksheet"""
+        for range in self.list_protected_ranges(self.id):
+            self.delete_protected_range(range)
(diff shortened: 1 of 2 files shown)


gspread/worksheet.py
```diff
@@ -1910,6 +1910,7 @@ def add_protected_range(
         description=None,
         warning_only=False,
         requesting_user_can_edit=False,
+        domain_users_can_edit=False,
```
alifeee, line 1913: this should be added to the Sphinx docs below (in function docstring)

gspread/worksheet.py
```diff
@@ -2964,3 +2966,26 @@ def cut_range(
         }
 
         return self.spreadsheet.batch_update(body)
+
+    def column_count(self):
```
alifeee, line 2970: I separated this out into #1274. It can be removed from here

alifeee (changes requested):
thanks for this PR! good stuff :)

We will have a think about how to get the email. I'm not sure how all the auth works, but... is there *always* an email? For example if you are using a different type of [authentication](https://docs.gspread.org/en/latest/oauth2.html).

Further to-dos:

- [ ] add tests
  - [ ] for protecting entire worksheet
  - [ ] for unprotecting entire worksheet
  - [ ] for protection kwarg domain_users_can_edit
  - [ ] for `worksheet.list_protected_ranges()`

gspread/worksheet.py
```diff
@@ -2964,3 +2966,26 @@ def cut_range(
...
+
+    def protect(self):
+        """Protect all ranges in current Worksheet"""
+        email = get_email_from_somewhere_tbd()  # TODO ?
+        last_cell = rowcol_to_a1(self.row_count, self.col_count)
+        self.add_protected_range(
+            f"A1:{last_cell}",
+            email,
+            description="LOCKED by gspread user",
+        )
```
lavigne958, line 2986: Did we try to call the API to add a new protected range but with no actual range. Something like, instead of giving: `sheet1!A1:Z99` what if we give it: `sheet1` as "range" does the API understand that we want to cover everything and then protects everything?? 🙃

lavigne958:
I removed the milestone from this PR, we won't finish it on time for the next release.

PR #1275, opened 2023-08-16
Fix permission test

### LINKED ISSUE #1268: Fix BackOffHTTPClient not returning error 403
Currently, we cannot reliably test raising of `PermissionError` (see #1244 #1245).

This is because for the tests, a [`BackOffHTTPClient`] is used, which has problems returning 403 errors.

https://github.com/burnash/gspread/blob/6259c2be9fae8e7567ea1796dd98e3c0d2570672/gspread/http_client.py#L329-L335

https://github.com/burnash/gspread/blob/6259c2be9fae8e7567ea1796dd98e3c0d2570672/tests/client_test.py#L110-L127

[`BackOffHTTPClient`]: https://github.com/burnash/gspread/blob/6259c2be9fae8e7567ea1796dd98e3c0d2570672/gspread/http_client.py#L310

If this is fixed, the test can be un-skipped.

With the normal HTTP client, it works fine (returns `PermissionError` immediately)

alifeee:
closes #1268

currently not working. not sure if I've made it right.

Tests work online, but not offline.

Would appreciate help.


alifeee:
wrong base. broken pr. ignore

PR #1281, opened 2023-08-24
Add maintain_size argument to get_value

botcs:
New feature for #947

The google API strips: 
- the trailing empty cells for each row in the returned range
- the trailing empty rows

This new feature allows the user to maintain the size of the range by adding cells containing an empty string `""`.

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -6,6 +6,7 @@ This module contains common worksheets' models.
 
 """
 
+import re
 import warnings
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


gspread/worksheet.py
```diff
@@ -361,13 +362,49 @@ def range(self, name=""):
...
+            end_col_int = 0
+            for i, col in enumerate(end_col[::-1]):
+                end_col_int += (ord(col) - 64) * (26 ** i)
+
+            num_cols = end_col_int - start_col_int + 1
+            
+        # calculate the number of rows
+        num_rows = int(end_row) - int(start_row) + 1
+
+        return num_rows, num_cols
```
lavigne958, line 398: we already have a function that does that computation :wink: 

see https://github.com/burnash/gspread/blob/4b4239fcd09a5312d5e5e9c32ccbaf86d9a84c7f/gspread/utils.py#L386

gspread/worksheet.py
```diff
@@ -464,7 +503,13 @@ def get_values(self, range_name=None, combine_merged_cells=False, **kwargs):
             worksheet.get_values('A2:B4', value_render_option=ValueRenderOption.formula)
         """
         try:
-            vals = fill_gaps(self.get(range_name, **kwargs))
+            if maintain_size:
```
lavigne958, line 506: please compare it to a the very specific value you expect so here it should be `if maintain_size is True` in order to prevent false positive like if `maintain_size` is a string or anything else that evaluates to `True`.

lavigne958, line 506: should be the other way around right ? if `maintain_size` is `True` we call the get range size, then fill gaps from the computed range size and not just the values right ?

gspread/worksheet.py
alifeee: does the function `a1_range_to_grid_range` work with named ranges? (see https://github.com/burnash/gspread/issues/947#issuecomment-1688552030)

e.g., what happens when the user asks for the range `"A3:B"` (wanting an expanding column) or just `"Sheet1"`. How can we know the size of the data requested?

alifeee:
replaced by #1305

Issue #1290, opened 2023-09-06
API error 403 since 5.11.0 [auth without browser]


DamonDay:
## this is how I auth without browser:
1. generate a token creds locally using browser (_client secret_)
2. hash the token then read it on databricks notebook (sort of Google Colab)<br>
## This method can help me complete Google authentication in a no-browser scenario, because I cannot do browser authentication on the server side
(due to company regulation, I cannot use service account solution.)
## however, since 5.11.0, this way fails (even locally):
`
import gspread
import pickle

secret = dbutils.secrets.getBytes(scope="gar", key="token") # use databricks API get secret token

creds = pickle.loads(secret)

gc = gspread.authorize(creds)

gc.open_by_key("GOOGLE_SHEET_ID")
`

`Traceback (most recent call last):
  File "/opt/miniconda3/envs/py395/lib/python3.9/site-packages/gspread/client.py", line 199, in open_by_key
    spreadsheet = Spreadsheet(self, {"id": key})
  File "/opt/miniconda3/envs/py395/lib/python3.9/site-packages/gspread/spreadsheet.py", line 40, in __init__
    drive_metadata = self.client.get_file_drive_metadata(self._properties["id"])
  File "/opt/miniconda3/envs/py395/lib/python3.9/site-packages/gspread/client.py", line 110, in get_file_drive_metadata
    res = self.request("get", url, params=params)
  File "/opt/miniconda3/envs/py395/lib/python3.9/site-packages/gspread/client.py", line 94, in request
    raise APIError(response)
gspread.exceptions.APIError: {'code': 403, 'message': 'Request had insufficient authentication scopes.', 'errors': [{'message': 'Insufficient Permission', 'domain': 'global', 'reason': 'insufficientPermissions'}], 'status': 'PERMISSION_DENIED', 'details': [{'@type': 'type.googleapis.com/google.rpc.ErrorInfo', 'reason': 'ACCESS_TOKEN_SCOPE_INSUFFICIENT', 'domain': 'googleapis.com', 'metadata': {'service': 'drive.googleapis.com', 'method': 'google.apps.drive.v3.DriveFiles.Get'}}]}`

AugustoBarros:
I also had the same issue, even using a service account.

"Downgrading" the package to 5.10.0 resolved my issue.

alifeee:
Hi both. This is now fixed in the patch v5.11.1. You should be able to update to v5.11.1 and your code should work.

[...]

Issue #1286, opened 2023-09-04
Google Drive API Required after 5.11.0 #1277


lathonez:
After 5.11.0 (and I assume #1277), I needed to enable the Google Drive API on my project in order to open a google sheet (using any mechanism).

The error was very helpful:

```
gspread.exceptions.APIError: {'code': 403, 'message': 'Google Drive API has not been used in project 7575758011 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/drive.googleapis.com/overview?project=7575758011 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.', 'errors': [{'message': 'Google Drive API has not been used in project 7575758011 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/drive.googleapis.com/overview?project=7575758011 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.', 'domain': 'usageLimits', 'reason': 'accessNotConfigured', 'extendedHelp': 'https://console.developers.google.com'}], 'status': 'PERMISSION_DENIED', 'details': [{'@type': 'type.googleapis.com/google.rpc.Help', 'links': [{'description': 'Google developers console API activation', 'url': 'https://console.developers.google.com/apis/api/drive.googleapis.com/overview?project=7575758011'}]}, {'@type': 'type.googleapis.com/google.rpc.ErrorInfo', 'reason': 'SERVICE_DISABLED', 'domain': 'googleapis.com', 'metadata': {'consumer': 'projects/7575758011', 'service': 'drive.googleapis.com'}}]}
```

But it threw me. I was making a fully unrelated code change and I couldn't figure out why I'd suddenly need to enable Google Drive API. Took me a while to track it back to this PR.

From the context, it looks like #1277 is fixing a bug, and I probably should have had Google Drive API enabled the whole time. 

I'm just raising this for others / in case you think it would be helpful to document the change.

alifeee:
Hi. Thanks for this issue :) - it is something I clearly did not anticipate when implementing #1277.

Technically, [our documentation] says you should enable both Drive API and Sheets API. However, I can see how many people (probably myself included) would skim over it and miss out the step of enabling Drive API.

[our documentation]: https://docs.gspread.org/en/v5.10.0/oauth2.html#enable-api-access-for-a-project

[...]

PR #1291, opened 2023-09-06
remove Drive API access on Spreadsheet init (FIX - VERSION 5.11.1)

alifeee:
Adding this line to `Spreadsheet.__init__()`

https://github.com/burnash/gspread/blob/817ebb86fd1b98e2eb643eb2e06ffd5bd63544ca/gspread/spreadsheet.py#L40-L41

has caused many issues (#1286, #1290) due to gspread being used with an API key/access token/etc which does *not* have access to the Drive API.

I propose that we remove this line (effectively reverting #1255) and release a new patch -> v5.11.1.

I think what #1255 fixes (removal of a single API call on two rarely used(?) spreadsheet properties) is not worth the implications that requiring the Drive API by default brings.

I would like to make this change asap but will wait for your approval @lavigne958.

### DIFF
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -37,9 +37,6 @@ class Spreadsheet:
         metadata = self.fetch_sheet_metadata()
         self._properties.update(metadata["properties"])
 
-        drive_metadata = self.client.get_file_drive_metadata(self._properties["id"])
-        self._properties.update(drive_metadata)
-
     @property
     def id(self):
         """Spreadsheet ID."""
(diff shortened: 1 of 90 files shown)


lavigne958 (approved):
I agree this brings too much problems to the users. 

That's a good call.

lavigne958:
You make the new release too ? 
We should probably add some mentions in the release note ? 🤔

alifeee:
I will make the new release.

Not sure what you want in the release notes, but you can edit it after I make it :). A bit like the note I made in the [5.11.0 release notes](https://github.com/burnash/gspread/releases/tag/v5.11.0)?

PR #1296, opened 2023-09-10
Make `get_values` and alias of `get`

### LINKED ISSUE #1285: Why does `get_values` exist?
`gspread.Worksheet` offers two ways of getting a range of cells: [`Worksheet.get`] and [`Worksheet.get_values`].

Why?

...

They return almost the same thing? [`Worksheet.get`] returns a [`ValueRange`], which inherits from `list`. [`Worksheet.get_values`] returns a list. `get_values` pads the lists so that they are square, whereas `get` does not.

```python
v_get_values = worksheet.get_values("A3:C6")
v_get = worksheet.get("A3:C6")

print("v_get_values: ", v_get_values)
print("v_get: ", v_get)
```

Screenshot of spreadsheet

```bash
$ py .\get_vs_get_values.py
v_get_values:  [['2', 'b', '2'], ['3', 'c', ''], ['4', 'd', '']]
v_get:  [['2', 'b', '2'], ['3', 'c'], ['4', 'd']]
```

[`Worksheet.get`]: https://github.com/burnash/gspread/blob/9cc862bc24f61de763944c59b2e1d270ef3b48cf/gspread/worksheet.py#L779-L870
[`Worksheet.get_values`]: https://github.com/burnash/gspread/blob/9cc862bc24f61de763944c59b2e1d270ef3b48cf/gspread/worksheet.py#L364-L477
[`ValueRange`]: https://github.com/burnash/gspread/blob/9cc862bc24f61de763944c59b2e1d270ef3b48cf/gspread/worksheet.py#L41

Why is one required over another? I propose that `get` should also merge the values, and then `get_values` is not required (can just call `get`)

alifeee:
closes #1285, #1303

Having both of these method is ambiguous. I have made it so only [`get`] is the real method, and [`get_values`] is its alias.

```python
def get_values(self, *args, **kwargs) -> ValueRange:
        """Alias for :meth:`~gspread.worksheet.Worksheet.get`"""
        return self.get(*args, **kwargs)
```

Potential problems:

- [`get_values`] used to return type `List[List[str]]`.
  [`get`] returns a `ValueRange`
  thus now `get_values` returns a `ValueRange` instead
  `ValueRange` subclasses `list`, so it's okay?

[`get`]: https://github.com/burnash/gspread/blob/7abf1e2533d4a8b94df5fb732b444b146c2d65da/gspread/worksheet.py#L835-L929
[`get_values`]: https://github.com/burnash/gspread/blob/7abf1e2533d4a8b94df5fb732b444b146c2d65da/gspread/worksheet.py#L383-L504


gspread/worksheet.py
```diff
@@ -926,7 +812,21 @@ def get(
...
+
+        if combine_merged_cells is True:
+            spreadsheet_meta = self.client.fetch_sheet_metadata(self.spreadsheet_id)
+            worksheet_meta = finditem(
+                lambda x: x["properties"]["title"] == self.title,
+                spreadsheet_meta["sheets"],
+            )
+            vals = combined_merge_values(worksheet_meta, vals)
+
+        return ValueRange.from_json(response, vals)
```
lavigne958, line 829: it feels odd to get the values then rework the then extract them again using `ValueRange` but this time the response object is almose useless.

I beleive we can simplify it like this:

```python
vals = fill_gaps(response.get("values", [[]]))
```

then bellow force update the response object with the values reformated:

```python
response["values"] = vals
return ValueRange.from_json(response)
```

which removes the changes on the `ValueRange.from_json` method too.

what do you think ? :thinking:

lavigne958 (commented):
I believe we can make it simpler, see comment.

Changing the outcome of the test is good sing of a breaking change :open_mouth: .

We should add a new warning about it.

lavigne958:
> closes #1285
> 
> Having both of these method is ambiguous. I have made it so only [`get`](https://github.com/burnash/gspread/blob/7abf1e2533d4a8b94df5fb732b444b146c2d65da/gspread/worksheet.py#L835-L929) is the real method, and [`get_values`](https://github.com/burnash/gspread/blob/7abf1e2533d4a8b94df5fb732b444b146c2d65da/gspread/worksheet.py#L383-L504) is its alias.
> 
> ```python
> def get_values(self, *args, **kwargs) -> ValueRange:
>         """Alias for :meth:`~gspread.worksheet.Worksheet.get`"""
>         return self.get(*args, **kwargs)
> ```
> 
> Potential problems:
> 
> * [`get_values`](https://github.com/burnash/gspread/blob/7abf1e2533d4a8b94df5fb732b444b146c2d65da/gspread/worksheet.py#L383-L504) used to return type `List[List[str]]`.
>   [`get`](https://github.com/burnash/gspread/blob/7abf1e2533d4a8b94df5fb732b444b146c2d65da/gspread/worksheet.py#L835-L929) returns a `ValueRange`
>   thus now `get_values` returns a `ValueRange` instead
>   `ValueRange` subclasses `list`, so it's okay?

I agree about the duplicate method.

about the linting: I think we can solve it using the propose changes in the code.

alifeee:
> Changing the outcome of the test is good sing of a breaking change 😮 .

Yes, good that this is 6.0.0 ;). Do you think the breaking change is too big?

> about the linting: I think we can solve it using the propose changes in the code.

Not sure what you mean by "linting"

alifeee:
Yes, this is a big change. It needs some thinking about.

The alternate options are:

- leave `get` as-is, and make `get_values` be identical (alias) apart from it pads values. The names don't really make sense why this is the case, though.
- try and make sure the API of `get` and `get_values` always remains the same

alifeee:
For maximum compatibility, `get_values` should be an alias of `get`, but return type `List[List[str]]`.

alifeee:
If you see #1285, `get` does not pad values, and `get_values` does.

(copied from #1285)

```bash
$ py .\get_vs_get_values.py
v_get_values:  [['2', 'b', '2'], ['3', 'c', ''], ['4', 'd', '']]
v_get:  [['2', 'b', '2'], ['3', 'c'], ['4', 'd']]
```

I propose the breaking change here that:

- `get` *does* start padding values

Then, the only difference between `get` and `get_values` is the return type (`List[List[str]]` vs `ValueRange`)

lavigne958:
> I propose the breaking change here that:
> 
> * `get` _does_ start padding values
> 
I agree with this change, but the default behavior must stay so:
 * get can pad the values
* when `get` is called by default it does not pad the values
* when `get_values` calls `get` it requests it to pad the` values.

 this way we keep the same behavior as before but we allow user to padd at both levels. Then we'll be able to deprecate the old behavior in the next+2 major release if we want to

> Then, the only difference between `get` and `get_values` is the return type (`List[List[str]]` vs `ValueRange`)

can we add some flags again to offer the choice of the returned value ? default: keep the old return type, flag is set: return ValueRange ?

*I'm sorry I edited your comment instead of replying ! :facepalm: *

lavigne958 (approved):

alifeee:
First of all, sorry about all the emails for a failing workflow... haha

I have changed this implementation. Now:

- The previous tests are not changed (just reformatted) - no longer a breaking change (with previous tests)
- Two tests have been added to verify the behaviour of `get` (no padding, ValueRange) and `get_values` (padding, List[List[Any]])
- `get` has two more kwargs: `pad_values` and `return_type`. I hope you can guess what these do.

Now, we can (in theory) merge this and it will not be a change to the current behaviour. In future, we can decide to change the default values of some kwargs (like `maintain_size` - #1306, `pad_values`, `return_type`)

lavigne958:
> First of all, sorry about all the emails for a failing workflow... haha
> 
> I have changed this implementation. Now:
> 
> * The previous tests are not changed (just reformatted) - no longer a breaking change (with previous tests)
> * Two tests have been added to verify the behaviour of `get` (no padding, ValueRange) and `get_values` (padding, List[List[Any]])
> * `get` has two more kwargs: `pad_values` and `return_type`. I hope you can guess what these do.
> 
> Now, we can (in theory) merge this and it will not be a change to the current behaviour. In future, we can decide to change the default values of some kwargs (like `maintain_size` - #1306, `pad_values`, `return_type`)

great ! good job.

I'll run some regression manual tests then I'll merge it if it works fine.

alifeee:
> I'll run some regression manual tests then I'll merge it if it works fine.

Here is a snippet I've been using

```python
import gspread

gc = gspread.service_account(filename="./creds.json")
spreadsheet = gc.open("gspread test spreadsheet")

worksheet = spreadsheet.sheet1

v_get_values = worksheet.get_values("A3:C6")
v_get = worksheet.get("A3:C6")

print("v_get_values: ", v_get_values)
print("v_get: ", v_get)

# types of values
print("type of v_get_values: ", type(v_get_values))
print("type of v_get: ", type(v_get))
```

gspread/worksheet.py
```diff
@@ -927,7 +822,28 @@ def get(
             self.spreadsheet_id, range_name, params=params
         )
 
-        return ValueRange.from_json(response)
+        vals = response.get("values", [[]])
+
+        if pad_values is True:
+            try:
+                vals = fill_gaps(vals)
```
lavigne958, line 829: We need to compute the requested size here then pass it to the `fill_gaps` function in order to extend the matrix `vals` otherwise the matrix `vals` is too small

lavigne958:
> > I'll run some regression manual tests then I'll merge it if it works fine.
> 
> Here is a snippet I've been using
> 
> ```python
> import gspread
> 
> gc = gspread.service_account(filename="./creds.json")
> spreadsheet = gc.open("gspread test spreadsheet")
> 
> worksheet = spreadsheet.sheet1
> 
> v_get_values = worksheet.get_values("A3:C6")
> v_get = worksheet.get("A3:C6")
> 
> print("v_get_values: ", v_get_values)
> print("v_get: ", v_get)
> 
> # types of values
> print("type of v_get_values: ", type(v_get_values))
> print("type of v_get: ", type(v_get))
> ```

That's exactly what I did  :laughing: and everything works fine, I checked-out many branches with the same code (master, feature/release_6_0_0, alias/get_values, ...) it's all good apart the flag that does not work

gspread/worksheet.py
alifeee: The padding behaviour is not currently in gspread. It will not be until https://github.com/burnash/gspread/pull/1305 is merged.

Thus, the code not padding is correct, and the same as it was before.

lavigne958: should we first merge #1305  ? so then the integration here will be simpler

alifeee: We could, yes. We would then have to merge master into v6 branch, since this PR targets v6.

alifeee:
We are waiting for #1305 to be merged into `feature/release_6_0_0` branch. Thus we are waiting for #1320.

alifeee:
This is now ready to merge. Technically, this introduces *no changes* to the public API (👻wooo👻)

I have removed all code from `worksheet.get_values` so it is now simply

https://github.com/burnash/gspread/blob/6627b534ea791e7cca42e2fc1a1fe07067eb6c9a/gspread/worksheet.py#L393-L402

All heavy lifting is now done in the `worksheet.get` function.

You may notice that `worksheet.get` now has two very similar kwargs: `pad_values` and `maintain_size`. To see what the difference is, please see [the readme](https://github.com/burnash/gspread/blob/alias/get_values/README.md#getting-a-range-of-values)

`pad_values` is introduced to differentiate the behaviour of `get` and `get_values` (see top of thread).

`maintain_size` is the new feature added by #1305

They cannot be combined into one kwarg for two reasons:

- this would be a breaking change (we could do it here since this is for 6.0.0)
- it would not be possible for indeteminate size ranges
    this is to say that `pad_values` operates on all arrays, just making them rectangular. `maintain_size` *must* know the size to pad the array to. Thus, you cannot use `maintain_size` for a call to "A:C", etc.

alifeee:
added #1303 into this PR...

tests/utils_test.py
```diff
@@ -375,6 +375,23 @@ def test_fill_gaps_with_value(self):
 
         self.assertEqual(actual, expected)
 
+    def test_fill_gaps_with_non_square_array(self):
+        """test fill_gaps function"""
+        matrix = [
+            [1, 2, 3, 4],
+            [5, 6, 7, 8],
```
lavigne958, line 382: Can you may be shift the second row `[5, 6, 7, 8]` to one on the right so we have a empty celle on the left side at the beginning of the line ?

`[None, 5, 6, 7 ,8]`

lavigne958 (approved):

tests/utils_test.py
alifeee: I can do that, but `fill_gaps` doesn't ever *replace* values, it only ever adds new values to the arrays (either to make an array square, or to extend the rows/columns)

so, `fill_gaps` applied to `[None, 5, 6, 7, 8]` --> `[None, 5, 6, 7, 8]`.

This becomes a question of what `filil_gaps` *should* do in this situation. It is just a `utils` function, so its scope was never fully defined. What do you think?

lavigne958: I see then leave it as is.

Issue #808, opened 2020-08-26
getallrecords() – can there be a "rows" option to get attributes from rows, not columns?


lexipenia:
**Is your feature request related to a problem? Please describe.**
Sometimes I have to work with spreadsheets that are "rotated 90 degrees" – that is, the creators use the rows to list important attributes, and then every item in the DB is listed in a column. The attributes list would have to be Column A for these sheets, not Row 1.

**Describe the solution you'd like**
An optional argument in getallrecords() that will be able to produce dictionaries using the first *column* to find attribute names, not the first row.

**Describe alternatives you've considered**
It's possible to use getallvalues() and iterate over things appropriately or create mappings. But a dictionary like the one getallrecords() produces would be much faster to work with.

alifeee:
This sounds like a good feature to have!

If you wanted to try implementing it we can help you with any questions.

AndrewBasem1:
@alifeee can I take this please after merging PR #1301 (issue #1294)?

alifeee:
> @alifeee can I take this please after merging PR #1301 (issue #1294)?

As soon as #1301 is merged, yes. This looks like a good thing to do afterwards.

PR #1301, opened 2023-09-21
feature -- adding `worksheet.get_records` to get specific row ranges

### LINKED ISSUE #1294: alternative to `worksheet.get_all_records()` for specific ranges or max num of rows
The `worksheet.get_all_records()` is a convenient method to download the sheet data into structured dictionary format, but it stops being convenient when the worksheet I'm trying to download data from has way too many rows, and I only need a small subset (due to bad internet connection, just wanting a maximum number of rows, so on..)

Creation of a new method that could be called `.get_some_records()` or `.get_records_from_range()` that will only download a certain range only, or a maximum number of `n` rows for example.

Getting the data as list of lists and doing the dictionary conversion myself, pretty easy to do, but still would be helpful if it was already present in the gspread module itself directly

AndrewBasem1:
Closes #1294

## Changes

1. added `get_records_subset` to provide similar behavior to `get_all_records` but allows the user to chose a first_row and last_row to get the data from only.
2. changed `get_all_records` to simply call `get_records_subset`
3. added `_validate_rows_ranges_for_get_records_subset` to validate the user inputs for (`head, first_row, last_row`)
4. added `_validate_headers_and_keys_for_get_records_subset` to do the following:
    1. validate `expected_headers` given by the user are unique
    1. validate that the obtained keys contains the `expected_headers`
    2. validate that the obtained keys are unique
5. added `_pad_values_and_keys_for_get_records_subset` to match the size of the keys and rows, to produce dictionaries as expected.
6. added tests for the new functions


gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
+        )
+        if not expected_headers_are_in_obtained_headers:
             raise GSpreadException(
                 "the given 'expected_headers' contains unknown headers: {}".format(
-                    expected - headers
+                    set(expected_headers) - set(keys)
                 )
             )
 
+    def _pad_values_and_keys_for_get_records_subset(self, values, keys, default_blank):
```
alifeee, line 617: we already have a function that does this: utils.fill_gaps

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
+            values = [row + [""] * (-values_wider_than_keys_by) for row in values]
+
+        return values, keys
+
+    def get_records_subset(
+        self,
+        empty2zero=False,
+        head=1,
+        first_row=None,
+        last_row=None,
```
alifeee, line 642: first_row and last_row should be required arguments. Otherwise this is just get_all_records again. Perhaps `head` should also be required

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
+
+        :param list expected_headers: (optional) List of expected headers, they must be unique.
+
+            .. note::
+
+                returned dictionaries will contain all headers even if not included in this list
+
+        """
+        # some sanity checks
+        head, first_row, last_row = self._validate_rows_ranges_for_get_records_subset(
```
alifeee, line 683: these should be done in-place. no need to define extra functions for them

alifeee (changes requested):
Thanks for this PR! You have put a lot of work into it! :)

Some questions from me:

- Why update all the cassettes? You should only need to update the ones for the tests you modify/add
- The new function should be named `get_records`, as this is simpler
- The first_row and last_row (any maybe head?) should be required arguments
- We do not need the methods `_validate_rows_ranges_for_get_records_subset` `_validate_headers_and_keys_for_get_records_subet` `_pad_values_and_keys_for_get_records_subset`. This logic can go in the function `get_records`

Some thoughts:

- instead of "first_row" and "last_row", we could instead ask for a `range` (user would enter "4:9", for example. This would then make it easier to translate the function to be column-based instead of row-based (see #808)

Thanks again for this work :)

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
+    def _validate_rows_ranges_for_get_records_subset(
+        self,
+        head,
+        first_row,
+        last_row,
+    ):
+        """Validates the given head, first_row and last_row for `get_records_subset`"""
+        if first_row is None:
+            first_row = head + 1
+        elif not (isinstance(first_row, int) and first_row >= head):
```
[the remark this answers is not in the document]
AndrewBasem1: Done

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
+        return head, first_row, last_row
 
-        # if no given expected headers, expect all of them
+    def _validate_headers_and_keys_for_get_records_subset(self, keys, expected_headers):
+        """Validates the returned keys and the given expected headers for `get_records_subset`"""
         if expected_headers is None:
             expected_headers = keys
+        else:
+            # validating the given expected headers
+            if not isinstance(expected_headers, list):
```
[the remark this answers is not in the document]
AndrewBasem1: Done

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
-        # Return an empty list if the sheet doesn't have enough rows
-        if len(data) <= idx:
-            return []
+        if last_row is None:
+            last_row = self.row_count
+        elif not (isinstance(last_row, int) and last_row >= first_row):
+            raise ValueError(
+                "last_row must be an integer greater than or equal to first_row"
+            )
+        elif last_row > self.row_count:
```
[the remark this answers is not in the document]
AndrewBasem1: I think if the user **explicitly** choose a `last_row` means that he/she wants **all the rows** starting from the `first_row` up to and including that `last_row`.

There are two scenarios:
1. the user chooses a `last_row` that's larger than the last non-empty row.
2. the user chooses a `last_row` that's larger than the sheet's `row_count`

I feel like we should give the user a warning message that we will be truncating the output to the last non-empty row within the sheet's `row_count`.

Let me know what you think about that.

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
-        if len(expected) != len(expected_headers):
-            raise GSpreadException("the given 'expected_headers' are not uniques")
+        # validating the headers in the worksheet
+        header_row_is_unique = len(keys) == len(set(keys))
+        if not header_row_is_unique:
+            raise GSpreadException("the header row in the worksheet is not unique")
 
-        if not expected & headers == expected:
+        # validating that the expected headers are part of the headers in the worksheet
+        expected_headers_are_in_obtained_headers = all(
```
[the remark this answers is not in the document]
AndrewBasem1: Done

gspread/worksheet.py
AndrewBasem1: I agree, Done.

gspread/worksheet.py
AndrewBasem1: I checked it actually, but they're not the same. my current function has some important logic and checks in it.

1. `utils.fill_gaps()`:
    1. takes a list of lists, and modifies it to a specific "row and column" dimensions.
    2. the `keys` list we have is not a list of lists.
    3. this will not raise any issues for duplicate headers.

2. `_pad_values_and_keys_for_get_records_subset`:
    1. takes a keys list.
    2. takes a values list of lists.
    3. if the keys are longer than a single list on the values list of lists:
        1. pads the values to have the same length as the keys
    4. if the keys are shorter than a single list on the values list of lists:
        1. if the keys already contains a `default_blank` value, or if it's shorter than the values by more than one, it raises an error
        2. otherwise, just adds a single `default_blank` value.

gspread/worksheet.py
AndrewBasem1: resolving this comment as https://github.com/burnash/gspread/pull/1301#discussion_r1334639018 is similar, will keep the discussion there.

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
...
+            first_row=head + 1,
+            default_blank=default_blank,
+            allow_underscores_in_numeric_literals=allow_underscores_in_numeric_literals,
+            numericise_ignore=numericise_ignore,
+            value_render_option=value_render_option,
+            expected_headers=expected_headers,
+        )
 
-        data = self.get_all_values(value_render_option=value_render_option)
+    def _validate_rows_ranges_for_get_records_subset(
```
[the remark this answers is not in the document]
AndrewBasem1: Yes, I had all of the following methods within the `get_records()` function itself at first:

1. `_validate_rows_ranges_for_get_records_subset`
2. `_validate_headers_and_keys_for_get_records_subset`
3. `_pad_values_and_keys_for_get_records_subset`

But Flake8 was really unhappy with its McCabe complexity, so I split them up.

Here's their current complexities:
```bash
>> python -m mccabe worksheet.py
...
556:4: 'Worksheet._validate_rows_ranges_for_get_records_subset' 7
583:4: 'Worksheet._validate_headers_and_keys_for_get_records_subset' 5
607:4: 'Worksheet._pad_values_and_keys_for_get_records_subset' 4
629:4: 'Worksheet.get_records_subset' 2
...
```

gspread/worksheet.py
```diff
@@ -542,40 +542,165 @@ def get_all_records(
                 returned dictionaries will contain all headers even if not included in this list
 
         """
-        idx = head - 1
+        return self.get_records_subset(
```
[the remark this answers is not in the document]
AndrewBasem1: Done

AndrewBasem1:
I've made a few changes, please check and let me know what you think about the current state.

Also, I'm not sure how replacing first and last_row with a range would help implement #808. I'd actually argue for renaming `first_row` to `first_index` and adding a `use_index` argument (this will be consistent with pandas, which I think a big part of the pythno community already uses)

tests/worksheet_test.py
```diff
@@ -914,6 +916,130 @@ def test_get_all_records_numericise_unformatted(self):
...
+            [4, 5, 6],
+            [7, 8, 9],
+            [10, 11, 12],
+        ]
+        cell_list = self.sheet.range("A1:C5")
+        for cell, value in zip(cell_list, itertools.chain(*rows)):
+            cell.value = value
+        self.sheet.update_cells(cell_list)
+
+        # test1 - set max_row only
```
alifeee, line 934: these comments are out of date? they are helpful, albeit likely unneeded

gspread/worksheet.py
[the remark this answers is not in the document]
alifeee: > But Flake8 was really unhappy with its McCabe complexity, so I split them up.

Understandable. I think it is fine (and preferable) to leave them in the original function. The branching paths are not so branching, and are really just argument validation, so I do not have a problem with the "increased complexity" that flake8 complains about.

gspread/worksheet.py
alifeee: I understand. Via [my other comment](https://github.com/burnash/gspread/pull/1301#discussion_r1335794229), I believe we should move the contents of `_pad_values_and_keys_for_get_records_subset` into the `get_records` function:

This section will remain as-is

https://github.com/AndrewBasem1/gspread/blob/a574f6bfc1bb67aef3260e81785407d4feca2b19/gspread/worksheet.py#L608-L615

and this section can be replaced by `utils.fill_gaps()`

https://github.com/AndrewBasem1/gspread/blob/a574f6bfc1bb67aef3260e81785407d4feca2b19/gspread/worksheet.py#L617-L619

alifeee:
> I've made a few changes, please check and let me know what you think about the current state.

Thanks! Your code changes are very clear, and your tests are layed out very clearly. It is good.

> Also, I'm not sure how replacing first and last_row with a range would help implement https://github.com/burnash/gspread/issues/808. I'd actually argue for renaming first_row to first_index and adding a use_index argument (this will be consistent with pandas, which I think a big part of the pythno community already uses)

This sounds good. My original thinking was that a user could select a range with "4:8" or "D:G", as one could use as indices in `get_values`. Then they would use a kwarg to select between indexing rows or columns first. Your solution is good to me also.

I will enable the workflow now. You may need to run the formatter with `tox -e format`

alifeee:
Cassettes are broken. I have fixed them in https://github.com/burnash/gspread/tree/feature/get_records_limit

I could not push to https://github.com/AndrewBasem1/gspread/tree/feature/get_records_limit

gspread/worksheet.py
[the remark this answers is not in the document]
AndrewBasem1: Sure will move them back inside :+1:

gspread/worksheet.py
AndrewBasem1: Okay will do that :+1:

tests/worksheet_test.py
AndrewBasem1: fixed them, they're up to date now

tests/worksheet_test.py
```diff
@@ -914,6 +916,140 @@ def test_get_all_records_numericise_unformatted(self):
...
+        records_list = [d0, d1]
+        self.assertEqual(read_records, records_list)
+
+        # test4 - set max_row and first_row equal to each other
+        read_records = self.sheet.get_records(first_row=3, last_row=3)
+        d0 = dict(zip(rows[0], rows[2]))
+        records_list = [d0]
+        self.assertEqual(read_records, records_list)
+
+        # test5 - set head only
```
alifeee, line 962: shouldn't this test return all rows? it only looks like it returns rows[2:4] when it should be rows[2:5] (or rows[2:])

alifeee (changes requested):

alifeee:
> [...] I'd actually argue for renaming first_row to first_index and adding a use_index argument [...]

Once this is done, and the CI passes, I think this is ready to merge :)

tests/worksheet_test.py
AndrewBasem1: it does return all rows (head is second row, first row is removed)

alifeee:
You will want to delete the cassette for `test_get_records`, and re-run the test online. Then, it looks like all should be good.

AndrewBasem1:
@alifeee you'll have to excuse my lack of expertise, it's my first time working with cassettes so I don't have a full understanding of them. hope this fixed the issue. Please check and let me know if there is anything else I can do

alifeee:
Thanks again for this PR!! You have been clear, helpful, and thoughtful about your changes. You write Python well.

I am happy to merge this in.

It will be usable in the next release of gspread, which should be v5.12.0 in a couple of weeks ([see milestone](https://github.com/burnash/gspread/milestone/33))

PR #1305, opened 2023-09-28
Add `maintain_size` to keep asked for size in `get`, `get_values`

### LINKED ISSUE #1289: size of range returned by get and get_values can be smaller than the input size
`Worksheet.get` and `Worksheet.get_values` return a smaller area than asked for if some of it is empty. For example, see this call of `worksheet.get("E7:I11")

Screenshot of spreadsheet

## Replication

```python
import gspread
gc = gspread.service_account(filename="google_credentials.json")
spreadsheet = gc.open_by_key("<spreadsheet_key>")
worksheet = spreadsheet.sheet1

values = worksheet.get("E7:I11")
print(values)
```

### expected output (size 5 x 5)
```bash
[
 ['', '', '', '', ''],
 ['', 'data', 'data', 'data', ''],
 ['', '', 'data', '', ''],
 ['', '', 'data', 'data', ''],
 ['', '', '', '', '']
]
```

### actual output (size 4 x 4)
```bash
[
 ['', '', '', ''],
 ['', 'data', 'data', 'data'],
 ['', '', 'data', ''],
 ['', '', 'data', 'data']
]
```

## Related issues

#910, #947, #1284

## Attempted fixes

#985, #1281

## steps to fix

- [x] add a boolean kwarg to [`Worksheet.get`] and [`Worksheet.get_values`], something like `maintain_size`
- [x] if `maintain_size` is true
  - [x] if the size is knowable (see below)
    - [x] get the size (number of rows/cols) asked for by the user from `range_name` using [`a1_to_rowcol`]
    - [x] use [`utils.fill_gaps`] to fill the empty values and return this
- [x] add test, i.e., `WorksheetTest.test_get_values_and_maintain_size`

[`Worksheet.get_values`]: https://github.com/burnash/gspread/blob/739b8fca7e91d798d71a446e8849f42b950c2840/gspread/worksheet.py#L352
[`Worksheet.get`]: https://github.com/burnash/gspread/blob/817ebb86fd1b98e2eb643eb2e06ffd5bd63544ca/gspread/worksheet.py#L784
[`utils.fill_gaps`]: https://github.com/burnash/gspread/blob/739b8fca7e91d798d71a446e8849f42b950c2840/gspread/utils.py#L538-L573
[`a1_to_rowcol`]: https://github.com/burnash/gspread/blob/817ebb86fd1b98e2eb643eb2e06ffd5bd63544ca/gspread/utils.py#L297

## Problems

There are many ways to define a range with Google Sheets. See the [documentation for a1 notation](https://developers.google.com/sheets/api/guides/concepts#a1_notation). Here is a summary. In green are the ranges we can know the size of. In red the ones we can't.

Screenshot of Google Sheets documentation

What we could do is resize the data if the size is knowable (cases 1 and 5 above) and leave it as-is if not (cases 2, 3, 4, 6, 7, 8 above).

## Questions

Should this be the default behaviour?

alifeee:
closes #1289

Users can now specify `maintain_size=True`, so that "A1:C3" will always be a 3x3 array, regardless of whether the cells are full or not. See #1289 more more explanation.

```python
>>> worksheet.get_values("A1:D4")
[
    ["1", "2"],
    ["3", "4"],
    ["5", "6"],
]
```

```python
>>> worksheet.get_values("A1:D4", maintain_size=True)
[
    ["1", "2", "", ""],
    ["3", "4", "", ""],
    ["5", "6", "", ""],
    ["",  "",  "", ""],
]
```

In the future (6.0.0), this could be changed to be default behaviour. Thoughts?

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -24,6 +24,7 @@ from .exceptions import IncorrectCellLabel, InvalidInputValue, NoValidUrlKeyFoun
 MAGIC_NUMBER = 64
 CELL_ADDR_RE = re.compile(r"([A-Za-z]+)([1-9]\d*)")
 A1_ADDR_ROW_COL_RE = re.compile(r"([A-Za-z]+)?([1-9]\d*)?$")
+A1_ADDR_FULL_RE = re.compile(r"[A-Za-z]+\d+:[A-Za-z]+\d+")  # e.g. A1:B2 not A1:B
 
 URL_KEY_V1_RE = re.compile(r"key=([^&#]+)")
 URL_KEY_V2_RE = re.compile(r"/spreadsheets/d/([a-zA-Z0-9-_]+)")
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -583,7 +583,7 @@ def cell_list_to_rect(cell_list):
     if not cell_list:
         return []
 
-    rows = defaultdict(lambda: {})
+    rows = defaultdict(dict)
 
     row_offset = min(c.row for c in cell_list)
     col_offset = min(c.col for c in cell_list)
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -891,6 +891,23 @@ def is_full_a1_notation(range_name: str) -> bool:
         return True
     return False
 
+
+def get_a1_from_absolute_range(range_name: str) -> str:
+    """Get the A1 notation from an absolute range name.
+    "Sheet1!A1:B2" -> "A1:B2"
+    "A1:B2" -> "A1:B2"
+
+    Args:
+        range_name (str): The range name to check.
+
+    Returns:
+        str: The A1 notation of the range name stripped of the sheet.
+    """
+    if "!" in range_name:
+        return range_name.split("!")[1]
+    return range_name
+
+
 if __name__ == "__main__":
     import doctest
 
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -887,9 +887,7 @@ def is_full_a1_notation(range_name: str) -> bool:
         >>> is_full_a1_notation("A1:B")
         False
     """
-    if A1_ADDR_FULL_RE.search(range_name):
-        return True
-    return False
+    return A1_ADDR_FULL_RE.search(range_name) is not None
 
 
 def get_a1_from_absolute_range(range_name: str) -> str:
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -868,6 +868,29 @@ def convert_colors_to_hex_value(
     return f"#{to_hex(red)}{to_hex(green)}{to_hex(blue)}"
 
 
+def is_full_a1_notation(range_name: str) -> bool:
+    """Check if the range name is a full A1 notation.
+    "A1:B2", "Sheet1!A1:B2" are full A1 notations
+    "A1:B", "A1" are not
+
+    Args:
+        range_name (str): The range name to check.
+
+    Returns:
+        bool: True if the range name is a full A1 notation, False otherwise.
+
+    Examples:
+
+        >>> is_full_a1_notation("A1:B2")
+        True
+
+        >>> is_full_a1_notation("A1:B")
+        False
+    """
+    if A1_ADDR_FULL_RE.search(range_name):
+        return True
+    return False
+
 if __name__ == "__main__":
     import doctest
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -991,15 +991,13 @@ class Worksheet:
 
         values = response.get("values", [])
 
+        # range_name must be a full grid range so that we can guarantee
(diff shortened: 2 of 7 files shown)


lavigne958:
Hi I'll look at the code when I get a chance.

About the default behavior I'm not too sure. This might be a big thing or not. I have no clue what users expect and if that's gonna break something.

We can introduce it. Introduce all the breaking change we have first then maybe, do a release 7.0.0 not too far later with this kind of breaking change. What do you think ?

alifeee:
> What do you think ?

I'm not sure. I don't have a good enough understanding of who uses gspread and how they use it to say whether this would have a large or small affect on people.

I think: people could call "A1:C4" and get a range that is not 3x4, and then be confused. They could then either:

- make their own function to pad the values
- read the gspread documentation and find "maintain_size"

I think the latter is less likely, which is why I suggested to make it default behaviour. If people did option 1 above, then their code would break, but they would just have to remove their padding function becase we provide it natively.

gspread/utils.py
```diff
@@ -867,6 +868,46 @@ def to_hex(value: float) -> str:
...
+
+        >>> is_full_a1_notation("A1:B2")
+        True
+
+        >>> is_full_a1_notation("A1:B")
+        False
+    """
+    if A1_ADDR_FULL_RE.search(range_name):
+        return True
+    return False
```
lavigne958, line 892: usually in programming languages doing:

```
if X
  return true
else
  return false
```

is idiomatic.

You should always find a way to do:

```
return X is True
```

or

```
return X is False
```

Here we could simplify with
```suggestion
return A1_ADDR_FULL_RE.search(range_name) is not None
```

gspread/worksheet.py
```diff
@@ -985,6 +989,21 @@ def get(self, range_name=None, **kwargs):
 
         response = self.spreadsheet.values_get(range_name, params=params)
 
+        values = response.get("values", [])
+
+        if kwargs["maintain_size"] is True and is_full_a1_notation(range_name):
+            a1_range = get_a1_from_absolute_range(range_name)
```
lavigne958, line 995: I feel like we overwrite the `range_name` above, or we could simply take the range_name and construct the complete range name for the GET request, then here we could simply do a call to `a1_range_to_grid_range` instead of de-constructing what we just built above.

it's no improvement it's just simpler to me, but it's up to you.

gspread/worksheet.py
```diff
@@ -985,6 +989,21 @@ def get(self, range_name=None, **kwargs):
...
+            grid_range = a1_range_to_grid_range(a1_range)
+            rows = grid_range.get("endRowIndex", self.row_count) - grid_range.get(
+                "startRowIndex", 0
+            )
+            cols = grid_range.get("endColumnIndex", self.col_count) - grid_range.get(
+                "startColumnIndex", 0
+            )
```
lavigne958, line 1002: I see the point of having the `self.row_count` but they are not reliable. the file could have been updated in between the last check and now.

As long as we are under the `if is_full_a1_notation` we can expect the keys `endRowIndex` and `endColumnIndex` 
What could be done here is a comment explaining we need to keep the check to make sure we have a fully bounded range so we can get the _end_ keys.

lavigne958:
> > What do you think ?
> 
> I'm not sure. I don't have a good enough understanding of who uses gspread and how they use it to say whether this would have a large or small affect on people.
> 
> I think: people could call "A1:C4" and get a range that is not 3x4, and then be confused. They could then either:
> 
> * make their own function to pad the values
> * read the gspread documentation and find "maintain_size"
> 
> I think the latter is less likely, which is why I suggested to make it default behaviour. If people did option 1 above, then their code would break, but they would just have to remove their padding function becase we provide it natively.

I agree, I would expect it too, the size of what I request to match.

What I am afraid of is if users have some kind of code around that, would it break ?

for sure the new parameter allows users to change using it for now, and then next major release we make default and then boom it's automatic.

gspread/worksheet.py
alifeee: The user provided `range_name` could be an absolute range name, so we have to parse it with `get_a1_from_absolute_range`.

gspread/worksheet.py
alifeee: yes

alifeee:
This line of the CI fails now

https://github.com/burnash/gspread/blob/07aab27de6761ddc26705040a9d374b9a7eafcaa/.github/workflows/main.yaml#L33

Not sure why...

alifeee:
CI fixed with #1319. Now ready for review

lavigne958 (commented):
In order for the users to notice the new parameter we should probably:

- update the examples on the readme.md and use it about everywhere ?
- add a mention in the readme.md about the new argument ?

alifeee:
Yes, that sounds good to me. I will be able to do it from next week.

lavigne958 (approved):

PR #1312, opened 2023-10-05
Fix warning message for `worksheet.update` method

### LINKED ISSUE #1310: Worksheet.update change in arguments and deprecation

-> Always appreciate your work for the community

---

**Describe the bug**

-> I got the below warning

```
.env\lib\site-packages\gspread\worksheet.py:1069: UserWarning: [Deprecated][in version 6.0.0]: method signature will change to: 'Worksheet.update(value = [[]], range_name=)' arguments 'range_name' and 'values' will swap, values will be mandatory of type: 'list(list(...))'
  warnings.warn(
```
I somewhat think that the warning message is vague, in that the part after `change to`  has an errata of `Worksheet.update(value*s*...` and overall I think the change is not clear; why not rewrite the warning into something like below?
```
.env\lib\site-packages\gspread\worksheet.py:1069: UserWarning: [Deprecated][in version 6.0.0]: method signature's arguments 'range_name' and 'values' will change their order. We recommend using named arguments. In addition, the argument 'values' will be mandatory of type: 'List[List]'. 
  warnings.warn(
```

**To Reproduce**
1. I got this warning just reproducing the first example code as shown in below.

**Code example***
```python
import gspread

# Open a sheet from a spreadsheet in one go
wks = gc.open("gspread_test").sheet1

# Update a range of cells using the top left corner address
wks.update('A1', [[1, 2], [3, 4]])

# Or update a single cell
wks.update('B3', "it's down there somewhere, let me take another look.")

# Format the header
wks.format('A1:B1', {'textFormat': {'bold': True}})
```

**Environment info:**
- Operating System [e.g. Linux, Windows, macOS]: Windows
- Python version: Python 3.10.4
- gspread version: gspread==5.11.3

**Additional**
I get confused, does this warning appear for every .update method? Not even when I change the type of values into a list of list?
Might be better to update your README for the best practices of using the update method.

Thanks alot

ksj20:
closes #1310 

Fixed the warning message for the worksheet update method for better readability

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -1187,8 +1187,9 @@ class Worksheet:
         warnings.warn(
             DEPRECATION_WARNING_TEMPLATE.format(
                 v_deprecated="6.0.0",
-                msg_deprecated="method signature will change to: 'Worksheet.update(value = [[]], range_name=)'"
-                " arguments 'range_name' and 'values' will swap, values will be mandatory of type: 'list(list(...))'",
+                msg_deprecated="Method signature's arguments 'range_name' and 'values' will change their order."
+                " We recommend using named arguments for minimal impact. In addition, the argument 'values' will be mandatory of type: 'List[List]'."
+                " (ex) Worksheet.update(values = [[]], range_name=) ",
             )
         )
         if is_scalar(range_name):


alifeee:
Thanks for this change! :)

The warning is clearer now.

Now that you understand why the warning is in place, I wonder if you have any opinion on this:

> I get confused, does this warning appear for every .update method? Not even when I change the type of values into a list of list?

Can you think of a way of validating that a user has used kwargs, and not args? I can't. If there is no way, then this PR is complete and we can merge it now.

ksj20:
1. add the warning under the is_scalar if branch https://github.com/burnash/gspread/blob/d076c7055a0365c5445c24913db061ecb6becfdf/gspread/worksheet.py#L1195
2. I guess you are looking for only allowing kwargs, something like the below? Then probably catch the TypeError? This is Python 3.10.4 by the way
```
>>> def func(*, a=None, b=None):
...     return a, b
...
>>> func(1,2)
TypeError: func() takes 0 positional arguments but 2 were given
>>> func(a=1,b=2)
(1, 2)
```

alifeee:
It was my understanding that the primary purpose of the warning was not to ensure that `values` was a 2D array, but to warn of the swapping of `values` and `range` in the next version (6.0.0)

Thus, I was thinking:

"Is there a way of telling if the user called `update("A1:A2", [[1, 2]])` or `update(range_name="A1:A2", values=[[1, 2]])`?"

If there is:

- we can make the warning only happen for the former

If there isn't:

- we have to send the warning even if the user is doing the "correct" behaviour

alifeee:
That is good to know.

However, I would not like to *disallow* using arguments. It should always be allowed in version 5, otherwise it will break people's code. I would like to be able to see if someone has used named arguments, while still allowing the use of normal arguments.

ksj20:
Yeah, look at this StackOverflow question, [Can I distinguish positional and keyword arguments from inside the called function?](https://stackoverflow.com/questions/68432070/can-i-distinguish-positional-and-keyword-arguments-from-inside-the-called-functi)

According to the SO question, to achieve what you said it should be something like this:
```python
def update(*args, range_name=None, values=[[]]):
    if args: # args is not empty - user passed deprecated positional arguments
        print((
            "Method signature's arguments 'range_name' and 'values' will change their order."
            " We recommend using named arguments for minimal impact. In addition, the argument 'values' will be mandatory of type: 'List[List]'."
            " (ex) Worksheet.update(values = [[]], range_name=) "
        ))
        range_name = args[0]
        if len(args) >= 2:
            values = args[1]
    return range_name, values

update('A5', [[1,2,3]]) # Warning appears
update(range_name=None, values=[[1,2,3]]) # No warning
```

alifeee:
Ah, I missed that when I read the thread! Good investigating :)

> According to the SO question, to achieve what you said it should be something like this:
> ...

That looks good :). Would you like to implement it with this PR?

> [...] for version 6 I think you should be using asterisks before named arguments to make the function more strict

I disagree, I think it is fine to leave it callable using arguments. I will also ask @lavigne958's opinion, since this is v6 related.

lavigne958:
Hi I think this is a bit overkilled, the version 6 is still not ready but at some point it will be and we don't want to wait too much either. 

To le we should keep the warning and as soon as we can we release version 6. 

In this way, if the user has updated the parameters I would recommend using some env variables to disable the warning. If the user sets the env variable we don't print the warning. 

As simple as that. And it's the users choice to disable that warning message.

alifeee:
hey, thanks again for this. will merge

PR #1320, opened 2023-10-19
Merge `master` into `feature/release_6_0_0`

alifeee:
Done apart from one test fails. Not sure why.

### DIFF
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -63,19 +63,6 @@ class Spreadsheet:
         return self._properties["modifiedTime"]
 
     @property
-    def updated(self):
-        """.. deprecated:: 2.0
-
-        This feature is not supported in Sheets API v4.
-        """
-        warnings.warn(
-            "Spreadsheet.updated() is deprecated, "
-            "this feature is not supported in Sheets API v4",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-
-    @property
     def timezone(self):
         """Spreadsheet timeZone"""
         return self._properties["timeZone"]
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -7,6 +7,7 @@ This module contains common spreadsheets' models.
 """
 
 from typing import Union
+import warnings
 
 from .exceptions import WorksheetNotFound
 from .urls import DRIVE_FILES_API_V3_URL, SPREADSHEET_DRIVE_URL
@@ -23,9 +24,6 @@ class Spreadsheet:
 
         metadata = self.fetch_sheet_metadata()
         self._properties.update(metadata["properties"])
-
-        drive_metadata = self.client.get_file_drive_metadata(self._properties["id"])
-        self._properties.update(drive_metadata)
 
     @property
     def id(self):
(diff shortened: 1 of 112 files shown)


alifeee:
This test fails

https://github.com/burnash/gspread/blob/fc336cbc98e2c65941915413655e28a566fca753/tests/worksheet_test.py#L148-L172

With the failure:

```bash
        if response.ok:
            return response
        else:
>           raise APIError(response)
E           gspread.exceptions.APIError: {'code': 400, 'message': 'Invalid value 
at \'data.values\' (type.googleapis.com/google.protobuf.ListValue), "A1:D4"', 'status': 'INVALID_ARGUMENT', 'details': [{'@type': 'type.googleapis.com/google.rpc.BadRequest', 'fieldViolations': [{'field': 'data.values', 'description': 'Invalid value at \'data.values\' (type.googleapis.com/google.protobuf.ListValue), "A1:D4"'}]}]}
```

Looks like setting the initial data isn't working.

Ok, I realised this was because `update` switched values. This was hard to spot. Maybe we should catch this in v6 and raise a GspreadException if the arguments are the wrong way round

alifeee:
I have merged `master` into `feature/release_6_0_0`

Pending @lavigne958's review. See changed files.

gspread/spreadsheet.py
```diff
@@ -45,8 +43,25 @@ def url(self):
...
+    def lastUpdateTime(self):
+        """Spreadsheet last updated time.
+        Only updated on initialisation.
+        For actual last updated time, use get_lastUpdateTime()."""
+        warnings.warn(
+            """
+            This is only updated on initialisation and is probably outdated by the time you use it.
+            For an up to date last updated time, use get_lastUpdateTime().
+            """
+        )
```
lavigne958, line 60: should we update the warning and mention something like: deprecated, please use `get_lastUpdateTime()` ?

gspread/worksheet.py
```diff
@@ -535,14 +541,8 @@ def get_all_values(
...
-        empty2zero: bool = False,
-        head: int = 1,
-        default_blank: str = "",
-        allow_underscores_in_numeric_literals: bool = False,
-        numericise_ignore: List[int] = [],
-        value_render_option: Optional[ValueRenderOption] = None,
-        expected_headers: Optional[List[Union[str, int, float]]] = None,
-    ) -> List[Dict[str, Union[int, float, str]]]:
+        **kwargs,
+    ):
```
lavigne958, line 545: we lost the typing of the returned value here

lavigne958:
> Ok, I realised this was because `update` switched values. This was hard to spot. Maybe we should catch this in v6 and raise a GspreadException if the arguments are the wrong way round

hum true that :thinking:  do we have a way to validate  that ?  something like checking the type of the values may be ?

gspread/spreadsheet.py
alifeee: probably, yes. I'll do this.

alifeee:
> hum true that 🤔 do we have a way to validate that ? something like checking the type of the values may be ?

I think we can do that. Could you create another issue to track it?

lavigne958 (approved):

PR #1322, opened 2023-10-19
add test_get_and_get_values_have_same_signature

### LINKED ISSUE #1303: Add tests for function arguments
Some functions are aliases of other functions. For example:

- `get_all_values` and `get_values`
- `get_all_records` and `get_records`

Over time, we may add kwargs and args to the original functions, and forget to add these to the aliases. We should add some simple tests to the test files that verify that the function arguments/kwargs are the same.

This issue alread arose once with `get` vs `get_values`.

alifeee:
closes #1303

the test currently fails, as `get` and `get_values` have different signatures. we must wait for #1296 to finish before this test.

```bash
AssertionError: <Signature (range_name=None, **kwargs)> != <Signature (range_name=None, combine_merged_cells=False, **kwargs)>
```

### DIFF
--- a/tests/cassettes/WorksheetTest.test_get_and_get_values_have_same_signature.json
+++ b/tests/cassettes/WorksheetTest.test_get_and_get_values_have_same_signature.json
@@ -1,0 +1,381 @@
+{
+    "version": 1,
+    "interactions": [
+        {
+            "request": {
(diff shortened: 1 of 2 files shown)


alifeee:
included in #1296

Issue #1334, opened 2023-10-25
`combine_merged_cells` cannot capture all merges


alifeee:
Merges are actually just one value in the top left cell, and some metadata. If the top-left cell is not in your requested range, you cannot faithfully recreate the merge. For example:

![image](https://github.com/burnash/gspread/assets/13833017/228c8a89-462f-4072-b989-9e41ba133172)

If you called `worksheet.get_values("F3:H4", combine_merged_cells=True)`, what would you expect?

1

| "big merge" | 2 | 3 |
| --- | --- | --- |
| 4 | 5 | 6 |

or 2

|  | 2 | 3 |
| --- | --- | --- |
| 4 | 5 | 6 |

We cannot recreate 1 without added complexity. We must return 2. This may be unexpected.

`combine_merged_cells` was added in #1215. Its addition has brought several problems (#1298, #1330).

What should we do? If we leave it in, it will always have this issue. If we remove it, it will be sad.

lavigne958:
Users will expect the value `Big Merge` in `F3`. We should add the extra complexity. When we list the merges and see that one overlaps the requested range then we should get and fill what is necessary.

For sure this will take some time and either add a very large util function or introduce a big code change in `Worksheet.py`

We should start working on it after 6.0.0 is out that would be a great improvement to bring after the release.

lavigne958:
Thinking about it I would solve it as:
- we get values from a range 
- if user requests combined merged values
- get the list of merges from the API
- check if any merge range overlaps the requested range in all boundaries 
   - in order to solve this, we might find some already optimized algorithm online 🤔
   - we could rely on some third party library too (check some classic packages like bumpy etc )
- for any overlapping merge set the values in the "empty" cells.

A different solution that we can try is: open an issue on Google side and request some API  changes or some extra parameters that does it all for us.

alifeee:
I think this is a very large change with lots to think about. `combine_merged_cells` has already caused two minor releases due to bugs (#1298 #1330). I am worried that making it even more complex is the wrong direction to go in. I think it is a small feature in the context of the size of gspread and not worth the effort to fix this issue.

I would rather add a warning that says it will not pull in data from outside the requested range, or as you say submit a request to Google's API.

alifeee:
pushing this past 6.0.0 release

PR #1335, opened 2023-10-25
fix `combine_merged_cells` when using from a range that doesn't start at `A1`

### LINKED ISSUE #1330: Combining merged cells producing incorrect result when used on a non-default cell range
**Describe the bug**
Calling `Worksheet.get_values()` with `combine_merged_cells=True` and a range name not starting at column A or row 1 ("B:C" for example) produces incorrect results when expanding the merged cells content.

**To Reproduce**
Create a worksheet with cells B1 and B2 merged and some other content on C1 and C2.
For example:

| A | B | C |
| --- | --- | --- |
| | foo | dummy 1 |
| |  | dummy 2 |

A call to `ws.get_values("B:C", combine_merged_cells=True)` outputs:

- ["foo", "dummy 1"]
- ["", "dummy 1"] => no "foo" and "dummy 1" instead of "dummy 2"

**Expected behavior**
The content of B1 is expanded and available on both rows:

- ["foo", "dummy 1"]
- ["foo", "dummy 2"]

**Environment info:**
- Operating System: Linux,
- Python version: 3.11
- gspread version: 5.11.3

**Additional context**
The issue is that the `combined_merge_values()` function does not take into account the requested cell range to offset the value expansion.

alifeee:
closes #1330 

add `start_row_index` and `start_col_index` to `utils.combined_merge_values` so that ranges that don't start at col 0, row 0 (i.e., `A1`) can have their merged cells merged.

n.b., not starting from the top left introduces more issues with `combine_merged_cells` -> #1334

This code is added to `worksheet.get_values`. This is to get the `start_row_index` and `start_col_index` from the requested range.

https://github.com/burnash/gspread/blob/0e698de225dadbce97e9bb085ed2ad1eaa7d22ce/gspread/worksheet.py#L495-L503

- [x] This may fail if a named range is used (see [A1 notation docs](https://developers.google.com/sheets/api/guides/concepts#a1_notation)) - add tests? or do not allow `combine_merged_cells` if a named range is used? since we don't know the start index?

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -14,7 +14,7 @@ from collections.abc import Sequence
 from functools import wraps
 from itertools import chain
 from math import inf
-from typing import Mapping
+from typing import Mapping, Optional
 from urllib.parse import quote as uquote
 
 from google.auth.credentials import Credentials as Credentials
@@ -910,17 +910,20 @@ def is_full_a1_notation(range_name: str) -> bool:
     return A1_ADDR_FULL_RE.search(range_name) is not None
 
 
-def get_a1_from_absolute_range(range_name: str) -> str:
+def get_a1_from_absolute_range(range_name: Optional[str]) -> str:
     """Get the A1 notation from an absolute range name.
     "Sheet1!A1:B2" -> "A1:B2"
     "A1:B2" -> "A1:B2"
+    None -> ""
 
     Args:
-        range_name (str): The range name to check.
+        range_name (str | None): The range name to check.
 
     Returns:
         str: The A1 notation of the range name stripped of the sheet.
     """
+    if range_name is None:
+        return ""
     if "!" in range_name:
         return range_name.split("!")[1]
     return range_name
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -14,7 +14,7 @@ from collections.abc import Sequence
 from functools import wraps
 from itertools import chain
 from math import inf
-from typing import Mapping, Optional
+from typing import Mapping
 from urllib.parse import quote as uquote
 
 from google.auth.credentials import Credentials as Credentials
@@ -910,20 +910,17 @@ def is_full_a1_notation(range_name: str) -> bool:
     return A1_ADDR_FULL_RE.search(range_name) is not None
 
 
-def get_a1_from_absolute_range(range_name: Optional[str]) -> str:
+def get_a1_from_absolute_range(range_name: str) -> str:
     """Get the A1 notation from an absolute range name.
     "Sheet1!A1:B2" -> "A1:B2"
     "A1:B2" -> "A1:B2"
-    None -> ""
 
     Args:
-        range_name (str | None): The range name to check.
+        range_name (str): The range name to check.
 
     Returns:
         str: The A1 notation of the range name stripped of the sheet.
     """
-    if range_name is None:
-        return ""
     if "!" in range_name:
         return range_name.split("!")[1]
     return range_name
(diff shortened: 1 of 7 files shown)


alifeee:
@lavigne958 once this is approved and merged, we can release 5.12.1.

This release should be the last before 6.0.0, so without another (5.12.2 e.g.), this will also the last chance to get an updated README onto the PyPi (the readme on the GitHub will always stay up to date)

lavigne958:
> * This may fail if a named range is used (see [A1 notation docs](https://developers.google.com/sheets/api/guides/concepts#a1_notation)) - add tests? or do not allow `combine_merged_cells` if a named range is used? since we don't know the start index?

When we request: same value for merged cells, we make a second call to the spreadsheet right ? to pull all the merges.
Can we extract the coordinates of the named range at the same time ? :thinking:

gspread/utils.py
```diff
@@ -894,17 +910,20 @@ def is_full_a1_notation(range_name: str) -> bool:
...
 
     Args:
-        range_name (str): The range name to check.
+        range_name (str | None): The range name to check.
 
     Returns:
         str: The A1 notation of the range name stripped of the sheet.
     """
+    if range_name is None:
+        return ""
```
lavigne958, line 926: I would recommend not ot change that function that is used by other worksheet method a bit everywhere. and instead pass a range name or skip calling it as this is made to handle a range name and return some value if we don't have a range name then we don't call that function.

In my opinion

gspread/worksheet.py
```diff
@@ -485,13 +485,23 @@ def get_values(self, range_name=None, combine_merged_cells=False, **kwargs):
...
             if combine_merged_cells is True:
                 spreadsheet_meta = self.spreadsheet.fetch_sheet_metadata()
                 worksheet_meta = finditem(
                     lambda x: x["properties"]["title"] == self.title,
                     spreadsheet_meta["sheets"],
                 )
-                return combined_merge_values(worksheet_meta, vals)
+                grid_range = a1_range_to_grid_range(
+                    get_a1_from_absolute_range(range_name),
+                )
```
lavigne958, line 497: regarding above comment I would suggest we check if range_name is `None` or not. then call the absolute or not.

I would suggest something like a filter pattern:

```python
if range_name is no None:
  range_name = get_a1_from_absolute_range(range_name)

grid_range = a1_range_to_grid_range(range_name)
```

gspread/utils.py
alifeee: I agree

alifeee:
> When we request: same value for merged cells, we make a second call to the spreadsheet right ? to pull all the merges.
> Can we extract the coordinates of the named range at the same time ? 🤔

Yes. In fact, we already request the spreadsheet metadata to get the merges:

https://github.com/burnash/gspread/blob/bcac8cb02ff58bda6355bf2b1ada131df3af32ed/gspread/worksheet.py#L490

So, I have reused this metadata to also find the `grid_range` of a named range:

https://github.com/burnash/gspread/blob/bcac8cb02ff58bda6355bf2b1ada131df3af32ed/gspread/worksheet.py#L496-L507

and other range types are in a different branch

https://github.com/burnash/gspread/blob/bcac8cb02ff58bda6355bf2b1ada131df3af32ed/gspread/worksheet.py#L508-L514

This may be too much logic for `get_values` and you might prefer it be offloaded to a `utils` function. It also is only implemented for `get_values`, as `get` and `get_values` have not yet been combined (happens in v6.0.0)

lavigne958 (approved):

lavigne958:
let's get release 5.12.1 out

alifeee:
agreed. Does that include a fix for #1354?

PR #1336, opened 2023-10-28
Warning/update swapped args

### LINKED ISSUE #1325: Find a way to catch wrong argument order in `Worksheet.update()`
## Overview
the method `Worksheet.update()` has change the order of its arguments.

Users may break their code when upgrading to v6.0.0

find a way if possible to identify the argument list has not been updated.

created from #1320

alifeee:
closes #1325

If they "seem" swapped, I swap them back and emit a warning. We may instead want to just `raise` a `ValueError`. Up to you @lavigne958

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -7,6 +7,7 @@ This module contains common worksheets' models.
 """
 
 import re
+import warnings
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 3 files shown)


lavigne958 (approved):
This is fine to me, this will help people use gspread even though their code is not correct ans possibly a lot of users won't change their code as may no understand what is the difference between argument list and _key-value arguments_ too.

PR #1337, opened 2023-10-31
Add type checking to lint

### LINKED ISSUE #1321: Add missing type hints
This work should be done in the [6.0.0 release branch](https://github.com/burnash/gspread/tree/feature/release_6_0_0/)

Run `mypy --install-types --non-interactive --ignore-missing-imports ./gspread` (from CI) to see which types are broken/missing

alifeee:
closes #1321

I have added `mypy` type checking to the lint workflow, so it now runs when you run

```bash
tox -e lint
```

I also fixed some typing problems. `mypy` now runs without error.

Is there anything else to type @lavigne958? or is this everything?

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -29,7 +29,6 @@ jobs:
         if: ${{ matrix.python >= '3.8' }}
       - run: tox -e lint
       - run: tox -e py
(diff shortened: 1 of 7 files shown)


bpshaver:
Consider https://github.com/burnash/gspread/pull/1339 a suggestion. You can only install strenum if the python version is below 3.11.

lavigne958 (approved):

PR #1340, opened 2023-10-31
chore(deps): strenum dependency only installed for Python versions be…

bpshaver:
…low 3.11

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -7,6 +7,7 @@ This module contains utility functions.
 """
 
 import re
+import sys
 from collections import defaultdict
 from collections.abc import Sequence
 from functools import wraps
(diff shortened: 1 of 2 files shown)


lavigne958:
Hi, thank you for your contribution, I'm sorry but I don't understand why we should select a different import depending on the python version. 
I check the Pypi page of `StrEnum` and it works for python version from 3.8 to 3.12 the same way, see here: https://pypi.org/project/StrEnum/

bpshaver:
No problem. Its just unnecessary to install `strenum` via `pip` if the end user is running Python 3.11 or above. You either have an ugly import statement or an unnecessary dependency. I choose the former, but its personal preference.

lavigne958:
> No problem. Its just unnecessary to install `strenum` via `pip` if the end user is running Python 3.11 or above.
> 
> Edit: You can _probably_ assume `strenum` will continue to work without any bugs or security issues in all future versions of Python, but that is just an assumption.

I understand better now, thank you for pointing out that part.

We need to be compliant with 5 different version of python, as much as we can we try not to introduce python version specific code. so far if installing the extra library works for all versions **and** is mandatory for any version bellow 3.11 we'll keep it and remove that dependency once we only handle version 3.11 and onward. this will keep it easier for us to maintain.

hope you understand.

bpshaver:
Of course! It is an interesting but ultimately minor decision. 

I tend to believe `explicit is better than implicit` but there are other bits from the Zen of Python would point the other way in this, e.g. `beautiful is better than ugly`. 

And pinning to a specific version (`==0.4.15`) is likely to address some issues, too.

PR #1342, opened 2023-11-01
Improve `Worksheet.sort()` signature

### LINKED ISSUE #1090: rework `sort` method signature

lavigne958:
Instead of using generic `kwargs`
Mention the argument `range` explicitly and add it's type.

closes #1090

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -14,6 +14,7 @@ from typing import (
     Iterable,
     Iterator,
     List,
+    Literal,
     Mapping,
     MutableMapping,
     Optional,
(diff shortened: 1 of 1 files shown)


alifeee (approved):
lgtm!

lavigne958:
I based this branch from missing type hints. When she base merged this PR should automatically change its base to feature/release_6_0_0

PR #1343, opened 2023-11-05
remove `use_index` and references to it in `get_records`

### LINKED ISSUE #1332: Remove unused parameter or use it
## Overview

I noticed the following unused parameter, we should either use it, or remove it. 

https://github.com/burnash/gspread/blob/f6d1b1b79752503012230f595a4f55ebd231a702/gspread/worksheet.py#L574

I'll have a look at the code itself a bit longer to check what is best.

alifeee:
closes #1332

`use_index` was added by #1301 to make implementing #808 easier in future

we can add it back later when the column-indexing functionality is added

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -571,6 +571,7 @@ class Worksheet:
         self,
         empty2zero=False,
         head=1,
+        use_index=0,
         first_index=None,
         last_index=None,
         default_blank="",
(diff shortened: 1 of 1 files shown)


lavigne958 (commented):
Wait I just thought of something, what if someone used it currently with the value 0? I suppose it would break their code the remove the argument 🤦‍♂️

Like if someone has the following code:

`sh.get_record(use_index=0)`

This works fine and of we removed it some users won't understand 😑

We need to add a new warning then we introduce the removal in the next version, again unless we leave it if it's easier. Now to me it seems it's more work to remove it than keep it 😒

alifeee:
you raise a good point. This PR now deprecates `use_index`, and #1345 removes it (from 6.0.0). I do not think this is a big enough change requiring an entry in the migration guide (since get_records was only added a few weeks ago)

lavigne958:
> you raise a good point. This PR now deprecates `use_index`, and #1345 removes it (from 6.0.0). I do not think this is a big enough change requiring an entry in the migration guide (since get_records was only added a few weeks ago)

great changes thanks !

lavigne958 (approved):

PR #1345, opened 2023-11-05
Remove/get records   use index

alifeee:
remove `get_records` (see #1332)


alifeee:
CI currently failing, will be fixed by #1344

lavigne958 (approved):

PR #1349, opened 2023-11-07
make deprecation warning dependent on if kwarg is used for client_factory

alifeee:
now, if a user does not specify client_factory, they will not get a warning


lavigne958 (approved):
good catch ! this will be less bothering for users :rocket:

Issue #1355, opened 2023-11-28
sheet.get_all_records() throws index error when sheet has no records


s2t2:
**Describe the bug**
sheet.get_all_records() throws an index error when the sheet is blank, or even when the sheet has only a header row with the column names.

**To Reproduce**
1. Make a blank sheet OR make a sheet with only a header row with the column names (but otherwise no records)
2. `sheet = doc.worksheet("your-sheet-name")`
3. `sheet.get_all_records()`

**Expected behavior**
The function should return an empty list, indicating there are no records.

**Code example***
see reproduction above

**Environment info:**
- Operating System [e.g. Linux, Windows, macOS]: mac, linux
- Python version: 3.x
- gspread version: various, including newest 5.12

lavigne958:
Hi thank you for raising the issue.

We are currently working on this method.

We'll bring a new fix soon.

This reminds me of a special case we used to handle, it looks we don't anymore.

Well fix it and update the tests too.

lavigne958:
thank @alifeee for taking this issue, I saw your code and checked your branch, I wait for you PR to be ready and will provide the review ASAP.

Issue #1354, opened 2023-11-27
Header whitespace regression in 5.12.0


timmc-edx:
5.12.0 no longer tolerates sheets with filled-in cells that are not under a header. (For example, there might be a summary table off to the side of the main data.)

## Repro data

aaa | bbb |   |  
-- | -- | -- | --
w | x |   |  
y | z |   | other

```
worksheet.get_all_records(expected_headers=['aaa'])
```

## Expected (old) behavior:

In 5.11.3 the output is `[{'aaa': 'w', 'bbb': 'x', '': ''}, {'aaa': 'y', 'bbb': 'z', '': 'other'}]`

## Unexpected (new) behavior:

5.12.0 raises `gspread.exceptions.GSpreadException: the header row in the worksheet contains multiple empty cells`

## Environment info

- Operating System: Ubuntu Linux
- Python version: 3.8
- gspread version: 5.12.0

alifeee:
hi! thanks for the issue :) 

This bug was fixed by https://github.com/burnash/gspread/pull/1353 after issue https://github.com/burnash/gspread/issues/1352 was filed.

This fix will be released in 5.12.1, which will release in the next month.

timmc-edx:
Oh, thanks! Somehow I missed that in my issue search.

timmc-edx:
Hmm... I think this is actually a different issue. I can still repro it with current master (093235844b00267).

alifeee:
yep

> Though we should keep in mind that in this particular scenario column C and D have empty headers so would collide in the check and will fail but in this case that is the wanted behavior. to prevent key/value overriding when building the dict.

Here, `expected_headers` is provided as `[]`, showing that the user wants to ignore duplicate headers.

PR #1357, opened 2023-11-30
Many fixes for `get_records`

### LINKED ISSUE #590: get_all_records() IndexError
I recently upgraded to the latest version of gspread (v 3.0.1). Now when I request certain worksheets with get_all_records() - I get the following error.

This same scripts ran without issues before I upgraded.

```
gameValues = sheet.get_all_records()
File "/Library/Python/2.7/site-packages/gspread/models.py", line 471, in get_all_records
keys = data[idx]
IndexError: list index out of range
```

I'm making the follow request:

`sheet = client.open(MyWorkbook).worksheet(MyWorksheet)`

MyWorkbook being the Workbook name and MyWorksheet being the Worksheet name.

The library seems to be retreieving the worksheet object fine when I print it.

`<Worksheet u'Schedule' id:242452861>`

alifeee:
closes #590, #629, #1354, #1355

### firstly, a test simplification

This is done for a few tests

```diff
...
- cell_list = self.sheet.range("A1:D4")
- for cell, value in zip(cell_list, itertools.chain(*rows)):
-   cell.value = value
- self.sheet.update_cells(cell_list
+ self.sheet.update("A1:D4", rows)
...
```

### add `test_get_all_records_with_blank_final_headers`

Regression test for #1354

tests `get_records` with some blank final headers, i.e.,

| "header 1" | "header 2" | "" | "" |
| --- | --- | --- | --- |

It should throw an error, but work fine with `expected_headers` set

### add `test_get_all_records_with_keys_blank`

Similar to above but with the entire keys row blank

### add `test_get_records_with_all_values_blank`

This is the beefy one. Regression test for #590, #629, #1355

Given the sheet

| "a" | "b" | "c" |
| --- | --- | --- |
| "" | "" | "" |
| "" | "" | "" |

What should `get_records` return?

- `[]` ?
- `[{"a": "", "b": "", "c": ""}, {"a": "", "b": "", "c": ""}]` ?

This is a hard question. My answer uses the kwarg `last_index`. If this is asked for (i.e., if a user asks for "five rows"), then I believe `get_records` should return a list of length asked for (i.e., length 5). If it is not asked for (i.e., if a user asks for "entire sheet"), then I believe `get_records` should return an empty list (not further because we "don't know" how long a spreadsheet is). This logic is explained by:

#### I ask for `get_records(first_index=2, last_index=4)`

I want `[{...}, {...}, {...}]`

#### I ask for `get_records()`

I want `[]`

#### I ask for `get_records(first_index=1)`

I want `[]`

#### I ask for `get_records(last_index=4)`

I want `[{...}, {...}, {...}]`

Thus, this is the logic I included in `test_get_records_with_all_values_blank`


alifeee:
> I will take a look ASAP. By the end of the weekend it should be done.

Super :) let me know if you think we should change anything

lavigne958:
Looks good to me :+1:

PR #1373, opened 2023-12-26
Fixed default value of merge_type parameter in merge_cells function docstring.

neolooong:
Fixed default value of merge_type parameter in merge_cells function docstring.

### DIFF
--- a/docs/api/utils.rst
+++ b/docs/api/utils.rst
@@ -4,6 +4,7 @@ Utils
 .. autonamedtuple:: gspread.utils.Dimension
 .. autonamedtuple:: gspread.utils.DateTimeOption
 .. autonamedtuple:: gspread.utils.ExportFormat
+.. autonamedtuple:: gspread.utils.MergeType
 .. autonamedtuple:: gspread.utils.MimeType
 .. autonamedtuple:: gspread.utils.PasteOrientation
 .. autonamedtuple:: gspread.utils.PasteType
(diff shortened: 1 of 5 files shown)


alifeee:
thanks, this is helpful! I always love a docs change :)

got me thinking, why aren't we using a tuple for this parameter? like for `DateTimeOption`?

https://github.com/burnash/gspread/blob/c3b61d0571bfc87a04fbbfe4e1cbcf6e2487fb6e/gspread/utils.py#L41-L43

perhaps we could change it... @lavigne958 thoughts ?

lavigne958:
Thanks for the doc update.

I think we missed it 😮
Yep we can upgrade it to a namedtuple but we should make sure people can keep using strings and it works. That I'm not sure 😬 if that does not work we can't change it and should just update the documentation

neolooong:
I am sure that after changing to namedtuple, people still can use string.

alifeee:
quick change! it looks good :)

now is probably also the time to add a test for `merge_cells` (and a `merge_type` test too), since there is not one in [`worksheet_test.py`](https://github.com/burnash/gspread/blob/master/tests/worksheet_test.py)

you are welcome to give this a go, or one of us will do it if you would rather not deal with the tests :)

lavigne958 (approved):
thanks for the fix ! 
I agree if you feel like it, feel free to add a test on merge cells, if not, no worries we'll add them.

Tested it, it works in all bellow cases:
- with `sheet.merge_cell("A1:B2", "MERGE_ALL")`
- with `sheet.merge_cell("A1:B2", gspread.utils.MergeType.merge_all)`
which makes it ISO functional with current code base :+1:

alifeee:
super. we will merge this and add a test for `merge_cells` in a week or so if there is no more activity here :)

lavigne958 (approved):
Thanks for the new tests !

Issue #1376, opened 2024-01-15
add `get_records` changes to README migration guide


alifeee:
The result of issue #1367 should be summarised and added to the [README v6 migration guide](https://github.com/burnash/gspread#v512-to-v60-migration-guide)

After that, we can make the v6.0.0 release

alifeee:
mention that users can use `to_record`

https://github.com/burnash/gspread/pull/1377#event-11492744453

PR #1377, opened 2024-01-15
Add util function `to_records` to build records

### LINKED ISSUE #1367: PROPOSAL: changes to `get_records`/`get_all_records`
`get_all_records` is a method which has provided much pain to maintain recently (issues #1363, #1355, #1354, #1352, #1294) (PRs #1364, #1357, #1345, #1343, #1301, #1028)

We propose that we simplify the method, and reduce its scope.

## New scope

Given a spreadsheet, `get_all_records` turns a header and following content into a list of records

<details>
<summary>API information</summary>

The API call will return to the pre-5.12 case:

https://github.com/burnash/gspread/blob/9cc862bc24f61de763944c59b2e1d270ef3b48cf/gspread/worksheet.py#L547

Even if the header is set to the 5th row, we still need to get all values, as we do not know the "final row" so we cannot use a row range like "5:999"

</details>

| my sheet | | |
| --- | --- | --- |
| fruit | price | |
| apple | 56 | |
| orange | 23 | |

```python
> get_all_records(header=2)
[
  {"fruit": "apple", "price": "56"},
  {"fruit": "orange", "price": "23"},
]
```

### Simplify the "duplicate headers" case

```diff
- get_all_records(expected_headers=["apple", "orange"])
+ get_all_records(allow_duplicate_headers=True)
```

### `get_records` method, i.e., range_specification.

This means that headers + data will be obtained in a *single* API call. This simplifies a lot since we no longer have to pad the header/values to be the same length (#1357), as they will be the same length by nature of being from the same API call.

```diff
- get_records(first_index=10, last_index=15)
```

### `empty2zero` kwarg.

Its job can be done by existing `default_blank`

<details>
<summary>Details</summary>

Technically these differ currently. `default_blank` is used to replace a *single* blank header (not more), and *also* replace blank values. However, I think it should change to only replace values. Blank header should be left as-is and if there are multiple, duplicate headers can be ignored (see above).

</details>

```diff
- get_all_records(empty2zero=True)
+ get_all_records(default_blank=0)
```

### `utils.to_records`

Add a utils function to turn a list of headers + values into a list of dictionaries

<details>
<summary>implementation details</summary>

this will basically be a wrapper of

https://github.com/burnash/gspread/blob/495a3290612755a433c06d7ce5a70b0271d0c695/gspread/worksheet.py#L757

</details>

```diff
+ utils.to_records(header, values)
```

This can be put into documentation to tell people how to make their own list of records, given a header and values.

```diff
+ Want to use get_records differently? You can make your own by using `utils.to_records`:
+ header = worksheet.get("4:4")
+ values = worksheet.get("20:40")
+ records = utils.to_records(header, values)
```

lavigne958:
Add a new util function that helps building record values. Records are list of dictionaries, each dictionary has the same set of keys, they are set from the given list of headers. Each key is associated to a value from the given matrix of values. They are as many dictionaries as they are lines in the matrix.

Each line in the matrix will be associated to a dictionary, each key is associated to a value from that line in the given order.

Example:
Headers: A B C
Values:
1 2 3
7 8 9
Result: [ { A: 1, B: 2, C: 3}, {A: 7, B: 8, C: 9}]

closes #1367

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -878,6 +878,16 @@ def get_a1_from_absolute_range(range_name: str) -> str:
     return range_name
 
 
+def to_records(
+    headers: Iterable[Any] = [], values: Iterable[Iterable[Any]] = [[]]
+) -> List[Dict[str, Union[str, int, float]]]:
+    """Builds the list of dictionaries, all of them have the headers sequence as keys set,
+    each key is associated to the corresponding value for the same index in the sub-list.
+    There are as many dictionaries as they are entry in the of given values."""
+
+    return [dict(zip(headers, row)) for row in values]
+
+
 # SHOULD NOT BE NEEDED UNTIL NEXT MAJOR VERSION
 # def deprecation_warning(version: str, msg: str) -> None:
 #     """Emit a deprecation warning.
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -55,6 +55,7 @@ from .utils import (
     is_full_a1_notation,
     numericise_all,
     rowcol_to_a1,
+    to_records,
 )
 
 CellFormat = TypedDict(
@@ -498,10 +499,10 @@ class Worksheet:
 
             # Read all rows from the sheet
             >>> worksheet.get_all_records()
-            {
+            [
                 {"A1": "A6", "B2": "B7", "C3": "C8"},
                 {"A1": "A11", "B2": "B12", "C3": "C13"}
-            }
+            ]
         """
         entire_sheet = self.get(
             value_render_option=value_render_option,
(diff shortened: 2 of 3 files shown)


alifeee:
this is good stuff, thank you :)

we should also make this function more visible to users - it should be added to the docstring of `get_all_records`

(and migration guide readme - covered by #1376)

alifeee (changes requested):
see above comment

alifeee:
> Thanks ! It's actually shorter than what I expected 🙄

haha.. yes! a great one-liner eh

> What do you think of the outcome of that global change ?
> Are you satisfied with it ?

Do you mean the simplification of `get_all_records`? I think I am happy with it. I will have to have another look. What do you think?

lavigne958:
> > What do you think of the outcome of that global change ?
> > Are you satisfied with it ?
> 
> Do you mean the simplification of `get_all_records`? I think I am happy with it. I will have to have another look. What do you think?

yes that's what I mean. you can once everything is merged, it might be easier to see the whole result.

I'm working currently on the docstring upgrade + migration guide (that part is quite long to explain, new functions, removed methods etc)

alifeee:
my apologies for merging this, I have restored the branch and reverted the merge commit with 51f9f25cd198f1ca69f530fa0a4cf6750f808097.

You may need to create a new PR :/

PR #1378, opened 2024-01-19
feature/add utils get records

### LINKED ISSUE #976: `get_all_records` could accept already pulled values
## Overview

the method `get_all_records` could accept the values from the user, instead of pulling them.

## Details

If we break the method `get_all_records` in 2 parts, we can have a second method that can accept values from the user,
which will be in charge of validation of the headers and then call the `get_all_records` to perform the construction of the list of dictionaries.

For original idea see #973.

## Proposal

- break the method `get_all_records` in 2 parts
  1. pull all the values
  2. generate the list of dicts
- update the current method/add a new method that can accept already pulled values
- from there call the part 2 of the `get_all_records` that will generate the list of dicts from the values
  (values that would be given by the user *or* pulled by the function)

### LINKED ISSUE #1367: PROPOSAL: changes to `get_records`/`get_all_records`

### LINKED ISSUE #1376: add `get_records` changes to README migration guide

lavigne958:
- improve docstring - add small example
- Add migration guide for method `get_all_records`

closes #1376 
closes #1367 
closes #976

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,40 +88,19 @@ file.sheet1.update_tab_color(tab_color)
 + age = spreadsheet.get_lastUpdateTime()
 ```
 
-### Removed method `Worksheet.get_records`
+### Replace method `Worksheet.get_records`
 
-The method `Worksheet.get_records()` has been removed.
-The function `utils.to_records` has been added.
-
-It takes hearder row, a matrix values and build the records from them.
-
-### Simplify method `Worksheet.get_all_records`
-
-v6 simplified the method `Worksheet.get_all_records`, it does not accept a bounded ranges of rows for
-the values to fill the records. The below arguments have been remove from the method signature:
-
-- first_index
-- last_index
-
-It now uses the function `utils.to_records`.
-
-Update the method signature as follow:
+In v6 you can now only get *all* sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
 
 ```diff
-  get_all_records(
-      empty2zero=False,
-      head=1,
--     first_index=None,
--     last_index=None,
-      default_blank="",
-      allow_underscores_in_numeric_literals=False,
-      numericise_ignore=[],
-      value_render_option=None,
-      expected_headers=None,
-  )
++ from gspread import utils
+  all_records = spreadsheet.get_all_records(head=1)
+- some_records = spreadsheet.get_all_records(head=1, first_index=6, last_index=9)
+- some_records = spreadsheet.get_records(head=1, first_index=6, last_index=9)
++ header = spreadsheet.get("1:1")[0]
++ cells = spreadsheet.get("6:9")
++ some_records = utils.to_records(header, cells)
 ```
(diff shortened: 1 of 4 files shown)


lavigne958:
thought I could re-open the PR but no, I had to rebase it on tarrget branch then cherry-pick the original commit :wink: 

ready for review with:

- migration guide
- extra docstring with examples
- new function `utils.to_records`
- use of the new function for `get_all_records`

alifeee (approved):
this looks good to me :)

I briefened the migration guide to fit in with the rest of it.

PR #1381, opened 2024-01-25
feature/merge master into release 6 0 0

lavigne958:
- Correct the default value of docstring of merge_cells
- use namedtuple instead string of merge_type
- add test for merge_cells and unmerge_cells
- Release v5.12.4
- Add migration guide for get_all_records

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -52,6 +52,15 @@ class Dimension(StrEnum):
     cols = "COLUMNS"
 
 
+class MergeType(StrEnum):
+    merge_all = "MERGE_ALL"
+    merge_columns = "MERGE_COLUMNS"
+    merge_rows = "MERGE_ROWS"
+
+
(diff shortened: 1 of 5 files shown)


gspread/utils.py
```diff
@@ -52,6 +52,15 @@ class Dimension(StrEnum):
     cols = "COLUMNS"
 
 
+class MergeType(StrEnum):
+    merge_all = "MERGE_ALL"
+    merge_columns = "MERGE_COLUMNS"
+    merge_rows = "MERGE_ROWS"
+
+
+SILENCE_WARNINGS_ENV_KEY = "GSPREAD_SILENCE_WARNINGS"
```
alifeee, line 61: is this used in v6?

alifeee (approved):
looks good to me :)

gspread/utils.py
lavigne958: No we don't. But I thought we would keep it, for the next major release, this way we can use the exact same env variable

PR #1384, opened 2024-01-28
Allow client to use external Session object

### LINKED ISSUE #1383: Cannot use `session` with `Client()` in v6.0.0
Hey there, I am using this in a function and I think the recent changes here might be causing things to break (though I'm not 100% sure). @lavigne958 do you mind taking a look at https://github.com/burnash/gspread/pull/1159?

specifically: `def authorize(credentials: Credentials, client_factory: Type[Client] = Client):`

Here's a snippet of the error that i'm getting around this:

```bash
---> 47 client = Client(credentials,session)
     48 # auth
     50 if len(x) == 44:

File /local_disk0/.ephemeral_nfs/envs/pythonEnv-477f2a82-afcb-4603-8095-3ff1ba3c7ab7/lib/python3.9/site-packages/gspread/client.py:41, in Client.__init__(self, auth, http_client)
     38 def __init__(
     39     self, auth: Credentials, http_client: HTTPClientType = HTTPClient
     40 ) -> None:
---> 41     self.http_client = http_client(auth)

TypeError: 'AssertionSession' object is not callable
```

The original function is being used in a spark environment.

<details><summary>here is the code:</summary>

```python
import json
import gspread
import pandas as pd
import re
import time
from gspread import Client
from authlib.integrations.requests_client import AssertionSession
from oauth2client.service_account import ServiceAccountCredentials

def read_google_sheet(x, y, *args, **kwargs):
    # x --> Spreadsheet name OR Spreadsheet key
    # y --> Sheet name
    # args --> Name of Temp Table
    # kwargs --> cell_range / make_columns_unique / append_to_col_header / clean_column_names (optional)

    scope = ['https://spreadsheets.google.com/feeds',
             'https://www.googleapis.com/auth/drive']

    credentials = ServiceAccountCredentials.from_json_keyfile_dict(json_object)

    def create_assertion_session(scopes, subject=None):
        token_url = json_object['token_uri']
        issuer = json_object['client_email']
        key = json_object['private_key']
        key_id = json_object['private_key_id']
        header = {'alg': 'RS256'}

        if key_id:
            header['kid'] = key_id
        # Google puts scope in payload
        claims = {'scope': ' '.join(scopes)}

        return AssertionSession(
            grant_type=AssertionSession.JWT_BEARER_GRANT_TYPE,
            token_endpoint=token_url,
            issuer=issuer,
            audience=token_url,
            claims=claims,
            subject=subject,
            key=key,
            header=header,
        )

    session = create_assertion_session(scope)
    client = Client(credentials, session)

    if len(x) == 44:
        try:
            sht = client.open_by_key(x)
        except:
            sht = client.open(x)
    else:
        sht = client.open(x)

    # Added lines below 3.17.22
    ss = client.open_by_key(sht.id)

    if len(re.sub("[^0-9]", "", str(y))) == len(str(y)):
        ws = sht.get_worksheet_by_id(int(y))
    else:
        ws = sht.worksheet(y)

    # Added lines below 3.17.22
    if kwargs is not None and 'cell_range' in kwargs and len(re.sub("[^0-9]", "", str(y))) == len(str(y)):
        range_data = f'{ws.title}{kwargs["cell_range"]}'
    elif kwargs is not None and 'cell_range' in kwargs:
        range_data = str(kwargs['cell_range'])
    else:
        range_data = ws.title

    values = ss.values_get(range_data)
    values = values['values']
    pd_df = pd.DataFrame(values)
    pd_df.fillna("", inplace=True)  # convert None --> ''
    # clean up dataset

    headers = pd_df.iloc[0]
    headers = [x.lower() for x in headers]  # convert series to list
    # get column headers from 0 index (ie. first) row

    cols = []

    if kwargs is not None and 'make_columns_unique' in kwargs:
        for x in range(0, len(headers)):
            if headers[:x].count(headers[x]) >= 1:
                cols.append(f'{headers[x]}{headers[:x].count(headers[x])}')
            else:
                cols.append(str(headers[x]))
    else:
        cols = headers

    pd_df = pd.DataFrame(pd_df.values[1:], columns=cols)
    # make columns unique --> if duplicate columns, then append the column index to end of respective column name

    if kwargs is not None and 'append_to_col_header' in kwargs:
        pd_df.columns = [f'{kwargs["append_to_col_header"]}{col_name}' for col_name in pd_df.columns]
        cols = [x.lower() for x in pd_df.columns.values]
        # append kwargs['append_to_col_header'] to beginning of column name

    try:
        spark_df = spark.createDataFrame(pd_df)
    except:
        cleaned_cols = []
        for x in range(0, len(cols)):
            if cols[:x].count(cols[x]) >= 1:
                cleaned_cols.append(f'{cols[x]}{cols[:x].count(cols[x])}')
            else:
                cleaned_cols.append(str(cols[x]))

        pd_df = pd.DataFrame(pd_df.values[1:], columns=cleaned_cols)
        spark_df = spark.createDataFrame(pd_df)
        # if there are still duplicates, then map the column index to the end of the column name by default

    if kwargs is not None and 'clean_column_names' in kwargs:
        pandas_before = spark_df.toPandas()

        pandas_before.columns = [re.sub("[^a-zA-Z0-9]+", "_", x) for x in pandas_before.columns]
        headers = [re.sub('__', '_', re.sub('___', '_', x)) for x in pandas_before.columns.values]
        headers = [x[:len(x) - 3] if x.endswith('___', 0, len(x)) else x.lower() for x in headers]
        headers = [x[:len(x) - 2] if x.endswith('__', 0, len(x)) else x.lower() for x in headers]
        headers = [x[:len(x) - 1] if x.endswith('_', 0, len(x)) else x.lower() for x in headers]
        pandas_before.columns = headers

        cols = []
        headers = [x.lower() for x in pandas_before.columns.values]
        for x in range(0, len(headers)):
            if headers[:x].count(headers[x]) >= 1:
                cols.append(f'{headers[x]}{headers[:x].count(headers[x])}')
            else:
                cols.append(str(headers[x]))

        pandas_before.columns = cols
        pd_df = pandas_before

    if args:
        spark_df = spark.createDataFrame(pd_df)
        spark_df.createOrReplaceTempView(args[0])

    return spark_df, pd_df, sht, ws, ss
```

</details>

lavigne958:
Allow new `gspread.client.Client` to be initialized using an external `requests.Session` object.

closes #1383

### DIFF
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -15,6 +15,7 @@ from google.auth.credentials import Credentials
 from google.oauth2.credentials import Credentials as OAuthCredentials
 from google.oauth2.service_account import Credentials as SACredentials
 from google_auth_oauthlib.flow import InstalledAppFlow
+from requests import Session
 
 from .client import Client
 from .http_client import HTTPClient, HTTPClientType
@@ -54,7 +55,9 @@ DEFAULT_SERVICE_ACCOUNT_FILENAME = DEFAULT_CONFIG_DIR / "service_account.json"
 
 
 def authorize(
-    credentials: Credentials, http_client: HTTPClientType = HTTPClient
+    credentials: Credentials,
+    http_client: HTTPClientType = HTTPClient,
+    session: Optional[Session] = None,
 ) -> Client:
     """Login to Google API using OAuth2 credentials.
     This is a shortcut/helper function which
@@ -62,11 +65,19 @@ def authorize(
     By default :class:`gspread.HTTPClient` is used (but could also use
     :class:`gspread.BackOffHTTPClient` to avoid rate limiting).
 
+    It can take an additional `requests.Session` object in order to provide
+    you own session object.
+
+    .. note::
+
+       When providing your own `requests.Session` object,
+       use the value `None` as `credentials`.
+
     :returns: An instance of the class produced by `http_client`.
     :rtype: :class:`gspread.client.Client`
     """
 
-    return Client(auth=credentials, http_client=http_client)
+    return Client(auth=credentials, http_client=http_client, session=session)
 
 
 def local_server_flow(
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -10,7 +10,7 @@ from http import HTTPStatus
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from google.auth.credentials import Credentials
-from requests import Response
+from requests import Response, Session
 
 from .exceptions import APIError, SpreadsheetNotFound, UnSupportedExportFormat
 from .http_client import HTTPClient, HTTPClientType, ParamsType
@@ -36,9 +36,12 @@ class Client:
     """
 
     def __init__(
-        self, auth: Credentials, http_client: HTTPClientType = HTTPClient
+        self,
+        auth: Credentials,
+        http_client: HTTPClientType = HTTPClient,
+        session: Optional[Session] = None,
     ) -> None:
-        self.http_client = http_client(auth)
+        self.http_client = http_client(auth, session)
 
     def get_file_drive_metadata(self, id: str) -> Any:
         """Get the metadata from the Drive API for a specific file


alifeee:
I am not really clear on how the client works.

Is this something that was possible in v5 but was made impossible in v6?

Either way, nice fix.

alifeee (approved):
This is good.

Is this something we would've caught if we'd tried to use gspread for some common actions?

lavigne958:
> I am not really clear on how the client works.

The client is the gspread entry point, it required some credentials and the Type of http client you want to use (standard/BackOff).

> 
> Is this something that was possible in v5 but was made impossible in v6?

Yes it is 😑😞 that's a regression out of the box right here. 

In v5 it used to take the following arguments to instantiate a new client:
- credentials
-  session object. 

Now with the rework and new client it takes:
- credentials
- the Type of http client to use

So no more ways to pass you own session object 😬

> 
> Either way, nice fix.

Thanks man ! 🙃

lavigne958:
> This is good.
> 
> Is this something we would've caught if we'd tried to use gspread for some common actions?

I don't think so, this can be caught only by having a test with a custom session object. Otherwise you don't see that you miss something in your interface.

alifeee:
> The client is the gspread entry point, it required some credentials and the Type of http client you want to use (standard/BackOff).

ok I think I'm getting confused between "client" (gspread client) and "client" (http client)

> I don't think so, this can be caught only by having a test with a custom session object. Otherwise you don't see that you miss something in your interface.

Then we should also add a regression test...

lavigne958:
> > The client is the gspread entry point, it required some credentials and the Type of http client you want to use (standard/BackOff).
> 
> ok I think I'm getting confused between "client" (gspread client) and "client" (http client)

yes sorry ! that's a very good point ! we should use *client* for the gspread client and *http_client* for the actual class that handles all HTTP request (the `HTTPClient` or `BackOffHTTPClient`)

> 
> > I don't think so, this can be caught only by having a test with a custom session object. Otherwise you don't see that you miss something in your interface.
> 
> Then we should also add a regression test...

yes, I propose we add it later exceptionally if that's ok for you ? that would allow us to release first for the users

PR #1392, opened 2024-01-30
bugfix/client export

### LINKED ISSUE #1385: Client functions in v6.0.0 do not work - `HTTP Client Object has no attribute insert_permission`

---

**Describe the bug**
When we are sharing a spreadsheet there is an issue with Sharing since the HTTPClient doesn't have the insert_permission function.

**To Reproduce**
1. Create a Spreadsheet
2. Use the share function.

**Expected behavior**
We should be able to share stuff.

lavigne958:
- fix missing method `export` in http client
- Add test to export methods
- Fix missing method in HTTP Client

closes #1385

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 from google.auth.credentials import Credentials
 from requests import Response, Session
 
-from .exceptions import APIError, SpreadsheetNotFound, UnSupportedExportFormat
+from .exceptions import APIError, SpreadsheetNotFound
 from .http_client import HTTPClient, HTTPClientType, ParamsType
 from .spreadsheet import Spreadsheet
 from .urls import (
@@ -239,15 +239,7 @@ class Client:
         .. _ExportFormat: https://developers.google.com/drive/api/guides/ref-export-formats
         """
 
-        if format not in ExportFormat:
-            raise UnSupportedExportFormat
-
-        url = "{}/{}/export".format(DRIVE_FILES_API_V3_URL, file_id)
-
-        params: ParamsType = {"mimeType": format}
-
-        r = self.http_client.request("get", url, params=params)
-        return r.content
+        return self.http_client.export(file_id=file_id, format=format)
 
     def copy(
         self,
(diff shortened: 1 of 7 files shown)


alifeee:
nice change :)

you are doing a relative lot of PRs, I hope you are having enough time for your sanity :)

do we know for sure that these are all the "missing methods"? could there be others? I suppose you have manually checked each `client` method to see if it would "be missing" or not.

Also, I understand this is a "bandage PR" to get people's code working, but, the methods should not be in the `http_client`, right? They belong in the gspread `Client` object, do you think? It seems unreasonable for every custom client to implement all the methods `copy`, `share`, etc.

alifeee (approved):
I will approve this for the minor patch, but I think we should start an issue to track/discuss whether these methods should be in the `http_client` or not

lavigne958:
> nice change :)
> 
> you are doing a relative lot of PRs, I hope you are having enough time for your sanity :)

Ha ha thanks :wink:  I do so far so good.
> 
> do we know for sure that these are all the "missing methods"? could there be others? I suppose you have manually checked each `client` method to see if it would "be missing" or not.
> 

Yes I checked, I noticed we miss some typing in the `__init__` method of the `Spredsheet` object, I added the typing temporarily to identify quickly the missing methods, then removed it so we don't mix unrelated changes.

> Also, I understand this is a "bandage PR" to get people's code working, but, the methods should not be in the `http_client`, right? They belong in the gspread `Client` object, do you think? It seems unreasonable for every custom client to implement all the methods `copy`, `share`, etc.

I agree, this is temporary, though we might keep it until the next major release (could be much closer than the last one), unless we find a way to move it and keep people's code working (as usual :eyes: )

I opened the issue #1399 to discuss it.

lavigne958:
> I will approve this for the minor patch, but I think we should start an issue to track/discuss whether these methods should be in the `http_client` or not

agreed, as mentioned above I created the issue to discuss it.

alifeee:
> I agree, this is temporary, though we might keep it until the next major release (could be much closer than the last one)

my preference would be to make as few major releases as possible, ever. It is not fun to deal with the churn of software

PR #1398, opened 2024-01-31
check oauth creds type using `isinstance`

### LINKED ISSUE #1390: gspread 6.0.0 OAuth requires re-authentication every time.

Did this.  From what I can tell, I'm using it properly.
[Link to Stackoverflow posting here.
](https://stackoverflow.com/questions/77896761/using-gspread-with-oauth-and-have-to-re-authorize-every-time-i-re-run-my-code)

---

**Describe the bug**
I'm trying to use gspread to access a google spreadsheet and I've been able to successfully do so, however, I have to re-authorize every time I run the code.

**To Reproduce**

1. Run code below
2. Authenticate when asked
3. See expected results in google sheet
4. Run code again
5. **Expect prior authentication to work, but doesn't.  Have to re-authenticate**
6. See expected results in google sheet.

**Expected behavior**
Expect subsequent executions of script to not have to re-authenticate to access Google spreadsheet.

**Code example***
```
import gspread

SCOPES = ["https://www.googleapis.com/auth/spreadsheets"]
SPREADSHEET_ID = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"

CRED_FILE = "client_secret.json"
AUTH_USER_FILE = 'auth_user.json'

gc = gspread.oauth(credentials_filename=CRED_FILE, scopes=SCOPES, authorized_user_filename=AUTH_USER_FILE)

# Open a sheet from a spreadsheet in one go
spreadsheet = gc.open_by_key(SPREADSHEET_ID)

ws = spreadsheet.get_worksheet(0)

# Update a range of cells using the top left corner address
ws.update([[1, 2], [3, 4]], 'A1')

```

**Screenshots**
N/A

**Environment info:**
OS: Mint Linux 21.3
Python: 3.10
gspread: 6.0.0
IDE: PyCHarm 2023.3.2

**Additional context**
I've played a bit directly with the Python googleapi package and the authentication works as I would expect... i.e. authentication required the first time, but not required for subsequent runs of the script (within timeout period of authentication).

alifeee:
closes #1390

Credentials object is renamed `OAuthCredentials` in v6.0.0, and is no longer `Credentials` type, so the type-check for "already logged in" was failing. See https://github.com/burnash/gspread/issues/1390#issuecomment-1919494986 for more context.

It is worth noting that types were being checked with `type()`. This it not reccomended as per [pylint](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html).

If we had been using `isinstance`, this bug would not appear, as

`if not type(creds) is Credentials:` did not work with `OAuthCredentials`
`if not isinstance(creds, Credentials):` would have worked with `OAuthCredentials` as it is a subclass of `Credentials`

- [ ] I do not know enough about `oauth` to add an adequate test so that this problem does not happen again
- [ ] I do not know enough about `oauth` to know if this renaming of `Credentials` -> `OAuthCredentials` could have other problems in `auth.py` or elsewhere

Thus, the result of this change is only tested manually, by me, on my computer.... :/

### DIFF
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -189,7 +189,7 @@ def oauth(
     authorized_user_filename = Path(authorized_user_filename)
     creds = load_credentials(filename=authorized_user_filename)
 
-    if not isinstance(creds, OAuthCredentials):
+    if not isinstance(creds, Credentials):
         with open(credentials_filename) as json_file:
             client_config = json.load(json_file)
         creds = flow(client_config=client_config, scopes=scopes)
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -189,7 +189,7 @@ def oauth(
     authorized_user_filename = Path(authorized_user_filename)
     creds = load_credentials(filename=authorized_user_filename)
 
-    if not type(creds) is Credentials:
+    if not isinstance(creds, OAuthCredentials):
         with open(credentials_filename) as json_file:
             client_config = json.load(json_file)
         creds = flow(client_config=client_config, scopes=scopes)


gspread/auth.py
```diff
@@ -189,7 +189,7 @@ def oauth(
     authorized_user_filename = Path(authorized_user_filename)
     creds = load_credentials(filename=authorized_user_filename)
 
-    if not type(creds) is Credentials:
+    if not isinstance(creds, OAuthCredentials):
```
lavigne958, line 192: very good catch ! 

Here we should compare to the base class `Credentials` because the function `load_credentials` above can return an instance of type `google.auth.external_account_authorized_user.Credentials` or `google.oauth2.credentials.Credentials`

lavigne958 (commented):
very good catch ! I believe what we need is betwenn what you found and what was originally here:

- using `isinstance` instead of `type`
- using the generic base class `Credentials`

gspread/auth.py
alifeee: sure, makes sense.

I chose `OAuthCredentials` because this is under the `oauth()` method. But, as I said, I do not know oauth well, so I have changed.

lavigne958:
> closes #1390
> 
> Credentials object is renamed `OAuthCredentials` in v6.0.0, and is no longer `Credentials` type, so the type-check for "already logged in" was failing. See [#1390 (comment)](https://github.com/burnash/gspread/issues/1390#issuecomment-1919494986) for more context.
> 
> It is worth noting that types were being checked with `type()`. This it not reccomended as per [pylint](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html).
> 
> If we had been using `isinstance`, this bug would not appear, as
> 
> `if not type(creds) is Credentials:` did not work with `OAuthCredentials` `if not isinstance(creds, Credentials):` would have worked with `OAuthCredentials` as it is a subclass of `Credentials`
> 
> * [ ]  I do not know enough about `oauth` to add an adequate test so that this problem does not happen again

Unfortunately we can't add tests on oauth because it requires your to open a browser to allow the access.
There are possible way to make it work... but it would take us a considerable amount of time that can be used elsewhere I suppose. We must remind ourselves to manually test it from time to time or when changing anything around it !

> * [ ]  I do not know enough about `oauth` to know if this renaming of `Credentials` -> `OAuthCredentials` could have other problems in `auth.py` or elsewhere

as I recommended we should keep comparing to the base class so this won't be a problem. 

> 
> Thus, the result of this change is only tested manually, by me, on my computer.... :/

lavigne958 (approved):

PR #1402, opened 2024-02-01
Fix missing attribute `spreadsheet` in `Worksheet`.

### LINKED ISSUE #1395: gspread 6.0.0 no longer compatible with gspread-formatting

---

**Describe the bug**
gspread 6.0.0 no longer compatible with gspread-formatting. Specifically, `get_conditional_format_rules(worksheet)` raises an `AttributeError: 'Worksheet' object has no attribute 'spreadsheet'. Did you mean: 'spreadsheet_id'?`. I checked and this is because indeed, the `Worksheet` object no longer has the attribute `spreadsheet`, but only `spreadsheet_id` (as one might see when printing `worksheet.__dict__`).

**To Reproduce**
1. on gspread 6.0.0 create a spreadsheet
2. worksheet = spreadsheet.get_worksheet(0)
3. worksheet.spreadsheet will raise an error. This is important because the gspread-formatting function, get_conditional_format_rules, uses worksheet.spreadsheet.

**Expected behavior**
I would have expected worksheet to have the attribute `spreadsheet` and that will sort out the incompatibility with gspread-formatting.

**Code example***
If applicable, provide a code example to help explain your problem.
```
import gspread
from gspread_formatting import get_conditional_format_rules

credentials = Credentials.from_service_account_info(<token>, scopes=['https://www.googleapis.com/auth/spreadsheets',
                           'https://www.googleapis.com/auth/drive.file',
                           'https://www.googleapis.com/auth/drive',
                           'https://spreadsheets.google.com/feeds'])
client = gspread.authorize(credentials)
worksheet = client.create(title=<title>, folder_id=<folder_id>)

get_conditional_format_rules(worksheet
```

**Environment info:**
- Operating System macOS:
- Python version 3.10
- gspread version 6.0.0

lavigne958:
The `Worksheet` object was missing the `spreadsheet` attribute.

Add it back to be backward compatible.

closes #1395

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -9,6 +9,7 @@ This module contains common worksheets' models.
 import re
 import warnings
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     Dict,
@@ -58,6 +59,9 @@ from .utils import (
     rowcol_to_a1,
     to_records,
 )
+
+if TYPE_CHECKING is True:
+    from .spreadsheet import Spreadsheet
 
 CellFormat = TypedDict(
     "CellFormat",
@@ -160,7 +164,7 @@ class Worksheet:
         spreadsheet_id: str,
         client: HTTPClient,
         properties: MutableMapping[str, Any],
-        spreadsheet: Any,
+        spreadsheet: "Spreadsheet",
     ):
         self.spreadsheet_id = spreadsheet_id
         self.client = client
@@ -168,7 +172,7 @@ class Worksheet:
 
         # kept for backward compatibility - publicly available
         # do not use if possible.
-        self.spreadsheet = spreadsheet
+        self._spreadsheet = spreadsheet
 
     def __repr__(self) -> str:
         return "<{} {} id:{}>".format(
@@ -181,6 +185,11 @@ class Worksheet:
     def id(self) -> int:
         """Worksheet ID."""
         return self._properties["sheetId"]
+
+    @property
+    def spreadsheet(self) -> "Spreadsheet":
+        """Parent spreadsheet"""
+        return self._spreadsheet
 
     @property
     def title(self) -> str:
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -160,10 +160,15 @@ class Worksheet:
         spreadsheet_id: str,
         client: HTTPClient,
         properties: MutableMapping[str, Any],
+        spreadsheet: Any,
     ):
         self.spreadsheet_id = spreadsheet_id
         self.client = client
         self._properties = properties
+
+        # kept for backward compatibility - publicly available
+        # do not use if possible.
+        self.spreadsheet = spreadsheet
 
     def __repr__(self) -> str:
         return "<{} {} id:{}>".format(
@@ -2402,6 +2407,7 @@ class Worksheet:
         client: HTTPClient,
         spreadsheet_id: str,
         sheet_id: int,
+        spreadsheet: Any,
         insert_sheet_index: Optional[int] = None,
         new_sheet_id: Optional[int] = None,
         new_sheet_name: Optional[str] = None,
@@ -2444,7 +2450,7 @@ class Worksheet:
 
         properties = data["replies"][0]["duplicateSheet"]["properties"]
 
-        return Worksheet(spreadsheet_id, client, properties)
+        return Worksheet(spreadsheet_id, client, properties, spreadsheet)
 
     def duplicate(
         self,
@@ -2471,6 +2477,7 @@ class Worksheet:
             self.client,
             self.spreadsheet_id,
             self.id,
+            self.spreadsheet,
             insert_sheet_index=insert_sheet_index,
             new_sheet_id=new_sheet_id,
             new_sheet_name=new_sheet_name,
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -161,11 +161,34 @@ class Worksheet:
 
     def __init__(
         self,
-        spreadsheet_id: str,
-        client: HTTPClient,
-        properties: MutableMapping[str, Any],
         spreadsheet: "Spreadsheet",
+        properties: MutableMapping[str, Any],
+        spreadsheet_id: Optional[str] = None,
+        client: Optional[HTTPClient] = None,
     ):
+        # This object is not intended to be created manually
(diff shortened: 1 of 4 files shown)


gspread/worksheet.py
```diff
@@ -160,11 +160,16 @@ def __init__(
...
         properties: MutableMapping[str, Any],
+        spreadsheet: Any,
     ):
         self.spreadsheet_id = spreadsheet_id
         self.client = client
         self._properties = properties
 
+        # kept for backward compatibility - publicly available
+        # do not use if possible.
+        self.spreadsheet = spreadsheet
```
alifeee, line 171: we can make this a private attribute with
_spreadsheet

and then make a method spreadsheet with @property decorator, where we can emit warnings

?

gspread/worksheet.py
```diff
@@ -2402,6 +2407,7 @@ def _duplicate(
         client: HTTPClient,
         spreadsheet_id: str,
         sheet_id: int,
+        spreadsheet: Any,
```
alifeee, line 2410: why no typing?

alifeee (commented):
looks good.

it is worrying that we do not have to change any tests for this behaviour.

Perhaps we should add one?

gspread/worksheet.py
lavigne958: you're right that's better, this way we can add some deprecation warnings when needed.

gspread/worksheet.py
lavigne958: because we won't keep it in my mind (I won't let it go sorry :stuck_out_tongue: ) 
and typing it requires importing `Spreadsheet`, which already imports `Worksheet` which creates a circular dependency.

We can work around that using some conditional imports, I'll try it

lavigne958: I managed to type it correctly :upside_down_face:

lavigne958:
> looks good.
> 
> it is worrying that we do not have to change any tests for this behaviour.
> 
> Perhaps we should add one?

you're right, I added a safeGuard test to keep access to public attributes and check their types too.

alifeee (commented):
looking at this made me think of something. In [v5 -> v6](https://github.com/burnash/gspread/compare/v5.12.4..v6.0.0), the [`Worksheet` constructor is changed](https://github.com/burnash/gspread/compare/v5.12.4..v6.0.0#diff-a7e599f9e04b2813160aa087d952a836bf000e3ae5b88627203c139d03df4ce3L122-L127)

```diff
class Worksheet
  def __init__(
    self,
-   spreadsheet,
+   spreadsheet_id: str,
+   client: HTTPClient,
    properties
  ):
```

This is a breaking change. It should be documented in the migration guide, as people now have to construct Worksheets differently.

Alternatively, this PR reintroduces the `spreadsheet` argument to `Worksheet.__init__()`, so we could try to make the `__init__` signature the same as in v5, so it is not a breaking change.

This would probably look something like


```diff
class Worksheet
  def __init__(
    self,
    spreadsheet,
    properties,
+   client: HTTPClient = None,
  ):
+   self.spreadsheet_id = spreadsheet.id
+   if client is None:
+      # raise warning/throw error/make client HTTPClient
```

Then people could still use `gspread.Worksheet(spreadsheet, properties)` as in v5.

What do you think?

lavigne958:
> good job with the recursive type-checking! :)

Thanks 🙏

> 
> looking at this made me think of something. In [v5 -> v6](https://github.com/burnash/gspread/compare/v5.12.4..v6.0.0), the [`Worksheet` constructor is changed](https://github.com/burnash/gspread/compare/v5.12.4..v6.0.0#diff-a7e599f9e04b2813160aa087d952a836bf000e3ae5b88627203c139d03df4ce3L122-L127)
> 
> ```diff
> class Worksheet
>   def __init__(
>     self,
> -   spreadsheet,
> +   spreadsheet_id: str,
> +   client: HTTPClient,
>     properties
>   ):
> ```
> 
> This is a breaking change. It should be documented in the migration guide, as people now have to construct Worksheets differently.
> 
> Alternatively, this PR reintroduces the `spreadsheet` argument to `Worksheet.__init__()`, so we could try to make the `__init__` signature the same as in v5, so it is not a breaking change.
> 
> This would probably look something like
> 
> ```diff
> class Worksheet
>   def __init__(
>     self,
>     spreadsheet,
>     properties,
> +   client: HTTPClient = None,
>   ):
> +   self.spreadsheet_id = spreadsheet.id
> +   if client is None:
> +      # raise warning/throw error/make client HTTPClient
> ```
> 
> Then people could still use `gspread.Worksheet(spreadsheet, properties)` as in v5.
> 
> What do you think?

This should not be a problem as the only way to obtain a worksheet is through the spreadsheet object (using like `get_worksheet()`).
So this means users are not expected to instantiate the object themselves. So any changes here should be transparent and not a breaking change. 

That's how the project is designed at least, but you never know what users could do with it.

Then we must  raise an error like you propose if someone tried to instantiate it with no client, so even if we keep the same constructor signature their code will raise some error anyway so they'll need to passe the HTTP client anyway.

I understand the issue, though to me it's not an expected use case by gspread.

alifeee:
> This should not be a problem as the only way to obtain a worksheet is through the spreadsheet object

This may be the most common way, but the Worksheet object is publicly accessible, no? Any user can use it like, 

```python
import Worksheet from gspread.Worksheet
... 
my_custom_worksheet = Worksheet(...)
```

So changing the initialisor signature would break this.

> So this means users are not expected to instantiate the object themselves

you can never expect users to only do as expected ;)

So maybe we make the breaking change still, but put it in the migration guide?

lavigne958:
> > This should not be a problem as the only way to obtain a worksheet is through the spreadsheet object
> 
> This may be the most common way, but the Worksheet object is publicly accessible, no? Any user can use it like,
> 
> ```python
> import Worksheet from gspread.Worksheet
> ... 
> my_custom_worksheet = Worksheet(...)
> ```
> 
> So changing the initialisor signature would break this.
> 
> > So this means users are not expected to instantiate the object themselves
> 
> you can never expect users to only do as expected ;)
> 
> So maybe we make the breaking change still, but put it in the migration guide?

You're right, I will upgrade it so the signature matches + the new extra argument (the HTTP Client)

We need to raise in case the HTTP Client is not given at least users will have an explicit error message here. 

Can you add it to the migraine guide please ?

alifeee (approved):
this looks good I will write the migration guide very soon :)

PR #1404, opened 2024-02-01
Fixup get empty cell value is `None`

### LINKED ISSUE #1403: return empty cell to old default (None not "None")
see https://github.com/burnash/gspread/pull/1344#issuecomment-1921469828

@lavigne958 this is a small change and should be able to squeeze into 6.0.1 :)

lavigne958:
Fixup regression where getting a cell
on an empty cell should return a valid `Cell`
object with a value `None`.

closes #1403

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -316,10 +316,14 @@ class Worksheet:
                 value_render_option=value_render_option,
                 return_type=GridRangeType.ValueRange,
             )
-            try:
-                value = str(data[0][0])
-            except IndexError:
-                value = str(None)
+
+            # we force a return type to GridRangeType.ValueRange
+            # help typing tool to see it too :-)
+            if isinstance(data, ValueRange):
+                value = data.first()
+            else:
+                raise RuntimeError("returned data must be of type ValueRange")
+
         except KeyError:
             value = ""
 
@@ -2209,7 +2213,7 @@ class Worksheet:
             str_query = query
 
             def match(x: Cell) -> bool:
-                if case_sensitive:
+                if case_sensitive or x.value is None:
                     return x.value == str_query
                 else:
                     return x.value.casefold() == str_query.casefold()
--- a/tests/worksheet_test.py
+++ b/tests/worksheet_test.py
@@ -1441,6 +1441,13 @@ class WorksheetTest(GspreadTest):
         self.assertEqual(value_ranges[0].first(), "A1")
 
     @pytest.mark.vcr()
+    def test_cell_return_first(self):
+        cell = self.sheet.cell(1, 1)
+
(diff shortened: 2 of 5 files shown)


gspread/worksheet.py
```diff
@@ -316,10 +316,14 @@ def cell(
...
-                value = str(data[0][0])
-            except IndexError:
-                value = str(None)
+
+            # we force a return type to GridRangeType.ValueRange
+            # help typing tool to see it too :-)
+            if isinstance(data, ValueRange):
+                value = data.first()
+            else:
+                raise RuntimeError("returned data must be of type ValueRange")
```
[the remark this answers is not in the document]
lavigne958: You're right there might be a possibility here, I will check that, I'm sure it could work 🤔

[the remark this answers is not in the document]
lavigne958: I tried something.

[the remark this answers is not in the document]
lavigne958: and it does not work :man_facepalming:

gspread/worksheet.py
```diff
@@ -2209,7 +2213,7 @@ def _finder(
             str_query = query
 
             def match(x: Cell) -> bool:
-                if case_sensitive:
+                if case_sensitive or x.value is None:
```
alifeee, line 2216: what is this?

tests/worksheet_test.py
```diff
@@ -1440,6 +1440,13 @@ def test_batch_get(self):
         self.assertEqual(value_ranges[1].range, "Sheet1!B4:D4")
         self.assertEqual(value_ranges[0].first(), "A1")
 
+    @pytest.mark.vcr()
+    def test_cell_return_first(self):
+        cell = self.sheet.cell(1, 1)
+
+        self.assertIsInstance(cell, gspread.cell.Cell)
+        self.assertIsNone(cell.value)
```
alifeee, line 1448: looks good. does cell sometimes return `""` instead of `None`? Is that what those guys in the linked issue were saying?

alifeee (approved):

gspread/worksheet.py
lavigne958: This for typing.

Just below we have the string compare method: `x.value.casefold()` and mypy tells me we can't do that if value is ` None` of course. It would result in an exception like `TypeError: 'NoneType' object is not callabl`.

And this will prevent us from it

tests/worksheet_test.py
lavigne958: I don't remember cells returning `""` empty string. So far they return either a value or `None` (of course that goes without the use of default value etc).

And no that's not what they were speaking of, their issue is related to us returning the string `"None"` instead of the value `None`. 

But yes the test here is to ensure what they were saying if you get an empty cell you get a `None` value.

PR #1407, opened 2024-02-03
Fix type hints at find method in worksheet.py

deftfitf:
Hi, I changed the type hints for the `in_row` and `in_column` arguments of the find method in `worksheet.py` from `Optional[bool]` to `Optional[int]`. 

Could you check this later?


lavigne958 (approved):
Thank you for this contribution 🙏
That's a good catch

PR #1409, opened 2024-02-06
update migration guide

### LINKED ISSUE #1408: Add new `Worksheet` init arguments to migration guide
## Overview
Add the new `Worksheet` init signature to the migration guide.

## Details

The `Worksheet` class got upgraded a bit.
It now requires new arguments in order to be initialized. 

1. the spreadsheet_id of its parent spreadsheet
2. the HTTP Client required to make HTTP requests.

Add these changes to the migration guide.

alifeee:
closes #1408

- add `Worksheet.__init__` change
- hide "rarer" changes under `<details>` so users are not overwhelmed by changes, and can get to "more examples" easier

guide now looks like (see [on branch](https://github.com/burnash/gspread/tree/af9eb4a9ed81ad1d428c9a3b0835439af9c760a7/))

image

I have hidden some of the suggested migration changes, as I think they are less important. I think it is more important for people reading the README to get to the `More Examples` section, especially if they are new to the library.

Thanks. Let me know if you think we should change this at all. Or just push to the branch :)

After this we can release 6.0.1?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ As well, `values` can no longer be a list, and must be a 2D array.
 + file.sheet1.update(range_name="B2:C2", values=[["54", "55"]])
 ```
 
+### More
+
+<details><summary>See More Migration Guide</summary>
+
 ### Change colors from dictionary to text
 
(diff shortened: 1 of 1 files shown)


lavigne958 (approved):
That's a very good change and it provides a better user experience. 👌

lavigne958:
Agreed we can now release 6.0.1 I'll do it in a minute

PR #1412, opened 2024-02-13
Fixup gspread client init arguments

### LINKED ISSUE #1411: Reorder gspread client init arguments
## Overview
Re-order gspread client `init()` argument in order to be backward compatible.

it was before:
`def __init__(self, auth, session=None)`

it now:
`def __init__(self, auth, httpClientType=HTTPClient, Session=None)`

it should be:
`def __init__(self, auth, session=None, httpClientType=HTTPClient)`

This will allow anyone using the previous version of gspread to keep using it.

see: https://github.com/aiguofer/gspread-pandas/issues/100

lavigne958:
In order to be backward compatible re-order gspread client init
arguments, and assign a default value to the new argument.

closes #1411

### DIFF
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -77,7 +77,7 @@ def authorize(
     :rtype: :class:`gspread.client.Client`
     """
 
-    return Client(auth=credentials, http_client=http_client, session=session)
+    return Client(auth=credentials, session=session, http_client=http_client)
 
 
 class FlowCallable(Protocol):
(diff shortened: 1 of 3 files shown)


lavigne958:
I will make a new release 6.0.2 right after this PR is merged in order to unblock gspread-pandas

lavigne958:
yes exactly, that's the same problem.
yes it is, never expected users to create their own client, here it's a bit special as aiguofer used to contribute to this repository. Not everyone can do it or will take the time to do it.

PR #1415, opened 2024-02-16
Improve back-off client

### LINKED ISSUE #1268: Fix BackOffHTTPClient not returning error 403

### LINKED ISSUE #1326: May be able to exclude 403 Forbidden from Backoff HTTP codes
https://github.com/burnash/gspread/blob/0a2f084a192ba8038900891c2d1c441cfa10d611/gspread/client.py#L603

Greetings,

I don't see where in the google API docs it states that `403` is also used for exceeding the request limits, seems like they have a specific status code for this (`429`). I believe this status code can be safely removed from the list in the backoff client to allow for immediate exceptions in cases of access related issues.

lavigne958:
Catch usage limit errors from the Drive API. it returns the HTTP code 403 (access denied) on usage limits and access denied. Find the difference in both case and retry when it's usage limits exceeded.

Improve back-off client code.

Add pretty print for the APIerror Exception.

closes #1268
closes #1326

### DIFF
--- a/gspread/exceptions.py
+++ b/gspread/exceptions.py
@@ -6,7 +6,7 @@ Exceptions used in gspread.
 
 """
 
-from typing import Dict, Optional, Union
+from typing import Any, Dict, Mapping, Optional, Union
 
 from requests import Response
 
@@ -40,15 +40,12 @@ class APIError(GSpreadException):
     such as when we attempt to retrieve things that don't exist."""
 
     def __init__(self, response: Response):
-        super().__init__(self._extract_text(response))
+        super().__init__(self._extract_error(response))
         self.response: Response = response
+        self.error: Mapping[str, Any] = response.json()["error"]
+        self.code: int = self.error["code"]
 
-    def _extract_text(
-        self, response: Response
-    ) -> Union[Dict[str, Union[int, str]], str]:
-        return self._text_from_detail(response) or response.text
-
-    def _text_from_detail(
+    def _extract_error(
         self, response: Response
     ) -> Optional[Dict[str, Union[int, str]]]:
         try:
@@ -56,6 +53,14 @@ class APIError(GSpreadException):
             return dict(errors["error"])
         except (AttributeError, KeyError, ValueError):
             return None
+
+    def __str__(self) -> str:
+        return "{}: [{}]: {}".format(
+            self.__class__.__name__, self.code, self.error["message"]
+        )
+
+    def __repr__(self) -> str:
+        return self.__str__()
 
 
 class SpreadsheetNotFound(GSpreadException):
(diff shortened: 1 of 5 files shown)


alifeee:
these changes look good to me. two things:

- the test in #1268 needs to be un-skipped, or the issue is not closed
- we should add a test with this new functionality.

lavigne958:
> these changes look good to me. two things:
> 
> * the test in [Fix BackOffHTTPClient not returning error 403 #1268](https://github.com/burnash/gspread/issues/1268) needs to be un-skipped, or the issue is not closed
> * we should add a test with this new functionality.

I enabled the previously skipped test.

I am not sure how to test this new code, we have a test that tries to open a forbidden spreadsheet.
We can't make a test that reaches the limits of the quotas by sending many requests: as this will take forever to record the first time and everytime we update the cassettes. last time I ran this code I could not reach the limit of quota due to the backoff client :face_with_spiral_eyes:

alifeee:
Can we mock the requests object for just this one test? Instead of using cassettes, we could mock requests to return 419 (or the right HTTP #), and then test that the client works as expected (likely by setting the timeout very low so it only retries once or twice)

lavigne958:
> Can we mock the requests object for just this one test? Instead of using cassettes, we could mock requests to return 419 (or the right HTTP #), and then test that the client works as expected (likely by setting the timeout very low so it only retries once or twice)

I ran some tests and I can't figure it out, I can change requests before they are made but they still need to be made, I can filter/change responses before they are returned to us but requests still need to be made with a valid response.

I can't find a way to filter that request specifically and change it's return code, in a some _easy/simple_ ways.

it looks like a lot of work for a automatic test. 

is it ok for you if we keep it like this for now ?

alifeee (approved):
> is it ok for you if we keep it like this for now ?

yes :) I think people will probably not hit the rate limit enough that it is *that* important...

lavigne958:
> > is it ok for you if we keep it like this for now ?
> 
> yes :) I think people will probably not hit the rate limit enough that it is _that_ important...

thank you, I agree, I tried it myself... :laughing:  never reached it

PR #1417, opened 2024-02-19
Bugfix/add set timeout

### LINKED ISSUE #1393: Client has no attribute set_timeout
**Describe the bug**
Using gspread 6.0.0 I have a `'Client' object has no attribute 'set_timeout'` error.

**To Reproduce**
```
        import gspread
        gspread_service = gspread.service_account(GSHEET_SERVICE_ACCOUNT_PATH)
        gspread_service.set_timeout(120)
```

**Expected behavior**
Update the timeout

**Screenshots**
image

**Environment info:**
- Operating System **Linux**
- Python version **3.11**
- gspread version **6.0.0**

lavigne958:
Add missing method to set the HTTP Timeout.

it got lost when moving HTTP requests to new HTTP Client.

Now the timeout can be:
- set using gspread client (like before, backward compatible)
- from the spreadsheet object: `spreadsheet.set_timeout()`
- from the sheet object: `sheet.set_timeout()`
- from the HTTP client anywhere it can be accessed:
  - from gspread service: `gs_client.http_client.set_timeout()`
  - from the spreadsheet: `spreadsheet.client.set_timeout()`
  - from the sheet: `sheet.client.set_timeout()`

closes #1393

### DIFF
--- a/gspread/http_client.py
+++ b/gspread/http_client.py
@@ -6,6 +6,7 @@ This module contains HTTPClient class responsible for communicating with
 Google API.
 
 """
+
 import time
 from http import HTTPStatus
 from typing import (
@@ -88,10 +89,12 @@ class HTTPClient:
 
         self.session.headers.update({"Authorization": "Bearer %s" % self.auth.token})
 
-    def set_timeout(self, timeout: Union[float, Tuple[float, float]]) -> None:
+    def set_timeout(self, timeout: Optional[Union[float, Tuple[float, float]]]) -> None:
         """How long to wait for the server to send
         data before giving up, as a float, or a ``(connect timeout,
         read timeout)`` tuple.
+
+        Use value ``None`` to restore default timeout
 
         Value for ``timeout`` is in seconds (s).
         """
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -7,9 +7,10 @@ This module contains common spreadsheets' models.
 """
 
 import warnings
-from typing import Union
+from typing import Any, Dict, Union
 
 from .exceptions import WorksheetNotFound
+from .http_client import HTTPClient
 from .urls import DRIVE_FILES_API_V3_URL, SPREADSHEET_DRIVE_URL
 from .utils import ExportFormat, finditem
 from .worksheet import Worksheet
@@ -18,7 +19,7 @@ from .worksheet import Worksheet
 class Spreadsheet:
     """The class that represents a spreadsheet."""
 
-    def __init__(self, http_client, properties):
+    def __init__(self, http_client: HTTPClient, properties: Dict[str, Any]):
         self.client = http_client
         self._properties = properties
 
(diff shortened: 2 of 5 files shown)


gspread/http_client.py
```diff
@@ -71,14 +72,19 @@ class HTTPClient:
     It will be created by the gspread.Client class.
     """
 
-    def __init__(self, auth: Credentials, session: Optional[Session] = None) -> None:
+    def __init__(
+        self,
+        auth: Credentials,
+        session: Optional[Session] = None,
+        timeout: Optional[Union[float, Tuple[float, float]]] = None,
```
alifeee, line 79: why a new init argument?

alifeee (changes requested):
looks good

was this method in Worksheet and Spreadsheet before? if not, why do we add it? it is not needed for backwards compatibility

The tests should be updated to reflect this change. We added a test for the missing client methods if I remember

lavigne958:
> looks good
> 
> was this method in Worksheet and Spreadsheet before? if not, why do we add it? it is not needed for backwards compatibility
not it was not, I thought I add extra ways to set timeout for users so it offer more entry point to set the timeout.
agreed it's not needed for backward compatibility.

> 
> The tests should be updated to reflect this change. We added a test for the missing client methods if I remember

I will check if I can add that, may be a new test that set a low timeout, so one day if we move it the test will fail when trying to call a method that does not exists, do you think this will be enough ? :thinking:

lavigne958:
Actually after taking a look I managed to add a test that _tests_ the `set_timeout` method and measures the time too in case we don't respect the timeout value. :partying_face:

lavigne958:
I was thinking, maybe it's better if I just add the missing method and nothing else. We'll see in the future to add other convenient methods ? It makes things simpler

lavigne958:
due to conflicts, rebased on master branch, simplified the changes by only adding the necessary function to be backward compatible

gspread/spreadsheet.py
```diff
@@ -18,7 +19,7 @@
 class Spreadsheet:
     """The class that represents a spreadsheet."""
 
-    def __init__(self, http_client, properties):
+    def __init__(self, http_client: HTTPClient, properties: Dict[str, Any]):
```
alifeee, line 22: why is this not caught by `mypy`?

alifeee (approved):

gspread/spreadsheet.py
lavigne958: I believe this is because we don't use mypy with the `--strict` mode. If a function is not typed then it's ok.

lavigne958: I wouldn't recommend it. The strict mode is very strict! It will take us forever to make it pass 🫣

We can do a second pass on the project and type what is missing, may be type de tests too.

After that we could run in strict mode and see what error rises and decide if we want to fix them.

alifeee: then I think for now we can leave out strict mode :)

lavigne958: yep, I created this issue to track: adding missing types in code + tests: #1430

PR #1426, opened 2024-02-27
Add missing method `import_csv()`

### LINKED ISSUE #1423: AttributeError: 'HTTPClient' object has no attribute 'import_csv'
I have been using gspread 5.12.4 with the following code without issue for some time.

I updated to 6.0.2 today and I am now receiving this message

AttributeError: 'HTTPClient' object has no attribute 'import_csv'

Rolling back to 5.12.4 (with no other changes), results in the csv uploading as expected.

dir(sh) gives (no import_csv) listed, although on 5.12.4 i do see "import_csv"

__class__
__delattr__
__dict__
__dir__
__doc__
__eq__
__format__
__ge__
__getattribute__
__gt__
__hash__
__init__
__init_subclass__
__le__
__lt__
__module__
__ne__
__new__
__reduce__
__reduce_ex__
__repr__
__setattr__
__sizeof__
__str__
__subclasshook__
__weakref__
auth
batch_update
export
fetch_sheet_metadata
get_file_drive_metadata
insert_permission
list_permissions
login
remove_permission
request
session
set_timeout
spreadsheets_get
spreadsheets_sheets_copy_to
timeout
values_append
values_batch_clear
values_batch_get
values_batch_update
values_clear
values_get
values_update

**To Reproduce**
try to call the client.import_csv() function

result:
AttributeError: 'HTTPClient' object has no attribute 'import_csv'

**Expected behavior**
the import_csv function is available

**Code example***

```
with open(secretFile, 'r') as f:
     secretFileJson = f.read()
serviceAcctDict     = json.loads(secretFileJson)
gc 			= gspread.service_account_from_dict(serviceAcctDict)
sh 			= gc.open('testTemps')
with open('mycsv.csv', 'r') as f:
    content = f.read()
sh.client.import_csv(sh.id, data=content)
```

**Screenshots**

**Environment info:**
- Operating System [e.g. Linux, Windows, macOS]:  Linux Mint, RaspberryOS
- Python version: Python 3.10.12 and Python 3.9.2
- gspread version 6.0.2 (has problem) 5.12.4 (is ok)

**Additional context**

lavigne958:
Prior to v6.x.y the method could be used as follow:

file = client.open('x')
file.client.import_csv()

Now file.client is of type HTTPClient
so we need to add the method back for backward compatibility

closes #1423

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -58,9 +58,6 @@ class MergeType(StrEnum):
     merge_rows = "MERGE_ROWS"
 
 
-SILENCE_WARNINGS_ENV_KEY = "GSPREAD_SILENCE_WARNINGS"
-
-
 class ValueRenderOption(StrEnum):
     formatted = "FORMATTED_VALUE"
     unformatted = "UNFORMATTED_VALUE"
@@ -270,13 +267,15 @@ def numericise_all(
     ignore = ignore or []
 
     numericised_list = [
-        values[index]
-        if index + 1 in ignore
-        else numericise(
-            values[index],
-            empty2zero=empty2zero,
-            default_blank=default_blank,
-            allow_underscores_in_numeric_literals=allow_underscores_in_numeric_literals,
+        (
+            values[index]
+            if index + 1 in ignore
+            else numericise(
+                values[index],
+                empty2zero=empty2zero,
+                default_blank=default_blank,
+                allow_underscores_in_numeric_literals=allow_underscores_in_numeric_literals,
+            )
         )
         for index in range(len(values))
     ]
@@ -917,6 +916,14 @@ def to_records(
 
 
 # SHOULD NOT BE NEEDED UNTIL NEXT MAJOR VERSION
+# DEPRECATION_WARNING_TEMPLATE = (
+#     "[Deprecated][in version {v_deprecated}]: {msg_deprecated}"
+# )
+
+
+# SILENCE_WARNINGS_ENV_KEY = "GSPREAD_SILENCE_WARNINGS"
+
+
 # def deprecation_warning(version: str, msg: str) -> None:
 #     """Emit a deprecation warning.
 
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -15,11 +15,7 @@ from requests import Response, Session
 from .exceptions import APIError, SpreadsheetNotFound
 from .http_client import HTTPClient, HTTPClientType, ParamsType
 from .spreadsheet import Spreadsheet
-from .urls import (
-    DRIVE_FILES_API_V3_COMMENTS_URL,
-    DRIVE_FILES_API_V3_URL,
-    DRIVE_FILES_UPLOAD_API_V2_URL,
-)
+from .urls import DRIVE_FILES_API_V3_COMMENTS_URL, DRIVE_FILES_API_V3_URL
 from .utils import ExportFormat, MimeType, extract_id_from_url, finditem
 
 
@@ -353,7 +349,7 @@ class Client:
         params: ParamsType = {"supportsAllDrives": True}
         self.http_client.request("delete", url, params=params)
 
-    def import_csv(self, file_id: str, data: Union[str, bytes]) -> None:
+    def import_csv(self, file_id: str, data: Union[str, bytes]) -> Any:
         """Imports data into the first page of the spreadsheet.
 
         :param str file_id:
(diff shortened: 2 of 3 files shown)


gspread/utils.py
```diff
@@ -270,13 +267,15 @@ def numericise_all(
...
-            allow_underscores_in_numeric_literals=allow_underscores_in_numeric_literals,
+        (
+            values[index]
+            if index + 1 in ignore
+            else numericise(
+                values[index],
+                empty2zero=empty2zero,
+                default_blank=default_blank,
+                allow_underscores_in_numeric_literals=allow_underscores_in_numeric_literals,
+            )
```
lavigne958, line 278: this has been auto-formated by my editor

gspread/utils.py
```diff
@@ -917,6 +916,14 @@ def to_records(
...
 
 # SHOULD NOT BE NEEDED UNTIL NEXT MAJOR VERSION
+# DEPRECATION_WARNING_TEMPLATE = (
+#     "[Deprecated][in version {v_deprecated}]: {msg_deprecated}"
+# )
+
+
+# SILENCE_WARNINGS_ENV_KEY = "GSPREAD_SILENCE_WARNINGS"
+
+
```
lavigne958, line 926: we lost these, I add them back for later

alifeee (commented):
This looks good. I only note:

I don't think `import_csv` makes sense within `HTTP_Client`? This method provides spreadsheet functionality, so it should be in `gspread.Client`?

I am not fully up to date with the idea of the `HTTP_Client`, but my understanding is that it was made so that people could make their own custom `HTTP_Client`s easier. Then, they would have to make their own `import_csv` method? Should this not belong to `Client`?

I see that it might *have* to be in `HTTP_Client`, because of hierarchical reasons. If this is the case, then I guess we must leave it.

lavigne958:
> This looks good. I only note:
> 
> I don't think `import_csv` makes sense within `HTTP_Client`? This method provides spreadsheet functionality, so it should be in `gspread.Client`?
> 
> I am not fully up to date with the idea of the `HTTP_Client`, but my understanding is that it was made so that people could make their own custom `HTTP_Client`s easier. Then, they would have to make their own `import_csv` method? Should this not belong to `Client`?
> 
> I see that it might _have_ to be in `HTTP_Client`, because of hierarchical reasons. If this is the case, then I guess we must leave it.


I know :disappointed: I faced the same thought, I wish it could only be located in the client itself and only be used like: `gs_client.import_csv()` which returns the new spreadsheet instance.

I agree, the HTTPClient allows users to make their own if necessary, and mostly it is here to handle only pure HTTP calls, on the least amount of *logic*.

The way the project was designed before, allowed users to reach the argument `client` in a spreadsheet and call methods on it, like this one. now that `client` argument is the HTTPClient so some logic has been removed.

what we can do is:
- leave the function only in the gspread client, like it is now
- do not put it back in the HTTPClient

which is in the end: make no changes, as currently the documentation of the function shows that you should use that function using the gspread client directly, it never shows that you can reach it from elsewhere.

or add it back to the HTTPClient so users can do: `spreadsheet.client.import_csv()` but we put back the deprecation warnings, in order to be very clean. and we live with it until next major release :thinking:

alifeee:
> or add it back to the HTTPClient so users can do: spreadsheet.client.import_csv() but we put back the deprecation warnings, in order to be very clean. and we live with it until next major release 🤔

This is a good question. I er towards adding it to the HTTPClient. As, in the end, if someone wanted to make their own, they would probably create a sub-class of the HTTPClient, and override the request methods. So the import_csv etc methods would remain.

Maximum backwards compatibility :)

lavigne958:
> > or add it back to the HTTPClient so users can do: spreadsheet.client.import_csv() but we put back the deprecation warnings, in order to be very clean. and we live with it until next major release 🤔
> 
> This is a good question. I er towards adding it to the HTTPClient. As, in the end, if someone wanted to make their own, they would probably create a sub-class of the HTTPClient, and override the request methods. So the import_csv etc methods would remain.
> 
> Maximum backwards compatibility :)

agreed, we can later depreciate it, like before for the v6.0.0, but this time we should have way less methods to depreciate ! so it should be much easier.

lavigne958:
the current pull request will fix it.
The original issue mentions: `AttributeError: 'HTTPClient' object has no attribute 'import_csv'`

and the pull request adds a new method `import_csv` to the `HTTPClient`.

I am waiting for your review/approval mostly now that we agree on leaving the new method. :upside_down_face: 

Unless: I did not understand and we still need to find a solution, which is fine.

alifeee (approved):

PR #1427, opened 2024-02-27
Fix wrapper `cast_to_a1_notation`

### LINKED ISSUE #1055: cast_to_a1_notation(method)   bugs?
cast_to_a1_notation(method) I believe there are bugs here

rowcol_to_a1()  does not want to take a list as in *args[:2]  but individual values

lavigne958:
fix the wrapper method cast_to_a1_notation.

It must check that the 4 first arguments are actual ints.

Then it should only try to cast the arguments at position 3 and 4 only, not the rest of the list.

closes #1055


alifeee:
If there were bugs before, and this fixes the bugs. We should have a test that proves it. :)

lavigne958:
> If there were bugs before, and this fixes the bugs. We should have a test that proves it. :)

You're right I'll add a test. I'll add a call to the method `define_named_range` with both a1 and index based coordinates. 

The detail is: it's a bug only on methods with no kwargs and multiple args with no default values. So its quiet specific.

lavigne958:
> > You're right I'll add a test. I'll add a call to the method define_named_range with both a1 and index based coordinates.
> 
> I am confused about the test. You added a single test that calls `self.sheet.define_named_range(1, 1, 2, 2, range_name)`. This uses index-based coordinates.
> 
> Where is the "a1 [...] based coordinate"?

I noticed we already have a test that calls this method with the _a1_ notation. So I added the one with coordinates. 

It makes more sense right to have them both in the test, I'll add it.

alifeee (approved):

lavigne958:
I looked around and found just above a test that adds a named range with A1 coordinates, so I merged the 2 tests into a single one that then deletes the named range then creates the exact same range with index based coordinates. and ensures everything is created the correct way.

PR #1428, opened 2024-02-28
Add new auth method API key

### LINKED ISSUE #1096: open a public sheet by using an API key instead of an oauth
- _**Is your feature request related to a problem? Please describe.**_
   - Accessing a sheet which is designated as public data with Oauth causes rate limit due to multiple users. 
Using an API key directly would allow more updates per second to get the sheet values

---
- **_Describe the solution you'd like_**
    - A new function to pull a sheets values similar to ```open_by_key(sheet_id)``` but allowing an option to pass an API key instead of the oauth tokens
---
- _**Describe alternatives you've considered**_
   - as far as alternatives I've considered.. would be another library but I would like to avoid that 
---
- _**Additional context**_
    - Add any other context or screenshots about the feature request here. [Link to other](https://github.com/xflr6/gsheets/blob/24e99f67e5f6bc49facc85f5db305caa63ccbdef/gsheets/api.py#L38)

lavigne958:
Add a new auth method that allows users to open public spreadsheet files with a simple API key.

closes #1096

### DIFF
--- a/docs/oauth2.rst
+++ b/docs/oauth2.rst
@@ -5,6 +5,7 @@ To access spreadsheets via Google Sheets API you need to authenticate and author
 
 * If you plan to access spreadsheets on behalf of a bot account use :ref:`Service Account <service-account>`.
 * If you'd like to access spreadsheets on behalf of end users (including yourself) use :ref:`OAuth Client ID <oauth-client-id>`.
+* If you'd like to **only** open public spreadsheets use :ref:`API key <api-key>`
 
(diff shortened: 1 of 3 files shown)


lavigne958:
> This looks good! A nice feature :)

Thanks 


> 
> Where is the test?

We don't have any test on authentication method.
We use the service account method to run the tests, but we don't have any on others as this requires constant internet connection to actually authenticate.

alifeee (approved):

PR #1431, opened 2024-03-08
update readme examples

### LINKED ISSUE #1410: Update examples of gspread usage /Readme.md
Hi, 

Their is no usage example of single cell update in the Read me as the example currently is not complete . 
simple add worksheet.update_acell('A1', 'Bingo!') to the updating cell example

alifeee:
closes #1410 

The examples in the readme were wrong, and did not work

I think we should go through and test them to see if they work. I will do this soon if you do not. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,10 +261,10 @@ cell_list = worksheet.findall(criteria_re)
 
 ```python
 # Update a single cell
-worksheet.update('B1', 'Bingo!')
+worksheet.update_acell('Bingo!', 'B1')
 
 # Update a range
-worksheet.update('A1:B2', [[1, 2], [3, 4]])
+worksheet.update([[1, 2], [3, 4]], 'A1:B2')
 
 # Update multiple ranges at once
 worksheet.batch_update([{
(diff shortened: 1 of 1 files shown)


lavigne958:
> closes #1410
> 
> The examples in the readme were wrong, and did not work
> 
> I think we should go through and test them to see if they work. I will do this soon if you do not. :)

I will test them now

lavigne958:
The below example don't work either:

- https://github.com/burnash/gspread/blob/20005407e46c9c637ebb1ad6c5345e82d27593f1/README.md?plain=1#L264
   we should swap the value and the range too, and put the value into a matrix `[[".."]]`

apart from that the rest works fine :+1:

alifeee:
re: https://github.com/burnash/gspread/issues/1410#issuecomment-1986206623, I may have got the examples here the wrong way round. not sure. have to test.

lavigne958 (approved):

lavigne958:
> re: [#1410 (comment)](https://github.com/burnash/gspread/issues/1410#issuecomment-1986206623), I may have got the examples here the wrong way round. not sure. have to test.

looks good, I checked it's the right way around now.

PR #1435, opened 2024-03-13
fix error message readability

imrehg:
The multi-line string in the code concatenates imperfectly when actually printed. Need to include spacing manually, unless one gets

> the header row in the worksheet is not uniquetry passing [...snip...]


burnash:
Hey @imrehg, thanks for spotting this and for the fix!

PR #1439, opened 2024-03-15
Allow "warning" type protected ranges

### LINKED ISSUE #1437: worksheet.add_protected_range() - impossible to create the `warning_only=True` range
Hi, I wonder if there's a bug.

When using `worksheet.add_protected_range(name="A1:B2", warning_only=True)`, I'm getting the `TypeError: Worksheet.add_protected_range() missing 1 required positional argument: 'editor_users_emails'`

However, adding `editor_users_emails` triggers a 400 Bad Request, as the library seems to always require it:

```
worksheet.add_protected_range(name="A1:B2", warning_only=True, editor_users_emails="me@example.com")

--- (and the error is)

ProtectedRange is warningOnly. Editors cannot be set on it.
```

I think the gspread library must omit the editor node altogether from the request if `warning_only=True`.

alifeee:
closes #1437

They were not allowed before because:

It seems that when `add_protected_range` was added it was assumed that `editor_users_emails` was always required. However, warning ranges do not require (and in fact, cannot contain) emails.

To fix this issue, we:

- change `editor_users_emails` from and `argument` to a `keyword argument` (if it is not changed in position, it should still be usable as an argument)
- add some logic in the code to not pass "editors" if `warning_only` is set

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2037,7 +2037,7 @@ class Worksheet:
     def add_protected_range(
         self,
         name: str,
-        editor_users_emails: Sequence[str],
+        editor_users_emails: Sequence[str] = [],
         editor_groups_emails: Sequence[str] = [],
         description: Optional[str] = None,
         warning_only: bool = False,
@@ -2091,7 +2091,7 @@ class Worksheet:
                             "description": description,
                             "warningOnly": warning_only,
                             "requestingUserCanEdit": requesting_user_can_edit,
-                            "editors": {
+                            "editors": None if warning_only else {
                                 "users": editor_users_emails,
                                 "groups": editor_groups_emails,
                             },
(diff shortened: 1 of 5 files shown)


lavigne958 (approved):
Great change ! Oddly enough the API allows a user to set a protected range with no editors (when providing an actual empty list of editors). 🤔

alifeee:
yes I wasn't sure what email to put in the test, since it would be different for whoever ran the test.

maybe the email is fetchable from an API somewhere?

lavigne958:
> yes I wasn't sure what email to put in the test, since it would be different for whoever ran the test.
> 
> maybe the email is fetchable from an API somewhere?

I don't think it's a problem. Leaving it like this endures it works with no emails too 🙄

PR #1444, opened 2024-03-21
Added 'add_data_validation` to `Workhsheet` [Issue #1420]

### LINKED ISSUE #1420: A way to add DataValidation
The API allows to setup data validation on a range via the `batchUpdate` end point: https://developers.google.com/sheets/api/samples/data#data-validation

Since it is part of the `batchUpdate` endpoint, anyone can use the low-level method to pass through the data. However, there should be a high-level method. or a kwarg in the low-level [`values_update`](https://github.com/burnash/gspread/blob/4a48fbdc4bc56724d3a1d8fcdb7f5793fb58f160/gspread/http_client.py#L138) method.  We should create a class `DataValidation` class, which will have all the options provided by the API.

It would have `to_json` method, that will be passed to the update payload when sending the request.

Passing a `kwarg` to the update methods to pass through json data.

muddi900:
Added a RealtiveDate enum

removed the realtive date enum

Closes #1420

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ file.sheet1.update_tab_color(tab_color)
 
 ### Replace method `Worksheet.get_records`
 
-In v6 you can now only get _all_ sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
+In v6 you can now only get *all* sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
 
 ```diff
 + from gspread import utils
@@ -278,7 +278,7 @@ worksheet.batch_update([{
 
 ### Get unformatted cell value or formula
 
-````python
+```python
 from gspread.utils import ValueRenderOption
 
 # Get formatted cell value as displayed in the UI
@@ -316,7 +316,7 @@ worksheet.add_validation(
    showCustomUi=True
 )
 
-````
+```
 
 ## Documentation
 
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ file.sheet1.update_tab_color(tab_color)
 
 ### Replace method `Worksheet.get_records`
 
-In v6 you can now only get *all* sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
+In v6 you can now only get _all_ sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
 
 ```diff
 + from gspread import utils
@@ -278,7 +278,7 @@ worksheet.batch_update([{
 
 ### Get unformatted cell value or formula
 
-```python
+````python
 from gspread.utils import ValueRenderOption
 
 # Get formatted cell value as displayed in the UI
@@ -292,7 +292,31 @@ from gspread.utils import ValueRenderOption
 # Get formula from a cell
 >>> worksheet.get("C2:D2", value_render_option=ValueRenderOption.formula)
 [['=1/1024']]
-```
+### Add data validation to a range
+
+```python
+import gspread
+from gspread.utils import ValidationConditionType
+
+# Restrict the input to greater than 10 in a single cell
+worksheet.add_validation(
+  'A1',
+  ValidationConditionType.number_greater,
+  [10],
+  strict=True,
+  inputMessage='Value must be greater than 10',
+)
+
+# Restrict the input to Yes/No for a specific range with dropdown
+worksheet.add_validation(
+  'C2:C7',
+   ValidationConditionType.one_of_list,
+   ['Yes',
+   'No',]
+   showCustomUi=True
+)
+
+````
 
 ## Documentation
 
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -118,6 +118,41 @@ class GridRangeType(StrEnum):
     ListOfLists = "ListOfLists"
 
 
+class ValidationConditionType(StrEnum):
+    number_greater = "NUMBER_GREATER"
+    number_greater_than_eq = "NUMBER_GREATER_THAN_EQ"
+    number_less = "NUMBER_LESS"
+    number_less_than_eq = "NUMBER_LESS_THAN_EQ"
+    number_eq = "NUMBER_EQ"
+    number_not_eq = "NUMBER_NOT_EQ"
+    number_between = "NUMBER_BETWEEN"
+    number_not_between = "NUMBER_NOT_BETWEEN"
+    text_contains = "TEXT_CONTAINS"
+    text_not_contains = "TEXT_NOT_CONTAINS"
+    text_starts_with = "TEXT_STARTS_WITH"
+    text_ends_with = "TEXT_ENDS_WITH"
+    text_eq = "TEXT_EQ"
+    text_is_email = "TEXT_IS_EMAIL"
+    text_is_url = "TEXT_IS_URL"
+    date_eq = "DATE_EQ"
+    date_before = "DATE_BEFORE"
+    date_after = "DATE_AFTER"
+    date_on_or_before = "DATE_ON_OR_BEFORE"
+    date_on_or_after = "DATE_ON_OR_AFTER"
+    date_between = "DATE_BETWEEN"
+    date_not_between = "DATE_NOT_BETWEEN"
+    date_is_valid = "DATE_IS_VALID"
+    one_of_range = "ONE_OF_RANGE"
+    one_of_list = "ONE_OF_LIST"
+    blank = "BLANK"
+    not_blank = "NOT_BLANK"
+    custom_formula = "CUSTOM_FORMULA"
+    boolean = "BOOLEAN"
+    text_not_eq = "TEXT_NOT_EQ"
+    date_not_eq = "DATE_NOT_EQ"
+    filter_expression = "FILTER_EXPRESSION"
+
+
 def convert_credentials(credentials: Credentials) -> Credentials:
     module = credentials.__module__
     cls = credentials.__class__.__name__
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -3246,7 +3246,7 @@ class Worksheet:
             ws.add_validation(
                 'A1',
                 ValidationConditionType.number_greater,
-                10,
+                [10],
                 strict=True,
                 inputMessage='Value must be greater than 10',
             )
@@ -3254,8 +3254,7 @@ class Worksheet:
             ws.add_validation(
                 'C2:C7',
                 ValidationConditionType.one_of_list,
-                'Yes',
-                'No',
+                ['Yes','No'],
                 showCustomUi=True
             )
         """
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -41,6 +41,7 @@ from .utils import (
     PasteOrientation,
     PasteType,
     T,
+    ValidationConditionType,
     ValueInputOption,
     ValueRenderOption,
     a1_range_to_grid_range,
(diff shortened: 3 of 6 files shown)


alifeee:
hi :) thanks for the PR

it looks like you have to run `tox -e format` locally to format the code, so the linter is happy.

after that, the CI should pass, and we will be able to review this change :)

alifeee:
Hi, that looks good. Sorry there keep being issues.

Try deleting the cassette, and re-running the test to generate it again. Sometimes there are conflicts.

gspread/worksheet.py
```diff
@@ -3201,3 +3202,51 @@ def cut_range(
         }
 
         return self.client.batch_update(self.spreadsheet_id, body)
+
+    def add_validation(
+        self,
+        source: str,
+        condition_type: ValidationConditionType,
+        *values: Iterable[Any],
+        **kwargs,
```
lavigne958, line 3211: Please remove the `kwargs` and use explicit arguments: `inputMessage`, `strict`, `showCustomUi`.

We tend to remove most kwargs and prevent using them when not necessary.

gspread/utils.py
```diff
@@ -118,6 +118,41 @@ class GridRangeType(StrEnum):
     ListOfLists = "ListOfLists"
 
 
+class ValidationConditionType(StrEnum):
```
lavigne958, line 121: Could you add a new enum too for `RelativeDate` possible values ?

Documented here: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/other#relativedate

gspread/worksheet.py
```diff
@@ -3201,3 +3202,51 @@ def cut_range(
...
+
+            ``condition_type`` values are explained here: `ConditionType`_
+
+            .._ConditionType: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/other#ConditionType
+
+
+        :param str source: The A1 notation of the source range to move
+        :param condition_type: The sort of condition to apply.
+        :param values: List of condition values or RelativeDate.
+        :type values: Any | RelativeDate
```
lavigne958, line 3225: Can we please add a couple or examples in the documentation like: `NUMBER_GREAtER`.

And/or (depending on the time to make it, a popular based condition like : ONE_OF_LIST

lavigne958 (commented):
Hi thank you for this PR, I need to test it and check the API to confirm.

lavigne958:
What type of cassette recording mode did you use ?

gspread/worksheet.py
```diff
@@ -3201,3 +3202,51 @@ def cut_range(
...
+        grid = a1_range_to_grid_range(source, self.id)
+
+        body = {
+            "requests": [
+                {
+                    "setDataValidation": {
+                        "range": grid,
+                        "rule": {
+                            "condition": {
+                                "type": condition_type,
```
lavigne958, line 3240: Here we should check that the condition type is part of the enum otherwise we should raise an error.

gspread/utils.py
muddi900: I originally did add an enum for Relative Date but it is not supported by validation requests.

gspread/worksheet.py
muddi900: Added in 811ff02b40a53757ca52d259b5b6ced59106365f

gspread/worksheet.py
muddi900: Added in 06c3f24113244b1b8a5f285e5b5976db396d3a22

gspread/worksheet.py
muddi900: added in 811ff02b40a53757ca52d259b5b6ced59106365f

gspread/worksheet.py
```diff
@@ -3201,3 +3204,86 @@ def cut_range(
         }
 
         return self.client.batch_update(self.spreadsheet_id, body)
+
+    def add_validation(
+        self,
+        source: str,
+        condition_type: ValidationConditionType,
+        *values: Iterable[Any],
+        inputMessage: str | None = None,
```
lavigne958, line 3213: This syntax won't work on earliest versions of python we support. please use `Optional[str]` instead

gspread/worksheet.py
```diff
@@ -3201,3 +3204,86 @@ def cut_range(
         }
 
         return self.client.batch_update(self.spreadsheet_id, body)
+
+    def add_validation(
+        self,
+        source: str,
+        condition_type: ValidationConditionType,
+        *values: Iterable[Any],
```
lavigne958, line 3212: I don't fancy having an unbounded list of arguments, it has been quiet complex to simplify this project and add typing.

I would prefer to have a single argument that hold either a list, a tuple etc (any iterable in fact) and keep the number of arguments fixed in size. 

@alifeee  what do you think about this ?

gspread/worksheet.py
```diff
@@ -3201,3 +3204,86 @@ def cut_range(
...
+        :param condition_type: The sort of condition to apply.
+        :param values: List of condition values.
+        :type values: Any
+        :param str inputMessage: Message to show for the validation.
+        :param bool strict: Whether to reject invalid data or not.
+        :param bool showCustomUi: Whether to show a custom UI(Dropdown) for list values.
+
+        **Examples**
+        .. code-block:: python
+            import gspread
```
lavigne958, line 3236:
```suggestion
        **Examples**

        .. code-block:: python

            import gspread
```
you must add a blank line before and after the code-block

gspread/worksheet.py
```diff
@@ -3201,3 +3204,86 @@ def cut_range(
...
+        strict: bool = False,
+        showCustomUi: bool = False,
+    ) -> Any:
+        """Adds a data validation rule to any given range.
+
+        .. note::
+
+            ``condition_type`` values are explained here: `ConditionType`_
+
+            .._ConditionType: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/other#ConditionType
```
lavigne958, line 3223:
```suggestion
            .. _ConditionType: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/other#ConditionType
```
you need a space between the start `..` and the URL reference `_ConditionType`

lavigne958 (commented):
thank you for this proposal. it's well made.

You do need to change the typing notation in order to pass the CI linter.

gspread/worksheet.py
muddi900: I was going back and forth between unbouded values and an iterable. I have changed it in a35c39fbf80ae6b866f0eb38cda73c131666c4e1

gspread/worksheet.py
```diff
@@ -3201,3 +3204,88 @@ def cut_range(
         }
 
         return self.client.batch_update(self.spreadsheet_id, body)
+
+    def add_validation(
+        self,
+        source: str,
```
lavigne958, line 3210: could you please rename the argument `source` to `range`, we intend to use `range` for range name.

gspread/worksheet.py
```diff
@@ -3201,3 +3204,88 @@ def cut_range(
...
+        condition_type: ValidationConditionType,
+        values: Iterable[Any],
+        *,
```
lavigne958, line 3213: I don't understand why is this unnamed argument here then if we have an actual list of values into a single argument.

gspread/worksheet.py
muddi900: I get lint errors when I do that.

lavigne958: I think we missunderstood each others, what I mean is: the method argument currently name `source` should be renamed `range` please.
This is a method argument name I don't see how it can fail the linter :thinking:  unless there is something I don't see here.

gspread/worksheet.py
```diff
@@ -3201,3 +3204,88 @@ def cut_range(
...
+        :param values: List of condition values.
+        :type values: Any
+        :param str inputMessage: Message to show for the validation.
+        :param bool strict: Whether to reject invalid data or not.
+        :param bool showCustomUi: Whether to show a custom UI(Dropdown) for list values.
+
+        **Examples**
+
+        .. code-block:: python
+            import gspread
```
lavigne958, line 3238: you do need a blank line here between `.. code-block:: python` and the first line of code bellow in order to pass documentation build step

gspread/worksheet.py
lavigne958: I checked-out your branch, and the tests are failing because of the `*` argument.

you need to remove it.
currently the arguments:
- self
- ~source~ range
- condition_type
- values

are regular args with no default value.

The arguments:
- inputMessage
- strict
- showCustomUi

are kwargs, because they have a default value set.

lavigne958 (commented):
Please follow the given indications so the test + documentation build can succeed. 

Could you please rename the argument `showCustomUi` to `showCustomUI` with a uppercase `i`, it's an acronym so it should be uppercase. 

after that it's all good to, we will merge it once the changes are applied.

lavigne958:
I checked and the tests won't pass until you remove the following file and record it again.
`tests/cassettes/WorksheetTest.test_add_validation.json`

Quick trick, in order to regenerate only this cassette you can use the following command:
`tox -e py -- -k test_add_validation`
don't forget to set the env variables `GS_CREDS_FILENAME` and `GS_RECORD_MODE=new_episodes` so it only records this missing cassette.

gspread/worksheet.py
lavigne958: This has not been resolved. Please add a blank line before you start your code lines.

lavigne958:
Great thanks. Last thing left is the documentation to fix then all tests should pass.

muddi900:
Seems all tests are working, I'll flatten the commits in a moment

lavigne958 (approved):

gspread/worksheet.py
```diff
@@ -3201,3 +3204,88 @@ def cut_range(
...
+                10,
+                strict=True,
+                inputMessage='Value must be greater than 10',
+            )
+
+            ws.add_validation(
+                'C2:C7',
+                ValidationConditionType.one_of_list,
+                'Yes',
+                'No',
```
alifeee, line 3258: does this example work with the new format of [values]?

lavigne958: good catch  ! no it does not in fact, values must be put into a single list.

alifeee:
this looks wonderful

would we like to put it in any docs like the main `rst` example file, or the README, so that people actually know this new function exists?

:)

muddi900:
If there is a markdown to rst converter, I'll be game.

alifeee:
hmm, not sure about markdown to `rst`...

but I would add a simple code snippet, like
```python
ws.add_validation(
                'C2:C7',
                ValidationConditionType.one_of_list,
                ['Yes','No'],
                showCustomUi=True
            )
```

to the README (1)...

<https://github.com/burnash/gspread/blob/master/README.md?plain=1>

...and the user guide (2)...

<https://github.com/burnash/gspread/blob/master/docs/user-guide.rst?plain=1>

... in an appropriate place. For (1), it's markdown. For (2), you can run `tox -e doc` to build the docs to see if they build properly.

tests/worksheet_test.py
```diff
@@ -1871,3 +1871,19 @@ def test_get_and_get_values_have_same_signature(self):
...
+
+    @pytest.mark.vcr()
+    def test_add_validation(self):
+        sheet = self.sheet
+        self.assertDictEqual(
+            sheet.add_validation(
+                "A1",
+                utils.ValidationConditionType.one_of_list,
+                "y",
+                "n",
```
alifeee, line 1883: like the docs, this is also wrong now, no? it is only luck that it passes

it should be ["y", "n"]

it may also be nice to add one other validation condition to the tests, like `text_contains`, or `number_not_between`, etc, to get a bit more coverage.

(I think there is also an argument for testing each type of condition, since we *are* the ones to be providing each `enum`)... @lavigne958 ?

README.md
```diff
@@ -94,7 +94,7 @@ file.sheet1.update_tab_color(tab_color)
 
 ### Replace method `Worksheet.get_records`
 
-In v6 you can now only get *all* sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
+In v6 you can now only get _all_ sheet records, using `Worksheet.get_all_records()`. The method `Worksheet.get_records()` has been removed. You can get some records using your own fetches and combine them with `gspread.utils.to_records()`.
```
lavigne958, line 97: Why is this changing please ?

README.md
```diff
@@ -278,7 +278,7 @@ worksheet.batch_update([{
 
 ### Get unformatted cell value or formula
 
-```python
+````python
```
lavigne958, line 281: here too why is this changing please ?

lavigne958 (commented):
great addon with the documentation.

please remove unwanted changes and we should be good to go

lavigne958 (approved):

lavigne958:
great work thank you for this contribution !

PR #1447, opened 2024-03-28
Add pypi trusted publisher

### LINKED ISSUE #1331: Investigate Pypi Trusted Publisher
## Overview

We should enable trusted publishers in order to publish packages to TestPypi and Pypi without the need of passwords and username. (Don't know if we need it but it's worth checking-out)

lavigne958:
Remove use of token, use Pypi OIDC from github.

closes #1331

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,6 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    environment: release
     steps:
     - name: Checkout
       uses: actions/checkout@v4
@@ -36,11 +39,6 @@ jobs:
     - name: Publish to TestPyPi
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
-        user: __token__
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository-url: https://test.pypi.org/legacy/
     - name: Publish to PyPi
       uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}


alifeee (approved):

lavigne958:
Should I push an _empty_ new release 6.1.1 in order to test the release workflow? If it does not work then we can revert for now and push a 6.1.2 🤔

alifeee:
We could use the test.pypi? and you can make a release but tag it as `6.1.1-alpha` and not tick "most recent release"?

lavigne958:
> We could use the test.pypi? and you can make a release but tag it as `6.1.1-alpha` and not tick "most recent release"?

we could, it makes me think about something that could be useful:

- it would be nice to have some way to run this release workflow just for tests / pushing only to test.Pypi

To do that we can add some filters in the workflow, it can be based on the inputs and decide if this is a real release or not.

What I have in mind is:

- based on the format of the tag, decide to push or not the final step to Pypi

our tags are all formatted as: `vX.Y.Z`

so we could filter: 
- any tag that does not comply exactly like this is for test purpose
- any tag that exactly matches this is for release.

meaning:
- when pushing a tag like `v6.1.0-test.1` will only run the tests and push to test-pypi


what do you think ?

alifeee:
> what do you think ?

This sounds like a good idea to me.

I tried to think of a time we would like to push a tag that is not `vX.X.X` to PyPi, but I cannot. So, I think there is no problem with this idea.

lavigne958:
> > what do you think ?
> 
> This sounds like a good idea to me.
> 
> I tried to think of a time we would like to push a tag that is not `vX.X.X` to PyPi, but I cannot. So, I think there is no problem with this idea.

you're right we never release 'alpha' or 'beta' or release-candidate versions.

What I will do is:
- push some changes in the PR so we push only to test-pypi
- then push a garbage tag on the branch of the PR (not on master branch)
- this will trigger a release workflow on that tag
- that tag contains a release workflow file with only push to test-pypi
- let the workflow run
- if it succeeded
- add the final step to push to real pypi
- ready to merge :+1:

lavigne958:
alright, that was a good test as it does not work, nice we could catch it before real release.

may be @burnash  knows, could you check please the settings you provided to Pypi just in case, when you get a chance ? matching with the workflow filename, the workflow name (which is different) and the env too ? 

thank you.

Issue #1430, opened 2024-03-01
Add missing types


lavigne958:
## Overview

Add missing typing in code, all function arguments + return values should be typed.

## Details

We found some parts of the code that is not typed, all arguments and all returned values should be typed.

If possible we can try to type the tests too ! it would help while developing tests to have some auto-completion too.

lavigne958:
add JSON typing using: https://github.com/kevinheavey/jsonalias

lavigne958:
I tried to type the low level HTTP methods using the JSON type provided and it does not work. 
Any time we return some JSON type, we try to access the data as a dict or a list etc and mypy complains and says:
the return value could be a str, or a dict or a list and str object have no method `__getitem__` (which is true).
So it can't make the link: we return some value with multiple types possible we use as one of the possible type. 

let's keep the Any for now. we don't have a better solution.

I'll type the test so it helps any rich test editor to auto-complete and that would do for now I suppose.

PR #1448, opened 2024-04-01
Add some missing typing in code

lavigne958:
Introduce a lot of `Any` Typing is most method in the `Spreadsheet` object return some JSON object and we can't type it.

ref: #1430

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -39,7 +39,9 @@ class Client:
     ) -> None:
         self.http_client = http_client(auth, session)
 
-    def set_timeout(self, timeout: Optional[Union[float, Tuple[float, float]]] = None):
+    def set_timeout(
+        self, timeout: Optional[Union[float, Tuple[float, float]]] = None
+    ) -> None:
         """How long to wait for the server to send
         data before giving up, as a float, or a ``(connect timeout,
         read timeout)`` tuple.
@@ -76,7 +78,7 @@ class Client:
         return files
 
     def _list_spreadsheet_files(
-        self, title=None, folder_id=None
+        self, title: Optional[str] = None, folder_id: Optional[str] = None
     ) -> Tuple[List[Dict[str, Any]], Response]:
         files = []
         page_token = ""
@@ -316,9 +318,9 @@ class Client:
                     continue
 
                 new_spreadsheet.share(
-                    email_address=p["emailAddress"],
-                    perm_type=p["type"],
-                    role=p["role"],
+                    email_address=str(p["emailAddress"]),
+                    perm_type=str(p["type"]),
+                    role=str(p["role"]),
                     notify=False,
                 )
 
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -27,9 +27,6 @@ class Spreadsheet:
         self._properties = properties
 
         metadata = self.fetch_sheet_metadata()
-        import pprint
-
-        pprint.pprint(metadata)
         self._properties.update(metadata["properties"])
 
     @property
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -7,10 +7,13 @@ This module contains common spreadsheets' models.
 """
 
 import warnings
-from typing import Any, Dict, Union
+from typing import Any, Dict, Generator, Iterable, List, Mapping, Optional, Union
 
+from requests import Response
+
+from .cell import Cell
 from .exceptions import WorksheetNotFound
-from .http_client import HTTPClient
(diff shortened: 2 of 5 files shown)


gspread/client.py
```diff
@@ -316,9 +318,9 @@ def copy(
                     continue
 
                 new_spreadsheet.share(
-                    email_address=p["emailAddress"],
-                    perm_type=p["type"],
-                    role=p["role"],
+                    email_address=str(p["emailAddress"]),
+                    perm_type=str(p["type"]),
+                    role=str(p["role"]),
```
alifeee, line 323: this is not a typing change

gspread/spreadsheet.py
```diff
@@ -19,37 +22,40 @@
 class Spreadsheet:
     """The class that represents a spreadsheet."""
 
-    def __init__(self, http_client: HTTPClient, properties: Dict[str, Any]):
+    def __init__(self, http_client: HTTPClient, properties: Dict[str, Union[str, Any]]):
```
alifeee, line 25: what is the point of `Union[str, Any]`? Is this not basically the same as `Any`?

gspread/spreadsheet.py
```diff
@@ -19,37 +22,40 @@
...
         self.client = http_client
         self._properties = properties
 
         metadata = self.fetch_sheet_metadata()
+        import pprint
```
alifeee, line 30: imports should be at the top

is pprint a default library?

gspread/spreadsheet.py
```diff
@@ -398,7 +425,7 @@ def del_worksheet(self, worksheet):
 
         return self.client.batch_update(self.id, body)
 
-    def del_worksheet_by_id(self, worksheet_id: Union[str, int]):
+    def del_worksheet_by_id(self, worksheet_id: Union[str, int]) -> Any:
```
alifeee, line 428: why not use Response type here?

alifeee (commented):
there is a lot to this ! you found some inspiration and a love of types. it is good effort :)

I have left a few comments

Some questions:

why was so much left untyped?

is there a json type? Is it `Response`? Why do many of the return types use `Any` instead of `Response`?

gspread/client.py
lavigne958: it's not, the reason here is:
- the method `original.list_permissions` return a Dict wich contains values as `str` or `bool`
- mypy complains that we try to set values that could be boolean to a string (perm_type=p["type"])

In order to make it understand this is a string we should use the `str` function.

gspread/spreadsheet.py
lavigne958: the idea is telling mypy it could be `Any`type but also it could be a string, therefor when you assigne the result value to a variable that expects a string, mypy is happy :upside_down_face:

gspread/spreadsheet.py
lavigne958: oupsy :see_no_evil:   that was for debugging purpose

gspread/spreadsheet.py
lavigne958: because the method `self.client.batch_update()` return a JSON value that we currently type as `Any`

lavigne958:
> there is a lot to this ! you found some inspiration and a love of types. it is good effort :)
> 
> I have left a few comments
> 
> Some questions:
> 
> why was so much left untyped?

I believe when we merge different branches we picked the code from the untyped branch and left it as-is.

> 
> is there a json type? Is it `Response`? Why do many of the return types use `Any` instead of `Response`?

No there is not :cry: 
No `Response` is a a type from the `requests` library, it represents the `Response` class that is returned from a response from an HTTP request.

We return Any because returning a JSON type is too complex.

Mypy works with a static type, and no type heritage. Meaning: when you define a type like `Mapping[str, Any]` you can't tell mypy that the `Any` for the possible values could be the exact same type as itself.

When you want to type some JSON and this JSON is a Dict with any kind of value, you must define all possible type you could expect: List[...], Mapping[..., ...], str, int, .... and for each List[X] you must define what could be inside, so again you must provide all possible types.... In the end it's exactly the same as Any which in short means -> it could be any type. 

there is a huge github issue on the official python/typing repo about JSON, the very first comment shows a very clear example of what we need (but don't have): https://github.com/python/typing/issues/182 

In the mean time, we are left with Any

lavigne958:
Now that I checked the issue again, I found this ! https://github.com/kevinheavey/jsonalias

which could do the trick.

I would recommend we first merge this like this, then later in a second PR we change all Any to some `JSONType`

alifeee:
all great reasoning. mypy has whims and we must conform to them

lgtm :)

PR #1449, opened 2024-04-04
More fixes for `Worksheet.update` argument ordering & single cell updating (i.e. now `Worksheet.update_acell`)

alexmalins:
Fixes similar to #1431  for docs, a docstring and a test.

- `Worksheet.update` default argument ordering is now `Worksheet.update(<values>, <range name>, ...`
- Single cell updating uses the `Worksheet.update_acell` method now

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ cell_list = worksheet.findall(criteria_re)
 
 ```python
 # Update a single cell
-worksheet.update_acell('Bingo!', 'B1')
+worksheet.update_acell('B1', 'Bingo!')
 
 # Update a range
 worksheet.update([[1, 2], [3, 4]], 'A1:B2')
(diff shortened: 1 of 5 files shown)


alifeee:
hey! thanks for this change !! :)

it is very helpful. I am not sure of the best way of keeping the code and the docs in sync 🤔

alifeee (approved):

lavigne958 (approved):

lavigne958:
> hey! thanks for this change !! :)
> 
> it is very helpful. I am not sure of the best way of keeping the code and the docs in sync 🤔

there isn't in small projects like ours :disappointed: we must think about updating it every time we change something. 
> 
> I will let @lavigne958 merge :)
will do !

PR #1451, opened 2024-04-15
add `get_notes`

nbwzx:
get_note can only get the content of the note located at one cell, we need get_notes to get all notes in the sheet

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2606,6 +2606,32 @@ class Worksheet:
 
         return self.client.batch_update(self.spreadsheet_id, body)
 
+    def get_notes(self, default_empty_value: Optional[str] = None) -> List[List[str]]:
+        """Returns a list of lists containing all notes in the sheet, or the empty list if the
+        sheet does not have a note.
+
+        .. note::
+
+            The resulting matrix is as long as the last row holding a note
+                (could be smaller than the last data row)
+
+            The resulting matrix is as large as the last column holding a note
+                (could be smaller than the last data column)
+
+        :param str default_empty_value: (optional) Determines which value to use
+            for cells without notes, defaults to None.
+        """
+        params: ParamsType = {"fields": "sheets.data.rowData.values.note"}
+        res = self.client.spreadsheets_get(self.spreadsheet_id, params)
+        data = res["sheets"][self.index]["data"][0].get("rowData", [])
+        notes: List[List[str]] = []
+        for row in data:
+            notes.append([])
+            for cell in row.get("values", []):
+                notes[-1].append(cell.get("note", default_empty_value))
+
+        return notes
+
     def get_note(self, cell: str) -> str:
         """Get the content of the note located at `cell`, or the empty string if the
         cell does not have a note.


gspread/worksheet.py
```diff
@@ -2607,6 +2607,28 @@ def unmerge_cells(self, name):
...
 
+    def get_notes(self):
+        """Returns a list of lists containing all notes in the sheet, or the empty list if the
+        sheet does not have a note.
+        """
+        url = SPREADSHEET_URL % (self.spreadsheet.id)
+        params = {"fields": "sheets/data/rowData/values/note"}
+        response = self.client.request("get", url, params=params)
+        response.raise_for_status()
+        response_json = response.json()
```
lavigne958, line 2618: Please use the dedicated method to obtain the notes: `self.client.spreadsheets_get(self.spreadsheet_id, params)`

it will build the necessary URL for you and return a JSON, as well as raising a dedicated exception in case of error.

gspread/worksheet.py
```diff
@@ -2607,6 +2607,28 @@ def unmerge_cells(self, name):
...
+
+        try:
+            data = response_json["sheets"][self.index]["data"][0]["rowData"]
```
lavigne958, line 2621: Up to the field `["sheets"][self.index]["data"]` you don't need to check the API will return it for sure. even if you don't have any notes.

gspread/worksheet.py
```diff
@@ -2607,6 +2607,28 @@ def unmerge_cells(self, name):
...
+            notes = []
+            for row in data:
+                notes.append([])
+                for cell in row.get("values", []):
+                    notes[-1].append(cell.get("note", ""))
```
lavigne958, line 2626: please explain the reason why you add an empty list into the list `notes`. this creates a matrix (may be you were looking for the method `extend()` ?

I'd rather start with an empty list (as you did so far), and then add the notes to this list.
the `row.get("values", [])` is a good approach. 
Then all you need is: append the note if any, and here you should use:
`notes.append(cell.get("note"))` which will get the note string value if any or None if none exists in the list.
appending None will have no effect, so the list only grows as we append values.

lavigne958 (commented):
Hi @nbwzx thank you for this feature this is realy nice !

I have a few comments on how you solve this issue. once solved we can merge it.

gspread/worksheet.py
nbwzx: Yes. I think a matrix would be better since we can know the row index and column index of specific note.
Also, the function get_values returns a 2d-array. I want get_values and get_notes correspond to each other.
Thanks.

gspread/worksheet.py
lavigne958: in a way yes, what I mean is: you can reach for sure the data up to `["sheets"][self.index]["data"][0]`, the API will always give you that JSON tree at least. Then from there you can use the python notation to get the entry of a map or a default values like:

```python
data = response_json["sheets"][self.index]["data"][0].get("rowData", [])
```
from here you always have a list to iterate over.

gspread/worksheet.py
lavigne958: I see, that is a good point ! then I'll suggest:

- you set the default value to `None` which is the type used to define an empty value
- you should add a new parameter to the function `default_empty_value` so the user can choose what to use as empty value, some might want/need empty string `""` others might want something else (like `None`).

gspread/worksheet.py
```diff
@@ -2607,6 +2607,25 @@ def unmerge_cells(self, name):
 
         return self.spreadsheet.batch_update(body)
 
+    def get_notes(self):
+        """Returns a list of lists containing all notes in the sheet, or the empty list if the
+        sheet does not have a note.
+        """
+        params = {"fields": "sheets/data/rowData/values/note"}
```
lavigne958, line 2614: I noticed this works, thought the official JSON field mask notation uses dot `.` to separate key, please replace all `/` with a `.` to be aligned with the official notation.

nbwzx:
I have updated based on your suggestions. Thanks for your detailed comments!

gspread/worksheet.py
```diff
@@ -2607,6 +2607,21 @@ def unmerge_cells(self, name):
 
         return self.spreadsheet.batch_update(body)
 
+    def get_notes(self, default_empty_value = None):
+        """Returns a list of lists containing all notes in the sheet, or the empty list if the
+        sheet does not have a note.
```
lavigne958, line 2612: I tried your code it works fine.

Though there is an unexpected detail:

The list built and returned to the user could be smaller than the list of data returned to the user.
The point is, I have notes up to the row 3 and I have data up to the row 10.
The returned list of notes has a length of 3 because after row 3 there are o more notes.

somehow, we should mention that resulting matrix size. I don't have the right comment yet but we'll find something.

please add a note to the documentation, as follow with some phrase like bellow (feel free to change the text if you find something better (mind the spaces they matter in rst documentation format)):

```rst
.. note::

   The resulting matrix is as long as the last row holding a note (could be smaller than the last data row)
   The resulting matrix is as large as the last column holding a note (could be smaller than the last data column)
```

on the top of that please add the documenation markers for the typing of your arguments and returned values.
ask for any help if needed.

gspread/worksheet.py
```diff
@@ -2607,6 +2607,21 @@ def unmerge_cells(self, name):
 
         return self.spreadsheet.batch_update(body)
 
+    def get_notes(self, default_empty_value = None):
```
lavigne958, line 2610: please add the typing of arguments and returned value.

lavigne958 (commented):
tried it, works like a charm, what is left to do:
- documentation
- typing

then we can merge.

nbwzx:
I have added the documentation and typing, thanks.

gspread/worksheet.py
```diff
@@ -2607,6 +2607,32 @@ def unmerge_cells(self, name):
 
         return self.spreadsheet.batch_update(body)
 
+    def get_notes(self, default_empty_value: str = None) -> List[List[str]]:
```
lavigne958, line 2610: This typing won't pass, either you should set the defautl value to `""` an empty string, change the typing from `str` to `Optional[str]`.

nbwzx: Yeah. Sorry I ignored it. Now I have changed the typing to Optional[str]

lavigne958:
Thank you for the feature.

Please rebase your code on the master branch then we can run the CI to check/test your code.

nbwzx:
Unfortunately, tox cannot be installed on my computer because of the dependency conflicts. I have fixed the missing typing from the workflow.

lavigne958 (approved):
great thank you for this contribution.

alifeee:
I think a test should have been added before merging this PR. My apologies for not suggesting this sooner, but I did not expect it to be merged when it was, so I did not comment yet. :)

I have made such a test in the `test/get-notes` branch -> https://github.com/burnash/gspread/compare/master..test/get-notes

This function also raises a question of default behaviour: should `get_notes` return a square array or a non-square array?

non-square arrays are unexpected from users. See #1289 & #1305 for much more information on this, but effectively, which of the following structures should a user expect?

```python
[
  ["top left note", None],
  [None, "bottom right note"],
]
```

```python
[
  ["top left note"],
  [None, "bottom right note"],
]
```

Whichever we think makes the most sense should be represented in a test, so that behaviour is maintained.

Finally, I think this is very wordy

>         .. note::
>            The resulting matrix is as long as the last row holding a note
>                (could be smaller than the last data row)
>            The resulting matrix is as large as the last column holding a note
>                (could be smaller than the last data column)

and would better be represented with an example like provided above.

As @lavigne958 says, thank you for the work on this ! :)

nbwzx:
> I think a test should have been added before merging this PR. My apologies for not suggesting this sooner, but I did not expect it to be merged when it was, so I did not comment yet. :)
> 
> I have made such a test in the `test/get-notes` branch -> https://github.com/burnash/gspread/compare/master..test/get-notes

Yeah, that's pretty good.


> This function also raises a question of default behaviour: should `get_notes` return a square array or a non-square array?
> 
> non-square arrays are unexpected from users. See #1289 & #1305 for much more information on this, but effectively, which of the following structures should a user expect?
> 
> ```python
> [
>   ["top left note", None],
>   [None, "bottom right note"],
> ]
> ```
> 
> ```python
> [
>   ["top left note"],
>   [None, "bottom right note"],
> ]
> ```
> 
> Whichever we think makes the most sense should be represented in a test, so that behaviour is maintained.

I think a square array would be better after I reading your links. Also, it makes the output of get_values and get_notes correspond to each other, which is I expected.

By the way, shall we add some parameters of get_values into get_notes, e.g. range_name, maintain_size?

alifeee:
> I think a square array would be better after I reading your links. Also, it makes the output of get_values and get_notes correspond to each other, which is I expected.

Technically, for historical reasons *both* cases (square and not) are "default". See https://github.com/burnash/gspread/pull/1296#issuecomment-1783925954 for context, but effectively, `get_values` and `get` are the same function, but have different defaults, so with

| "this is " | - |
| --- | --- |
| - | "some text" |

and

```python
print("worksheet.get")
print(worksheet.get("A1:B2"))
print("worksheet.get_values")
print(worksheet.get_values("A1:B2"))
```

you get

```text
worksheet.get
[
  ['this is'],
  ['', 'some text']
]
worksheet.get_values
[
  ['this is', ''],
  ['', 'some text']
]
```

So... I am not sure which default this function should have, or whether it should have `maintain_size` kwargs. My preference is to keep it as simple as possible (so leave it as is), and add a note that users can use `utils.fill_gaps` themselves if they want a square array. For example, with `utils.fill_gaps(arr, len(arr), max(len(a) for a in arr))`.

> By the way, shall we add some parameters of get_values into get_notes, e.g. range_name, maintain_size?

I don't think so, because adding utilities and kwargs makes things complicated and hard to maintain. For example, `get_all_values` got out of hand before -> https://github.com/burnash/gspread/issues/1367, so I think keeping this simple works best.

I do think if there are no notes, `get_notes` should return `[[]]` instead of `[]` as this is what `worksheet.get` does.

Run

```python
print(worksheet.get_notes())
print(worksheet.get("H2:I3"))
```

on an empty sheet to get

```text
[]
[[]]
```

(I say this also partly because the above `fill_gaps` example works on `[[]]`, but errors on `[]`)

nbwzx:
Thank you for the explanation.
I agree with these changes:

1. add test for get_notes
2. add an example in documentation
3. get_notes return [[]] instead of [] if there are no notes

You have already finished (1) in [test/get-notes](https://github.com/burnash/gspread/tree/test/get-notes), I don't know how can I help with others since it is a merged pr. Shall I open a new pr or is it convenient for you to update them?

alifeee:
if you'd like to contribute, you can pull the branch with the test and make a new PR.

Thanks for the enthusiasm!

lavigne958:
> @alifeee @nbwzx I understand this is closed, just wondering why not returning `Mapping[str, str]` (key by a1, value by note content) . As it will reduce the confusion of introducing matrix, is there any other api returning matrixes?

Hi when using gspread most data  you get is in a matrix format (or a single list format when requesting a single row) so the general usage is to iterate over a matrix.

Iteration over a dict is not standard, the only method providing a `dict` result is `get_all_records` which returns a list of dict due to parsing of headers.

In this way we agree to return a matrix (list of list).

twosheds:
I think get_notes would be more useful if  if could be applied to a range.

alifeee:
> I think get_notes would be more useful if if could be applied to a range.

Hi @twosheds, thanks for the suggestion! Since this PR is closed, could you instead open an issue? thanks

PR #1453, opened 2024-04-16
Add property expiry in gspread client

### LINKED ISSUE #1169: Specific exception in case an authorization has expired
In order not to need to create a new authorization each time I need it (sometimes I use it 8 times per code loop), I create a global object:
```
import gspread

client_gspread = None
def send_sheet(id_sheet, page_sheet, sheet_data):
    global client_gspread
    if not client_gspread:
        client_gspread = gspread.service_account(filename='client_secret.json')

    sheet = client_gspread.open_by_key(id_sheet).worksheet(page_sheet)
    sheet.clear()
    sheet.update([sheet_data.columns.values.tolist()] + sheet_data.values.tolist())
```
The problem with this is that authorization has an expiration time, which we find in `client_gspread.auth.expiry`

It would be interesting if there was a single exception that handled the case of expired authorization, to be used directly when you need to update the authorization.

In the documentation in the area explaining how to authorization, it would be nice to also add how long an authorization lasts, so that new users can already create their projects knowing the time limitations of authorizations. Oh sure, it would also be nice to add to the documentation the answer to the question "if I perform a real action using gspread, is the authorization expiration time updated or fixed?"

Because when the authorization is created, when analyzing its `vars`, we find the `timeout` attribute, which has no value, so the person may mistakenly understand that there is no expiration date.

lavigne958:
In gspread client add a new property expiry which returns the expiry datetime object from the used credentials.

The expiry only exists for OAuth credentials otherwise it returns None.

### DIFF
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -6,6 +6,7 @@ This module contains Client class responsible for managing spreadsheet files
 
 """
 
+from datetime import datetime
 from http import HTTPStatus
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -38,6 +39,18 @@ class Client:
         http_client: HTTPClientType = HTTPClient,
     ) -> None:
         self.http_client = http_client(auth, session)
+
+    @property
+    def expiry(self) -> Optional[datetime]:
+        """Returns the expiry date of the curenlty loaded credentials
+
+        :returns: (optional) datetime the expiry date time object.
+
+        .. note::
+
+           It only applies to gspread client created using oauth
+        """
+        return self.http_client.auth.expiry
 
     def set_timeout(
         self, timeout: Optional[Union[float, Tuple[float, float]]] = None


alifeee:
Hi :)

This seems good. It seems like something difficult to test(?)

The issue #1169 asked for:

> It would be interesting if there was a single exception that handled the case of expired authorization, to be used directly when you need to update the authorization.

I do not see that here. Is it a new `GspreadException` type that we could raise?

lavigne958:
> Hi :)
> 
> This seems good. It seems like something difficult to test(?)

yes because it needs the OAuth authentication method to be tested :cry: 

> 
> The issue #1169 asked for:
> 
> > It would be interesting if there was a single exception that handled the case of expired authorization, to be used directly when you need to update the authorization.
> 
> I do not see that here. Is it a new `GspreadException` type that we could raise?

 I chose not to raise an exception as the expiry can be refreshed on the first API call. and it's easier if the user checks the expiry and chooses to raise when it happens.

I'll try to catch the API error in case of expired token but it needs me to run a query, then wait for at least a week and try again. So far I did not managed to trigger some API error it just refreshes silently

alifeee:
okay. It seems a simple enough change 👍.

Not sure if it fixes #1169 though as they asked for an Exception to be thrown(?)

lavigne958:
> okay. It seems a simple enough change 👍.
> 
> Not sure if it fixes #1169 though as they asked for an Exception to be thrown(?)

true, I won't close the user issue, I will let the user choose if this is enough and then close the issue it solves the issue.

lavigne958:
@alifeee you're ok to merge this as is ? knowing that it's not solving #1169 . but so far it provides the info to the user.

alifeee:
> @alifeee you're ok to merge this as is ? knowing that it's not solving #1169 . but so far it provides the info to the user.

sure, I am ok. I still do not fully understand auth, so I would not know when this would be useful. But if it is good for you it is good for me. :)

lavigne958:
> sure, I am ok. I still do not fully understand auth, so I would not know when this would be useful. But if it is good for you it is good for me. :)

auth works using a class that encapsulate the HTTP requests. It does that by doing authentication calls before the actual call we want to do sometimes or adds some token to the HTTP call without us knowing about it. 

Here the auth class can return the expiry of the given token, if it is expired then the user is expected to remove the file `authorized_user.json` in order to force a new login.

alifeee:
ok, so the user would be expected to do something like...

```python
if client.expiry > datetime.today():
  # throw error! cannot do things properly!
# no error, continue
worksheet.update_acell("A1", "hi")
```

If I have misunderstood, perhaps you could provide a pseudocode sample in which this attribute could be used helpfully? :)

lavigne958:
> ok, so the user would be expected to do something like...
> 
> ```python
> if client.expiry > datetime.today():
>   # throw error! cannot do things properly!
> # no error, continue
> worksheet.update_acell("A1", "hi")
> ```
> 
> If I have misunderstood, perhaps you could provide a pseudocode sample in which this attribute could be used helpfully? :)

yes that's how it could be used.
I don't feel confident enough in the value hold by the authentication library to trigger an exception, but I'm fine exposing this value for any user to access it.

lavigne958:
To be honest I check the expiry value and it seems 1h in the past, I tried to print the value with UTC timezone and still 1h in the past, so that means the expiry date provided is wrong as it's already past so I am wondering on how is this value accurate.

I am now running a script that uses oauth, sends a request every hours I will check if the expiry value has changed.

alifeee:
ok. seems a good change 👌 can merge

lavigne958:
actually after my last test, the script has been running for over a week and never expired.
I believe we should close this PR. The expiry attribute is not reliable, it does not actually provide an expected expiry date.

lavigne958:
@alifeee I am ready to merge this PR, after checking the behavior, even if the credentials are expired, we can make requests and it still works anyway. and credentials get refreshed ! 

if anyone needs to know we can expose this property anyway even if it has a limited usage it does not break anything and provides information to the user.

Waitting for you review.

alifeee (approved):

PR #1461, opened 2024-05-01
Some updates on `get_notes`

nbwzx:
1. add test for get_notes
2. add an example in documentation
3. get_notes return [[]] instead of [] if there are no notes

See [#1251 (comment)](https://github.com/burnash/gspread/pull/1451#issuecomment-2084156934) for more details.

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2606,7 +2606,7 @@ class Worksheet:
 
         return self.client.batch_update(self.spreadsheet_id, body)
 
-    def get_notes(self, default_empty_value: Optional[str] = None) -> List[List[str]]:
+    def get_notes(self, default_empty_value: Optional[str] = "") -> List[List[str]]:
         """Returns a list of lists containing all notes in the sheet.
 
         .. note::
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2623,7 +2623,7 @@ class Worksheet:
         """
         params: ParamsType = {"fields": "sheets.data.rowData.values.note"}
         res = self.client.spreadsheets_get(self.spreadsheet_id, params)
-        data = res["sheets"][self.index]["data"][0].get("rowData", [])
+        data = res["sheets"][self.index]["data"][0].get("rowData", [{}])
         notes: List[List[str]] = []
         for row in data:
             notes.append([])
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2635,12 +2635,12 @@ class Worksheet:
             >>> print(arr)
             [
                 ["A1"],
-                [None, "B2"]
+                ["", "B2"]
             ]
             >>> print(gspread.utils.fill_gaps(arr, len(arr), max(len(a) for a in arr), None))
             [
-                ["A1", None],
-                [None, "B2"]
+                ["A1", ""],
+                ["", "B2"]
             ]
         """
         params: ParamsType = {"fields": "sheets.data.rowData.values.note"}
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2620,6 +2620,21 @@ class Worksheet:
 
         :param str default_empty_value: (optional) Determines which value to use
             for cells without notes, defaults to None.
+
+        Examples::
+
+            # Note data:
+            #      A      B
+            #
+            # 1    A1     None
+            # 2    None   B2
+
+            # Read all notes from the sheet
+            >>> worksheet.get_notes()
+            [
+                ["A1"],
+                [None, "B2"]
+            ]
         """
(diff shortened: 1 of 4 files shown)


alifeee:
One note, calling `get_note` on a cell with no note returns the empty string `""`, whereas calling `get_notes` returns `None`. I feel that this default could be changed to align with `get_note`. What do you think?

I noticed this because `fill_gaps` defaults to filling with `""`, not `None`, so I had to specify `None` to keep with the default for `get_notes`.

For an example, run on a cell with no note:

```bash
>>> print(repr(worksheet.get_note("A2")))
''
```

nbwzx:
Sure. I agree with that. Thank you.

alifeee:
> Sure. I agree with that. Thank you.

Would you like to make this change or shall I? :)

nbwzx:
> > Sure. I agree with that. Thank you.
> 
> 
> 
> Would you like to make this change or shall I? :)

I want to know what @lavigne958 thinks of this change from None to empty string. In my previous pr, he mentioned that (I'll suggest you set the default value to None which is the type used to define an empty value).

lavigne958 (approved):

lavigne958:
> > > Sure. I agree with that. Thank you.
> > 
> > 
> > Would you like to make this change or shall I? :)
> 
> I want to know what @lavigne958 thinks of this change from None to empty string. In my previous pr, he mentioned that (I'll suggest you set the default value to None which is the type used to define an empty value).

the changes look good.

About the default value, The value `None` seems more appropriate:
- we return a list of values here, not a single value
- within that list in order to show an empty value `None` seems a good result
- In single note method like `get_note` we expect a returned value as we expect to get a note from a cell.
  (though we could change that to something elese, but thats's a breaking change... so it should wait)

The way I see it is: when a user uses the function, we always need to check for empty string, or with `None` we ca simply check if value is not `None` which seems easier to read like:

```python
notes = sheet.get_notes()

for note in notes:
  if note is not None:
    do_something(note)
```

if you both believe an empty string is better, then I have no problem, go for empty string :upside_down_face:

nbwzx:
> The way I see it is: when a user uses the function, we always need to check for empty string, or with `None` we ca simply check if value is not `None` which seems easier to read like:

I don't understand this. For functions handing strings, we usually don't need to specifically check for empty string since it is type `str`. So I think it is more easy to use without using `if note is not None`. @alifeee Could please make this change if you think empty string is better?

lavigne958:
> > The way I see it is: when a user uses the function, we always need to check for empty string, or with `None` we ca simply check if value is not `None` which seems easier to read like:
> 
> I don't understand this. For functions handing strings, we usually don't need to specifically check for empty string since it is type `str`. So I think it is more easy to use without using `if note is not None`. @alifeee Could please make this change if you think empty string is better?

make sense, the returned matrix could contain only valid strings so either empty string `""` or the note content `abcd` 
As long as the result is _logic_ and makes sense for the users I'm fine with it.

We could keep the open logic to the user by keeping the default value, we set this value to empty string `""` but if the user wants something else... they can !

alifeee:
`get` and `get_values` return `""` by default, no?

What is the difference between `if note is not None` and `if note is not ""`?

I got lost in your opinions of whether you prefer `""` or `None` for the default

lavigne958:
> `get` and `get_values` return `""` by default, no?
> 
> What is the difference between `if note is not None` and `if note is not ""`?

there is none in term of code, both need to check for value presence. The difference is in the semantic of the value `None` which represents "no value" or a the actual empty string reprents a string that is here but empty. 

it philosophical :laughing: 
> 
> I got lost in your opinions of whether you prefer `""` or `None` for the default

I prefer `""` as you said it makes more sense for the users.

alifeee:
A small review, mostly by searching for `= None`, and `[^"]""[^"]` (regex search) in `worksheet.py` and `utils.py`

Functions using `None` as a default

- `ValueRange.first`

Functions using `""` as a default

- `Worksheet.cell`
- `Worksheet.get_all_records`
- `Worksheet.get_note`
- `utils.numericise`
- `utils.numericise_all`
- `utils.rightpad`
- `utils.fill_gaps` (used by Worksheet.range & Worksheet.get)

From this I would say we should strive to use `""` as a default for "a value", as it is in-line with expectations. There may be cases where this is not desired. I will make this change here, and then I think this PR is ready? :)

alifeee:
rebased because of https://github.com/burnash/gspread/pull/1464

lavigne958 (approved):

alifeee (approved):
@lavigne958 I think this should have gotten into 6.1.1, and so we should release 6.1.2 to include this ;)

lavigne958:
> @lavigne958 I think this should have gotten into 6.1.1, and so we should release 6.1.2 to include this ;)

my bad, I'll do the release right now !

PR #1463, opened 2024-05-10
Fix/api key auth version

### LINKED ISSUE #1454: gspread depends on google-auth>2.4.0 but only specifies google-auth>=1.12.0
**Describe the bug**
I have gspread installed against version 1.35.0 of google-auth which is within the constraints currently specified by `pyproject.toml`  (`google-auth>=1.12.0`) and get the following error upon `import gspread`:

`ModuleNotFoundError: No module named 'google.auth.api_key'`.

Looks like this module [only came into existence in google-auth v2.4.0](https://github.com/googleapis/google-auth-library-python/commit/3b15092b3461278400e4683060f64a96d50587c4#diff-4abb41958447ee7c433a05ab8f3d736e63dd5b38d7c64ddd7142c7c69cf194db) so that's the constraint that should be actually specified

**To Reproduce**
1. Install gspread alongside any version of google-auth less than 2.4.0
2. Import gspread
3. Observe `ModuleNotFoundError: No module named 'google.auth.api_key'`.

alifeee:
closes #1454 

changing `test-requirements.txt` means that the tests for `api_key` auth *do* fail.

Adding the "fix" means the tests "work" (they do not throw an exception)

### DIFF
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -12,8 +12,10 @@ from pathlib import Path
 from typing import Any, Dict, Iterable, Mapping, Optional, Protocol, Tuple, Union
 
 from google.auth.credentials import Credentials
+
 try:
     from google.auth.api_key import Credentials as APIKeyCredentials
+
     GOOGLE_AUTH_API_KEY_AVAILABLE = True
 except ImportError:
     GOOGLE_AUTH_API_KEY_AVAILABLE = False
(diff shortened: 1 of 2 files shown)


lavigne958 (approved):
nice fix ! 

I noticed our test suite uses a different version for the google auth lib.

I coul add a new test in the CI that checks the version of the dependencies in the file `pyproject.toml` and in the `test-requirements.txt` file and ensures they are the same ! otherwise we forgot to update one or another.

alifeee:
> I coul add a new test in the CI that checks the version of the dependencies in the file `pyproject.toml` and in the `test-requirements.txt` file and ensures they are the same ! otherwise we forgot to update one or another.

I don't think this is necessary. After this PR, they will be the same.

If we update the version in pyproject.toml, it will be presumably because we had to (to fix something broken). When we fix it, the tests should break too (as they will be using an older version of auth libraries).

I will merge this 👍

PR #1465, opened 2024-05-13
add note about runnings tests to contrib guide

alifeee:
I run tests without `tox`. I find this easier.

I think people should have this knowledge. They should also know that the CI runs tests with tox.

Also removed `-v` (verbose) and `-s` (silence stderr/stdout) as I do not think these are useful, and having lots of extra CLI arguments is imposing for new CLI users.


lavigne958 (approved):
Nice ! I'll make the v6.1.1 release without this one, it only updates the documentation which is only on github

PR #1469, opened 2024-05-15
Remove StrEnum dependency and added custom class[issue #1462]

### LINKED ISSUE #1462: Remove StrEnum dependency
gspread dependencies are:

https://github.com/burnash/gspread/blob/c740b88e072962ffe772ec3bd90907a579524692/pyproject.toml#L30-L34

StrEnum [amounts to](https://github.com/irgeek/StrEnum/blob/ab34b770aacac80431cd77f28770a60144679d38/strenum/__init__.py#L21-L49)

```python
class StrEnum(str, enum.Enum):
    def __new__(cls, value, *args, **kwargs):
        if not isinstance(value, (str, enum.auto)):
            raise TypeError(
                f"Values of StrEnums must be strings: {value!r} is a {type(value)}"
            )
        return super().__new__(cls, value, *args, **kwargs)

    def __str__(self):
        return str(self.value)

    def _generate_next_value_(name, *_):
        return name
```

We ought not need to rely on a dependency, and can think of a way of including this code in the project, so that the dependencies are just google auth libraries.

@lavigne958 thoughts?

muddi900:
Fixes #1462

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -6,6 +6,7 @@ This module contains utility functions.
 
 """
 
+import enum
 import re
 from collections import defaultdict
 from collections.abc import Sequence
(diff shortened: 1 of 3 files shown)


lavigne958 (approved):
Thank you for this contribution !

PR #1474, opened 2024-06-01
Fix code block formatting typo in README

agrvz:
Small fix to correct formatting of code block example in README for "Add data validation to a range".

Before:
before

After:
after


lavigne958:
Thank you for this contribution!

I'll merge this changes regardless of the failed CI for this time only, exceptionally.

PR #1475, opened 2024-06-02
[feature] Add 'expand_table' feature

### LINKED ISSUE #1414: cell.expand('table') feature requests
In the past, I used to use [xlwings](https://docs.xlwings.org/en/latest/index.html) library for Excel. In their API they have a very handy way of selecting "tables"  starting for a specific cell. 

you could use it, for example, to get data from a series of sheets that have table with variable size but always start from the cell `"A1"`. so you "expand" the selection similar to what you do when do `cmd+left_arrow`, `cmd+down_arrow` on the worksheet, 

The usage is better explained in the documentation:
[`range.expand('table')` ](https://docs.xlwings.org/en/latest/api/range.html#xlwings.Range.expand)

this is a snippet of the usage, expand have one of three arguments `"table", "right","down"`. 
```python
import xlwings as xw
wb = xw.Book()
sheet1 = wb.sheets[0]
sheet1.range('A1').value = [[None, 1], [2, 3]]
sheet1.range('A1').expand().address
sheet1.range('A1').expand('table').addres
```

I implemented the `"table"` version for myself just because I found it very handy. 
```python
    def expand_table(worksheet: gspread.worksheet.Worksheet, cell: gspread.Cell) -> str:
        """
        Expands a cell range in gspread based on non-null adjacent cells, similar to
        xlwings.Range('A1').expand('table').
        """

        if not cell.value:
            raise ValueError(f"Cell at {cell} is empty.")
        
        row_values = worksheet.row_values(cell.row)
        row_values = row_values[row_values.index(cell.value):]
        row_values = row_values[:row_values.index('')] if '' in row_values else row_values
        right_boundary_col = cell.col + len(row_values) -1
        
        column_values = worksheet.col_values(right_boundary_col)
        column_values = column_values[column_values.index(worksheet.cell(row=cell.row,col=right_boundary_col).value):]
        column_values = column_values[:column_values.index('')] if '' in column_values else column_values
        bottom_boundary_row = cell.row + len(column_values) -1

        up_a1 = gspread.utils.rowcol_to_a1(cell.row,cell.col)
        low_a1 = gspread.utils.rowcol_to_a1(bottom_boundary_row,right_boundary_col )

        return f'{up_a1}:{low_a1}' 

```
 maybe there's a way of doing this using `gspread`, but I couldn't find it. 
 
Thanks in advance!

lavigne958:
Add a new feature that allows a user to expand a cell range
into a table.

the expand will look for the right most cell with adjacent value.
the expand will look for the bottom most cell with adjacent value.
the expand will table down from top left celle range to bottom right
value.

closes #1414

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -986,49 +986,37 @@ def to_records(
 
 
 def _expand_right(values: List[List[str]], start: int, end: int, row: int) -> int:
-    """This is a private function, returning the column index of the first empty cell
+    """This is a private function, returning the column index of the last non empty cell
     on the given row.
 
     Search starts from ``start`` index column.
     Search ends on ``end`` index column.
     Searches only in the row pointed by ``row``.
-
-    If no empty value is found, it will return the given ``end`` index.
     """
-    for column in range(start, end):
-        # in case the given row is smaller that what is being asked
-        if column >= len(values[row]):
-            return len(values[row]) - 1
-
-        if values[row][column] == "":
-            return column
-
-    return end
+    try:
+        return values[row].index("", start, end) - 1
+    except ValueError:
+        return end
 
 
 def _expand_bottom(values: List[List[str]], start: int, end: int, col: int) -> int:
-    """This is a private function, returning the row index of the first empty cell
+    """This is a private function, returning the row index of the last non empty cell
     on the given column.
 
     Search starts from ``start`` index row.
     Search ends on ``end`` index row.
     Searches only in the column pointed by ``col``.
-
-    If no empty value is found, it will return the given ``end`` index.
     """
     for rows in range(start, end):
         # in case we try to look further than last row
         if rows >= len(values):
             return len(values) - 1
 
-        # this row is smaller than the others, just keep looking
-        if col >= len(values[rows]):
-            continue
+        # check if cell is empty (or the row => empty cell)
+        if col >= len(values[rows]) or values[rows][col] == "":
+            return rows - 1
 
-        if values[rows][col] == "":
-            return rows
-
-    return end
+    return end - 1
 
 
 def find_table(
@@ -1042,10 +1030,12 @@ def find_table(
 
         * ``TableDirection.right``: expands right until the first empty cell
         * ``TableDirection.down``: expands down until the first empty cell
-        * ``TableDirection.table``: expands right until the first empty cell, then down until the first empty cell
+        * ``TableDirection.table``: expands right until the first empty cell and down until first empty cell
 
-    Regardless of the direction this function always returns a matrix of data, even if it has
-    only one column.
+    In case of empty result an empty list is restuned.
+
+    When the given ``start_range`` is outside the given matrix of values the exception
+    `~gspread.exceptions.InvalidInputValue` is raised.
 
     Example::
 
@@ -1064,7 +1054,7 @@ def find_table(
 
     .. note::
 
-       the ``TableDirection.table`` will first look right, then look down.
+       the ``TableDirection.table`` will look right from starting cell then look down from starting cell.
        It will not check cells located inside the table. This could lead to
        potential empty values located in the middle of the table.
 
@@ -1083,28 +1073,37 @@ def find_table(
     row -= 1
     col -= 1
 
+    if row >= len(values):
+        raise InvalidInputValue(
+            "given row for start_range is outside given values: start range row ({}) >= rows in values {}".format(
+                row, len(values)
+            )
+        )
+
+    if col >= len(values[row]):
+        raise InvalidInputValue(
+            "given column for start_range is outside given values: start range column ({}) >= columns in values {}".format(
+                col, len(values[row])
+            )
+        )
+
     if direction == TableDirection.down:
-        rightMost = col + 1
+        rightMost = col
         bottomMost = _expand_bottom(values, row, len(values), col)
 
     if direction == TableDirection.right:
+        bottomMost = row
         rightMost = _expand_right(values, col, len(values[row]), row)
-        bottomMost = row + 1
 
     if direction == TableDirection.table:
         rightMost = _expand_right(values, col, len(values[row]), row)
-
-        checkColumn = rightMost
-        if checkColumn != 0:
-            checkColumn -= 1
-
-        bottomMost = _expand_bottom(values, row, len(values), checkColumn)
+        bottomMost = _expand_bottom(values, row, len(values), col)
 
     result = []
 
     # build resulting array
-    for rows in values[row:bottomMost]:
-        result.append(rows[col:rightMost])
+    for rows in values[row : bottomMost + 1]:
+        result.append(rows[col : rightMost + 1])
 
     return result
 
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -168,6 +168,12 @@ class ValidationConditionType(StrEnum):
     filter_expression = "FILTER_EXPRESSION"
 
 
+class TableDirection(StrEnum):
+    table = "TABLE"
+    down = "DOWN"
+    right = "RIGHT"
+
+
 def convert_credentials(credentials: Credentials) -> Credentials:
     module = credentials.__module__
     cls = credentials.__class__.__name__
@@ -977,6 +983,130 @@ def to_records(
     """
 
     return [dict(zip(headers, row)) for row in values]
+
+
+def _expand_right(values: List[List[str]], start: int, end: int, row: int) -> int:
+    """This is a private function, returning the column index of the first empty cell
+    on the given row.
+
+    Search starts from ``start`` index column.
+    Search ends on ``end`` index column.
+    Searches only in the row pointed by ``row``.
+
+    If no empty value is found, it will return the given ``end`` index.
+    """
+    for column in range(start, end):
+        # in case the given row is smaller that what is being asked
+        if column >= len(values[row]):
+            return len(values[row]) - 1
+
+        if values[row][column] == "":
+            return column
+
+    return end
+
+
+def _expand_bottom(values: List[List[str]], start: int, end: int, col: int) -> int:
+    """This is a private function, returning the row index of the first empty cell
+    on the given column.
+
+    Search starts from ``start`` index row.
+    Search ends on ``end`` index row.
+    Searches only in the column pointed by ``col``.
+
+    If no empty value is found, it will return the given ``end`` index.
+    """
+    for rows in range(start, end):
+        # in case we try to look further than last row
+        if rows >= len(values):
+            return len(values) - 1
+
+        # this row is smaller than the others, just keep looking
+        if col >= len(values[rows]):
+            continue
+
+        if values[rows][col] == "":
+            return rows
+
+    return end
+
+
+def find_table(
+    values: List[List[str]],
+    start_range: str,
+    direction: TableDirection = TableDirection.table,
+) -> List[List[str]]:
+    """Expands a list of values based on non-null adjacent cells.
+
+    Expand can be done in 3 directions defined in :class:`~gspread.utils.TableDirection`
+
+        * ``TableDirection.right``: expands right until the first empty cell
+        * ``TableDirection.down``: expands down until the first empty cell
+        * ``TableDirection.table``: expands right until the first empty cell, then down until the first empty cell
+
+    Regardless of the direction this function always returns a matrix of data, even if it has
+    only one column.
+
+    Example::
+
+        values = [
+            ['', '',   '',   '', ''  ],
+            ['', 'B2', 'C2', '', 'E2'],
+            ['', 'B3', 'C3', '', 'E3'],
+            ['', ''  , ''  , '', 'E4'],
+        ]
+        >>> utils.find_table(TableDirection.table, 'B2')
+        [
+            ['B2', 'C2'],
+            ['B3', 'C3'],
+        ]
+
+
+    .. note::
+
+       the ``TableDirection.table`` will first look right, then look down.
+       It will not check cells located inside the table. This could lead to
+       potential empty values located in the middle of the table.
+
+    .. warning::
+
+       Given values must be padded with `''` empty values.
+
+    :param list[list] values: values where to find the table.
+    :param gspread.utils.TableDirection direction: the expand direction.
+    :param str start_range: the starting cell range.
+    :rtype list(list): the resulting matrix
+    """
+    row, col = a1_to_rowcol(start_range)
+
+    # a1_to_rowcol returns coordinates starting form 1
+    row -= 1
+    col -= 1
+
+    if direction == TableDirection.down:
+        rightMost = col + 1
+        bottomMost = _expand_bottom(values, row, len(values), col)
+
+    if direction == TableDirection.right:
+        rightMost = _expand_right(values, col, len(values[row]), row)
+        bottomMost = row + 1
+
+    if direction == TableDirection.table:
+        rightMost = _expand_right(values, col, len(values[row]), row)
+
+        checkColumn = rightMost
+        if checkColumn != 0:
+            checkColumn -= 1
+
+        bottomMost = _expand_bottom(values, row, len(values), checkColumn)
+
+    result = []
+
+    # build resulting array
+    for rows in values[row:bottomMost]:
+        result.append(rows[col:rightMost])
+
+    return result
 
 
 # SHOULD NOT BE NEEDED UNTIL NEXT MAJOR VERSION
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -1035,7 +1035,7 @@ def find_table(
     In case of empty result an empty list is restuned.
 
     When the given ``start_range`` is outside the given matrix of values the exception
-    `~gspread.exceptions.InvalidInputValue` is raised.
+    :class:`~gspread.exceptions.InvalidInputValue` is raised.
 
     Example::
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -41,6 +41,7 @@ from .utils import (
     PasteOrientation,
     PasteType,
     T,
+    TableDirection,
     ValidationConditionType,
     ValueInputOption,
     ValueRenderOption,
@@ -53,6 +54,7 @@ from .utils import (
     convert_colors_to_hex_value,
     convert_hex_to_colors_dict,
     fill_gaps,
+    find_table,
     finditem,
     get_a1_from_absolute_range,
     is_full_a1_notation,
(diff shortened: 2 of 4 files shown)


gspread/worksheet.py
```diff
@@ -3336,3 +3338,48 @@ def add_validation(
         }
 
         return self.client.batch_update(self.spreadsheet_id, body)
+
+    def expand_table(
```
alifeee, line 3342: this function could have a simplified name. I am not sure "expand_table" is that user-friendly? perhaps just "expand"? or maybe it is good

alifeee (changes requested):
this is a nice self-contained change. good work here :)

I think we need more tests, as I foresee a lot of edge cases, and it would be good to catch those before the feature is in gspread.

For example, what is the expectation for expanding this table from the top left?

| 1 | 1 | - | 1 |
| --- | --- | --- | --- |
| 1 | 1 | 1 | 1 |
| - | - | 1 | - |
| 1 | 1 | 1 | 1 |

This is not clear to me. Also the internal (private) `utils` expanding functions are unclear what they do.

As well, we should have a `utils` test *and* a `worksheet` test. Currently there is only `utils`

gspread/utils.py
```diff
@@ -979,6 +985,99 @@ def to_records(
     return [dict(zip(headers, row)) for row in values]
 
 
+def _expand_right(values: List[List[str]], start: int, end: int, row: int) -> int:
+    # find rightmost value
```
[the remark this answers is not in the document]
lavigne958: true ! I was quiet short ~and lazy~ on this one :laughing: 

I updated it.

gspread/utils.py
```diff
@@ -979,6 +985,99 @@ def to_records(
...
+    for column in range(start, end):
+        if values[row][column] == "":
+            return column
+
+    return end
+
+
+def _expand_bottom(values: List[List[str]], start: int, end: int, col: int) -> int:
+    # find the bottommost value
```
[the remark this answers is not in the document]
lavigne958: I updated it.

gspread/utils.py
```diff
@@ -979,6 +985,99 @@ def to_records(
...
+    for rows in range(start, end):
+        print("bottom check row: {} col {}".format(rows, col))
```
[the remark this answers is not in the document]
lavigne958: true :disappointed:  :facepalm:

gspread/utils.py
```diff
@@ -979,6 +985,99 @@ def to_records(
...
+        [
+            ['B2', 'C2'],
+            ['B3', 'C3'],
+        ]
+
+
+    .. note::
+
+       the ``TableDirection.table`` will first look right, then look down.
+       Any empty value in the midle of the table will be ignored.
```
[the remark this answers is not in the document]
lavigne958: true, I will rework the docstring !

here in your example: it will stop on the first empty cell on the right, so V2 is the rightmost cell. Then it will go down until D2

gspread/worksheet.py
```diff
@@ -3336,3 +3338,48 @@ def add_validation(
...
+        self,
+        direction: TableDirection,
```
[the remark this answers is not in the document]
lavigne958: right, that would help users, I suppose the table direction will be the most useful one.

gspread/worksheet.py
lavigne958: I checked and in the source library where this idea comes from it's called `expand()` We could use the same name so it matches.

gspread/worksheet.py
```diff
@@ -3336,3 +3338,48 @@ def add_validation(
...
+        start_range: str = "A1",
```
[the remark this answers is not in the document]
lavigne958: I changed the name, I'm not too sure about how to call that range coordinates

gspread/worksheet.py
```diff
@@ -3336,3 +3338,48 @@ def add_validation(
...
+
+        Example::
+
+            values = [
+                ['', '',   '',   '', ''  ],
+                ['', 'B2', 'C2', '', 'E2'],
+                ['', 'B3', 'C3', '', 'E3'],
+                ['', ''  , ''  , '', 'E4'],
+            ]
+            >>> worksheet.expand_table(TableDirection.table, 'B2')
```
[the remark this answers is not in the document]
lavigne958: right I added an empty cell right in the middle.

gspread/worksheet.py
```diff
@@ -3336,3 +3338,48 @@ def add_validation(
...
+            [
+                ['B2', 'C2'],
+                ['B3', 'C3'],
+            ]
+
+
+        .. note::
+
+           the ``TableDirection.table`` will first look right, then look down.
+           Any empty value in the midle of the table will be ignored.
```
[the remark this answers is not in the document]
lavigne958: updated

gspread/worksheet.py
```diff
@@ -3336,3 +3338,48 @@ def add_validation(
...
+
+        :param gspread.utils.TableDirection direction: the expand direction
+        :param str start_range: the starting cell range.
+        :rtype list(list): the resulting matrix
+        """
+
+        return find_table(self.get(), direction, start_range)
```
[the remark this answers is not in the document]
lavigne958: right, I'll pass the kwargs from `expand()` to `get()`. Only the arguments changing the data.

tests/utils_test.py
```diff
@@ -503,3 +503,42 @@ def test_to_records(self):
...
+        values = [
+            ["A1", "B1", "C1", "D1"],
+            ["", "B2", "C2", "", "D2"],
+            ["", "B3", "C3", "", "D3"],
+            ["", "", "", "", "D4"],
+        ]
+
+        # table = utils.find_table(values, utils.TableDirection.table, "B2")
+        # right = utils.find_table(values, utils.TableDirection.right, "B2")
+        # down = utils.find_table(values, utils.TableDirection.down, "B2")
```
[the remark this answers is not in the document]
lavigne958: that was a leftover when I was finishing the last test, it saves me some time. 
I revert the comments

lavigne958:
> this is a nice self-contained change. good work here :)
> 
> I think we need more tests, as I foresee a lot of edge cases, and it would be good to catch those before the feature is in gspread.

I thought about it and I thought:
we need to put limits to what we want to handle here:
Do we want to handle empty cells in the middle of the table ?
Do we want to handle cases when the given table is not square ? 
 
> 
> For example, what is the expectation for expanding this table from the top left?
> 
> 1	1	-	1
> 1	1	1	1
> -	-	1	-
> 1	1	1	1

I would expect something like:
- assuming the first row in the above example is data and part of the table, top left 1 is A1 coordinate:
| 1 | 1 |
| 1 | 1 |

- assuming the first row are headers we don't take them:
| 1 | 1 |


> This is not clear to me. Also the internal (private) `utils` expanding functions are unclear what they do.
> 
> As well, we should have a `utils` test _and_ a `worksheet` test. Currently there is only `utils`
so far the method in `Worksheet` class does nothing particular, but I'll add test just in case.

alifeee:
> I thought about it and I thought:
> we need to put limits to what we want to handle here:
> Do we want to handle empty cells in the middle of the table ?
> Do we want to handle cases when the given table is not square ?

I think a reasonable implementation is "the same as what happens when you CTRL+RIGHT and CTRL+DOWN", i.e., to stop just before empty cells.

As for whether that is what other people will expect... that's another question... if people think it should work a certain way, they may think there is a bug... which is why very clear docstrings are good

> assuming the first row are headers we don't take them:
> | 1 | 1 |

wait, huh? why are we ignoring the headers? the function should just return a `List[List]]` with all items in the table starting from the called cell?

I will look back at the code now :)

gspread/utils.py
```diff
@@ -979,6 +985,116 @@ def to_records(
...
+def find_table(
+    values: List[List[str]],
+    start_range: str,
+    direction: TableDirection = TableDirection.table,
+) -> List[List[str]]:
+    """Expands a list of values based on non-null adjacent cells.
+
+    Expand can be done in 3 directions defined in :class:`~gspread.utils.TableDirection`
+
+        * ``TableDirection.right``: expands only same row as starting cell on the right side up to
```
alifeee, line 1031: "expands right until the first empty cell"

gspread/utils.py
```diff
@@ -979,6 +985,116 @@ def to_records(
...
+          first null/empty cell
+        * ``TableDirection.down``: expands only same column as starting cell to the bottom up to
```
alifeee, line 1033: "expands down until the first empty cell"

gspread/utils.py
```diff
@@ -979,6 +985,116 @@ def to_records(
...
+          first null/empty cell
+        * ``TableDirection.table``: expands in both direction, first right then down.
```
alifeee, line 1035: "expands right until the first empty cell, then down until the first empty cell"

gspread/utils.py
```diff
@@ -979,6 +985,116 @@ def to_records(
...
+
+    if direction == TableDirection.right:
+        rightMost = _expand_right(values, col, len(values[row]), row)
+        bottomMost = row + 1
+
+    if direction == TableDirection.table:
+        rightMost = _expand_right(values, col, len(values[row]), row)
+
+        checkColumn = rightMost
+        if checkColumn != 0:
```
alifeee, line 1084: is it possible for checkColumn to be 0?

gspread/worksheet.py
```diff
@@ -3336,3 +3338,57 @@ def add_validation(
         }
 
         return self.client.batch_update(self.spreadsheet_id, body)
+
+    def expand(
+        self,
+        top_left_range_name: str = "A1",
+        direction: TableDirection = TableDirection.table,
+        value_render_option: Optional[ValueRenderOption] = None,
+        date_time_render_option: Optional[DateTimeOption] = None,
```
alifeee, line 3347: I think maybe we do not want to duplicate all the options for `get`. We would rather provide the user no choice and say

"if you want to get different values, use `get` as you desire, and then use `utils.expand_table`"

as there are already missing options (e.g., `combine_merged_cells`, `pad_values`), and more may be added in future

so we remove value_render_option and date_time_render_option in favour of telling user to use `get` themself. what do you think?

gspread/worksheet.py
```diff
@@ -3336,3 +3338,57 @@ def add_validation(
...
+    ) -> List[List[str]]:
+        """Expands a cell range based on non-null adjacent cells.
+
+        Expand can be done in 3 directions defined in :class:`~gspread.utils.TableDirection`
+
+            * ``TableDirection.right``: expands only same row as starting cell on the right side up to
+              first null/empty cell
+            * ``TableDirection.down``: expands only same column as starting cell to the bottom up to
+              first null/empty cell
+            * ``TableDirection.table``: expands in both direction, first right then down.
```
alifeee, line 3357: see comments on `utils` docstring

gspread/worksheet.py
```diff
@@ -3336,3 +3338,57 @@ def add_validation(
...
+            potential empty values located in the middle of the table.
+
+        :param gspread.utils.TableDirection direction: the expand direction
+        :param str start_range: the starting cell range.
+        :rtype list(list): the resulting matrix
+        """
+
+        values = self.get(
+            value_render_option=value_render_option,
+            date_time_render_option=date_time_render_option,
```
alifeee, line 3392: what happens if `get` does not return a 2D array?

for example if it returns

<pre>
- - - d
- a
- t
- - - a
</pre>

then will functions break? should we provide `pad_values=True`?

alifeee (commented):
you are doing good work on this change :)

I have left some comments. I am a little worried that a new feature will result in something we do not consider, causing errors. But I think we will be ok now.

thanks!!!

alifeee:
I have added some new tests. In particular, they are "what should happen" tests. They are:

### 1 - missing header item

```python
        values = [
            ["A1", "", "C1", ""],
            ["A2", "B2", "C2", ""],
            ["A3", "B3", "C3", ""],
            ["", "", "", ""],
        ]
        expected_table = [
            ["A1"],
            ["A2"],
            ["A3"],
        ]
```

### 2 - missing initial cell

```python
        values = [
            ["", "B1", "C1", ""],
            ["A2", "B2", "C2", ""],
            ["A3", "B3", "C3", ""],
            ["", "", "", ""],
        ]
        expected_table = [
            ["", "B1", "C1"],
            ["A2", "B2", "C2"],
            ["A3", "B3", "C3"],
        ]
```

### 3 - missing first column item

```python
        values = [
            ["A1", "B1", "C1", ""],
            ["", "B2", "C2", ""],
            ["A3", "B3", "C3", ""],
            ["", "", "", ""],
        ]
        expected_table = [
            ["A1", "B1", "C1"],
            ["", "B2", "C2"],
            ["A3", "B3", "C3"],
        ]
```

### 4 -  missing last column item

```python
        values = [
            ["A1", "B1", "C1", ""],
            ["A2", "B2", "", ""],
            ["A3", "B3", "C3", ""],
            ["", "", "", ""],
        ]
        expected_table = [
            ["A1", "B1", "C1"],
        ]
```

## Conc.

Are these what you expect? Personally I might want case 4 to end up in the entire table, as I reckon I am more likely to have a ***full*** first column, and empty gaps in the final column.

However, if I use the `CTRL+RIGHT` and `CTRL+DOWN` shortcuts in most spreadsheet software, then the tables above are the same. So I am happy with the behaviour above.

I am ready to merge this 👍. However, we should also add it to the list of examples and documentation somewhere.

alifeee (changes requested):
include in examples docs. otherwise approve :)

lavigne958:
thank you for the extra tests ! :star: 

I am a bit confused with the tests and what I see when I try it online using keyboard controls :disappointed: 

1. in this case, I agree with you, when I run the same data set with `CTRL+RIGHT` the cell jumps to `C1`... which is not expected at all.... I believe here we should keep our behavior it makes more sense to me according to our documentation and what I would expect from the function
2. I disagree, the first cell is empty, then the array stops here, and returns an empty array :roll_eyes:  I checked online with keyboard controls and for some reasons it jumps to celle `B1` :facepalm: then when you go down it goes to `B3` (as expected)... I don't know what choice is the best here :disappointed: 
3. I completely agree with it, as expected, like the documentation says: we can't scan for empty cells in the middle of the table.
4. I completely agree with it, though online when I try, it get to `C1`, nice, then I use `CTRL+DOWN` then it jumps to `C3` :facepalm: I believe our example is more logic and provides the expected result. 

so for 1., 3. and 4. I agree with your tests and the result is what I would expect.
what is left is case 2., do you think if we start on a blank cell we should start looking right/down for the next cell if it has a value or look right/down for the first non empty cell and start from here ? :thinking:

alifeee:
I have made this sheet

https://docs.google.com/spreadsheets/d/1huMwgagFCGTMVBf5mONk_W74-qHZUnpASHuVHnU6Hsk/edit?usp=sharing

image

I think maybe we do not "do the same as CTRL+arrow keys does", as this is confusing. I think we find another rule to follow (perhaps "the same as what xlwings does")

lavigne958:
Hey thanks for the working sample ! I completely agree with you some behaviors are completely unexpected :disappointed: 

To me we should go for what feels logical to us, which seems to be what is described in the `xlwing` library:
- going right until first empty cell is found
- going then down until first empty cell is found.
- ignore empty starting cell (like you mentioned)
- table direction should be: _down then right_ as mention in their documentation here: [expand parameters](https://docs.xlwings.org/en/latest/api/range.html#id14)

does that sounds good to you ? :thinking: 
after all we can chose the implementation we like as well, regardless of what the Spreadsheet UI in a web browser does :laughing:

gspread/utils.py
```diff
@@ -979,6 +985,130 @@ def to_records(
...
+        # in case we try to look further than last row
+        if rows >= len(values):
+            return len(values) - 1
+
+        # this row is smaller than the others, just keep looking
+        if col >= len(values[rows]):
+            continue
+
+        if values[rows][col] == "":
+            return rows
```
lavigne958, line 1029: here I noticed we have rows smaller than the starting row sometimes.
I assumed we should keep looking for rows bellow, but I believe this is wrong.

we should look down for any empty value `""` OR any row smaller than starting row right ?

alifeee: I prefer to answer questions by writing tests. not sure exactly what you mean

alifeee:
I believe my preference is to "find a table" by "complete row 1 and column 1"

so, to look at these `h` columns

| h | h | h | h |
| --- | --- |  ---  | --- |
| h | - | - | - |
| h | - | - | - |
| h | - | - | - |

...and if they are empty, terminate the table there.

I have updated the tests to do this (please have a look at the tests to see if you agree). What do you think?

gspread/utils.py
lavigne958: what I mean is currently we can have a response from the API that looks like this:

||A |B|C|
|-|-| - | -|
|1|v1|v2|v3|
|2|v4|v5| |
|3|v6|v7|v8|

and currently in the code we:
1. start looking right from A1, until C1 (latest value in the row)
2. then we start to look down.
3. when we check C2, the row 2 returned by the API is smaller than the starting row 1
4. currently we just ignore it and keep looking to the row below
5. I believe that's wrong, we should stop. 

So the code `if col >= len(values[rows]):` is wrong we should not continue but instead consider it as an empty cell.

lavigne958:
> I believe my preference is to "find a table" by "complete row 1 and column 1"
> 
> so, to look at these `h` columns
> 
> h	h	h	h
> h	-	-	-
> h	-	-	-
> h	-	-	-
> ...and if they are empty, terminate the table there.
> 
> I have updated the tests to do this (please have a look at the tests to see if you agree). What do you think?

I agree with this, the only detail would be: look down first then right. 
to follow what xlwings does. it's completely arbitrary but it makes us ISO with the xlwings behavior.

gspread/utils.py
alifeee: yes. With my suggestion (look for full top row & left column), this should not matter(?)

lavigne958:
I did quite a big rework on the iteration functions.

all tests now pass, though the tests already here needed to be adjusted: 
- we now look down first
- then right
- so the expected result changes.

it works as expected, like xlwings.

alifeee:
why does xlwings use the left column and *bottom* row?

it makes more sense to me that one would desire the *top* row?

lavigne958:
I know, that's what I did first, but we agreed on making it ISO with xlwings.

I don't mind changing it, I just want us to agree on one implementation.

alifeee:
I don't think that's how xlwings works. I think xlwings uses the top row and left column.

See here for their implementation: https://github.com/xlwings/xlwings/blob/a159cc9cb7ef5f168ad0d902934404bce4a76386/xlwings/expansion.py#L31-L52

(I can't parse it with my mind fully but it looks to use the top row, not the bottom)

see also this comment: https://github.com/xlwings/xlwings/issues/557#issuecomment-242794085

lavigne958:
> I don't think that's how xlwings works. I think xlwings uses the top row and left column.
> 
> See here for their implementation: https://github.com/xlwings/xlwings/blob/a159cc9cb7ef5f168ad0d902934404bce4a76386/xlwings/expansion.py#L31-L52
> 
> (I can't parse it with my mind fully but it looks to use the top row, not the bottom)
> 
> see also this comment: [xlwings/xlwings#557 (comment)](https://github.com/xlwings/xlwings/issues/557#issuecomment-242794085)

you're right ! 
same I can't parse the code with my mind, but I understand the same thing and it makes sense. to start from the given cell, go right until empty cell, from starting cell again, go down until empty cell.

this provides the right and down boundaries then the rest is just to extract the matrix from here.

I'll update the code, expect potentially a big code change/refactoring then !

lavigne958:
## Update

- reworked the way we look for table, look right from starting cell, then look down **from** starting cell
- updated the tests suite to reflect the behavior
  - empty starting cell will return an empty list `[]`
  - empty cell on the way right or down will stop the search
  - empty cell in the **middle** of the matrix will be ignored
  - in the assert functions in the test, iterate over the expected result and expect the values to match
    - I notice tests passing when it should not because we iterate over the result and it's empty so we don't compare anything
- updated the doc a bit
- added extra checks on the size of the matrix and the starting cell coordinates in order to raise Exception if starting cell is outside the range of the given matrix.

I hope this covers any wrong inputs, the expected behavior from all (most?) users.

alifeee:
this looks all good to me!

only final thing is that I would expect a table like

| | 1 | 2 | 3 |
| --- | --- | --- | --- |
| a | 0.342 | 0.145 | 0.166 |
| b | 0.982 | 0.576 | 0.600 |
| c | 0.340 | 0.285 | 0.585 |

to expand properly, whereas it seems you suggest with

> empty starting cell will return an empty list []

that it would return as an empty list

lavigne958:
Yes to me this function will return the table starting from the given coordinates. The starting cell is part of the resulting table. so if the starting cell is indeed empty then the table will be empty.

So far that seems to be the last key point where we can't find seems to be best. Should we just put it as an option like "ignore empty starting cell" ? Of true then we follow your choice of false we follow mine. This way the user decides the behavior.

What do you think?

alifeee:
apologies for the delay

while I think a special case for a blank row 0 column 0 would be nice (e.g., for a labelled adjacency matrix), it is probably not worth the complication (e.g., if the row 0 column 0 is empty, but also row 1 column 0 and row 0 column 1 are empty, it should return `[]` not `[[""]]`)

Thus, I think we return `[]` as you say. If people want their table to be parsed, they should fill the first cell with e.g., `"-"`, or only call the function on the table contents, and use `get_row` and `get_column` (e.g.,) to get the headings.

However, there should be a test to validate this behaviour.

After that, I think this is ready to merge!

lavigne958:
I added a new commit with a test to validate the top left cell empty scenario.
In case we find a top left empty cell we return an empty list `[]`.
Is this the test you referred to in you example ? if not let me know I'll add extra tests if need be.

I added an extra test too in the basic test for find_table to check when we reach the end of the list and the end of the matrix. just in case.

alifeee:
this is good with me!

with the amount of work spent on a new feature... let's put reference to it in all the relevant places in the documentation :)

lavigne958:
> this is good with me!
> 
> with the amount of work spent on a new feature... let's put reference to it in all the relevant places in the documentation :)

great call, it's done :heavy_check_mark: 
I tried to make it minimal with all the important details.

lavigne958:
before merging I rebased the whole thing in order to solve conflicts of any.

alifeee (approved):
seems good to me ! I am assuming you have manually tested the worksheet function (as there is only a utils function)

I hope you have had fun with this one ;]

I say we can merge this.

lavigne958:
yes I did :wink: it works like a charm. I just pushed a quick fix on the docstring to make the link to the raised exception work in the documentation.

PR #1477, opened 2024-06-05
Remove passing exception as args to super in APIError

mike-flowers-airbnb:
First, thanks for this tool!

We have a pipeline where exceptions from packages such as `gspread` are pickled and stored for further processing. Unfortunately, the `APIError` exception is not able to be unpickled as the args it passed to the `Exception` super are different from what is passed into the exception itself.

The sample test extension exposes this error as seen below:
```
    def __init__(self, response: Response):
        super().__init__(self._extract_error(response))
        self.response: Response = response
>       self.error: Mapping[str, Any] = response.json()["error"]
E       AttributeError: 'dict' object has no attribute 'json'

gspread/exceptions.py:45: AttributeError
```

This PR ultimately only replaces the extracted error with the raw response passed to the kwargs. Given that `__str__` and `__repr__` are overloaded and all other usage of this exception use the other members, there is no visible impact from this change.

### DIFF
--- a/tests/worksheet_test.py
+++ b/tests/worksheet_test.py
@@ -1915,5 +1915,7 @@ class WorksheetTest(GspreadTest):
         with self.assertRaises(APIError) as ex:
             sheet.update(values="X", range_name="A1")
 
-        # Ensure that the exception is able to be pickled
-        pickle.loads(pickle.dumps(ex.exception))  # nosec
+        # Ensure that the exception is able to be pickled and unpickled
+        # Further ensure we are able to access the exception's properties after pickling
+        reloaded_exception = pickle.loads(pickle.dumps(ex.exception))  # nosec
+        self.assertEqual(reloaded_exception.args[0]["status"], "INVALID_ARGUMENT")
--- a/tests/worksheet_test.py
+++ b/tests/worksheet_test.py
@@ -1,4 +1,5 @@
 import itertools
+import pickle
 import random
 import re
 from inspect import signature
@@ -1911,4 +1912,8 @@ class WorksheetTest(GspreadTest):
             {"spreadsheetId": self.spreadsheet.id, "replies": [{}]},
         )
 
-        self.assertRaises(APIError, sheet.update, values="X", range_name="A1")
+        with self.assertRaises(APIError) as ex:
+            sheet.update(values="X", range_name="A1")
+
+        # Ensure that the exception is able to be pickled
+        pickle.loads(pickle.dumps(ex.exception))
--- a/tests/worksheet_test.py
+++ b/tests/worksheet_test.py
@@ -1,5 +1,5 @@
 import itertools
-import pickle
+import pickle  # nosec
 import random
 import re
 from inspect import signature
@@ -1916,4 +1916,4 @@ class WorksheetTest(GspreadTest):
             sheet.update(values="X", range_name="A1")
 
         # Ensure that the exception is able to be pickled
-        pickle.loads(pickle.dumps(ex.exception))
+        pickle.loads(pickle.dumps(ex.exception))  # nosec
--- a/gspread/exceptions.py
+++ b/gspread/exceptions.py
@@ -40,9 +40,10 @@ class APIError(GSpreadException):
     such as when we attempt to retrieve things that don't exist."""
 
     def __init__(self, response: Response):
-        super().__init__(response)
+        error = dict(response.json()["error"])
+        super().__init__(error)
         self.response: Response = response
-        self.error: Mapping[str, Any] = response.json()["error"]
+        self.error: Mapping[str, Any] = error
         self.code: int = self.error["code"]
 
     def __str__(self) -> str:
@@ -52,6 +53,9 @@ class APIError(GSpreadException):
 
     def __repr__(self) -> str:
         return self.__str__()
+
+    def __reduce__(self) -> tuple:
+        return self.__class__, (self.response,)
 
 
 class SpreadsheetNotFound(GSpreadException):
--- a/gspread/exceptions.py
+++ b/gspread/exceptions.py
@@ -40,7 +40,7 @@ class APIError(GSpreadException):
     such as when we attempt to retrieve things that don't exist."""
 
     def __init__(self, response: Response):
-        super().__init__(self._extract_error(response))
+        super().__init__(response)
         self.response: Response = response
         self.error: Mapping[str, Any] = response.json()["error"]
         self.code: int = self.error["code"]
--- a/gspread/exceptions.py
+++ b/gspread/exceptions.py
@@ -6,7 +6,7 @@ Exceptions used in gspread.
 
 """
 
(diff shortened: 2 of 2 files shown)


alifeee:
The lint is failing for unrelated reasons. Please bear with us while #1478 is solved :)

lavigne958:
Hi @mike-flowers-airbnb thank you for this improvement!

I need to run a couple of tests first.

As you remove the use of the function `extract_error` and we already extract it in the response object just below could you please remove the method `extract_error` please ?

mike-flowers-airbnb:
> Hi @mike-flowers-airbnb thank you for this improvement!
> 
> I need to run a couple of tests first.
> 
> As you remove the use of the function `extract_error` and we already extract it in the response object just below could you please remove the method `extract_error` please ?

Sure thing, removed the obsolete function

lavigne958:
@mike-flowers-airbnb problem should be solved now, please rebase your branch on laster commit on master branch in order to pass the CI.

mike-flowers-airbnb:
✅ Done , thanks!

lavigne958 (changes requested):
Hi I ran som regression tests. I found an issue in the current design:

- current a user is allowed to do something like this:
```python
try:
    sheet = file.del_worksheet_by_id(112341462553536987)
except gspread.exceptions.APIError as e:
    pprint(e.args[0]["status"])
```

In our current code base, this will return:
- the first argument passed to the `Exception` parent class
- this argument is a `dict`, so it will look for the key `code` in this `dict`.
- This returns the error status returned by the API.


With your current implementation we face the following error with the above code:
```
  File "/..../exception_test.py", line 10, in <module>
    print(e.args[0]["code"])
TypeError: 'Response' object is not subscriptable
```

This is because the first argument passed to the `Exception` parent class is no longer a dict but a `Response` object.

I tried the below code that would allow us to make the changes **and** keep the old behavior and prevent regression.

```python
class APIError(GSpreadException):
    """Errors coming from the API itself,
    such as when we attempt to retrieve things that don't exist."""

    def __init__(self, response: Response):
        super().__init__(response)
        self.response: Response = response
        self.error: Mapping[str, Any] = response.json()["error"]
        self.code: int = self.error["code"]

    @property
    def args(self):
        return [dict(self.error)]
```

the above `@property` on `args` will allow us to catch anyone trying to access the `args` attribute and return the error object as a dict.

@alifeee  what do you think of this _workaround_ ? 

apart from that, which could introduce a breaking change, I am fine with the changes.

mike-flowers-airbnb:
Appreciate the callout here @lavigne958 . Definitely don't want to be breaking exception catch flows.

I did some searching and found that there is actually a proper form to expose the ability for pickling to work when args don't match. I had incorrectly assumed that args were *supposed* to match the underlying `Exception` args, but it looks like there is a built-in capability to support this scenario by using [__reduce__](https://docs.python.org/3/library/pickle.html#object.__reduce__). If there's no objections, I'd propose that we use this instead as it will achieve the desired effect while not tampering with other lower-level functionality. Let me know what you think.

lavigne958:
> Appreciate the callout here @lavigne958 . Definitely don't want to be breaking exception catch flows.
> 
> I did some searching and found that there is actually a proper form to expose the ability for pickling to work when args don't match. I had incorrectly assumed that args were _supposed_ to match the underlying `Exception` args, but it looks like there is a built-in capability to support this scenario by using [**reduce**](https://docs.python.org/3/library/pickle.html#object.__reduce__). If there's no objections, I'd propose that we use this instead as it will achieve the desired effect while not tampering with other lower-level functionality. Let me know what you think.

that's a great news, if I understand correctly your point, you don't need these changes anymore from gspread right ?
If yes, then I'll be keen on making the changes your propose, but in the next major release so we can introduce the breaking change with a proper warning and users can adjust themselves. 

let me know if that's ok with you and I'll put your issue in the next major release milestone.

mike-flowers-airbnb:
Sorry I mean that I can change this PR so that the exception implements `__reduce__` and no compatibility will be lost while also introducing new pickling support. Let me add this in real quick to show.

mike-flowers-airbnb:
I just pushed a new commit to implement it. You'll notice that the `args` property is now completely backwards compatible and the exception is able to be properly pickled/unpickled. I also extended the existing test to include your regression test. LMK your thoughts.

lavigne958:
> I just pushed a new commit to implement it. You'll notice that the `args` property is now completely backwards compatible and the exception is able to be properly pickled/unpickled. I also extended the existing test to include your regression test. Let me know your thoughts, thanks.

thank you for the prompt reply, I'll have a look and test it. looks qiet good to me !

mike-flowers-airbnb:
@lavigne958 , hey there, just checking in to see if anything else is needed for this PR. Thanks.

lavigne958:
> @lavigne958 , hey there, just checking in to see if anything else is needed for this PR. Thanks.

Hi so far what is mostly needed is "time", I didn't have any time recently to take a look. 

With the coming days I will have more time to take a look at it. 

Sorry for the delay.

mike-flowers-airbnb:
> > @lavigne958 , hey there, just checking in to see if anything else is needed for this PR. Thanks.
> 
> Hi so far what is mostly needed is "time", I didn't have any time recently to take a look.
> 
> With the coming days I will have more time to take a look at it.
> 
> Sorry for the delay.

No worries, definitely understand, thanks for the update!

tests/worksheet_test.py
```diff
@@ -1911,4 +1912,10 @@ def test_add_validation(self):
...
         )
 
-        self.assertRaises(APIError, sheet.update, values="X", range_name="A1")
+        with self.assertRaises(APIError) as ex:
+            sheet.update(values="X", range_name="A1")
+
+        # Ensure that the exception is able to be pickled and unpickled
+        # Further ensure we are able to access the exception's properties after pickling
+        reloaded_exception = pickle.loads(pickle.dumps(ex.exception))  # nosec
+        self.assertEqual(reloaded_exception.args[0]["status"], "INVALID_ARGUMENT")
```
lavigne958, line 1921: don't you need to indent this part of the code so it's under the `with` statement, otherwise you argument `ex` is not reliably accessible after the `with` statement exists ?

lavigne958 (commented):
looks good to me, with regression test as extra ! thank you.
Only a small detail about the indentation just in case then: looks good to merge

tests/worksheet_test.py
mike-flowers-airbnb: Nah, the context object `ex` remains populated after the with and the code execution actually stops inside the `with` block after the `sheet.update` as it raises the error.
See [official docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) for their example on this which is what I'm following.

lavigne958: Alright to me.

lavigne958 (approved):
LGTM

alifeee (approved):

PR #1481, opened 2024-06-13
ignore jinja CVE

### LINKED ISSUE #1478: lint failure
The linting is failing now

<https://github.com/burnash/gspread/actions>

What has changed? I am not sure.

Here is a working workflow <https://github.com/burnash/gspread/actions/runs/9259695428/job/25472114091>

[working.txt](https://github.com/user-attachments/files/15746025/working.txt)

Here is a broken one <https://github.com/burnash/gspread/actions/runs/9259695827/job/25472115123>

[workingnot.txt](https://github.com/user-attachments/files/15746026/workingnot.txt)

I don't see a big difference in the lint step

image

This requires more investigating.

lavigne958:
fix CI by skiping unwanted CVE report.
fix CI by updating typing on google auth library

closes #1478

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -30,6 +30,6 @@ jobs:
       - run: tox -e lint
       - run: tox -e py
       - run: shopt -s globstar && pyupgrade --py3-only **/*.py # --py36-plus
-      - run: safety check -i 42559 -i 62044 -i 67599 # pip version issue, we don't care about it, we don't ship it
+      - run: safety check -i 42559 -i 62044 -i 67599 -i 70612 # pip version issue, we don't care about it, we don't ship it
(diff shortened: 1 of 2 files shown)


lavigne958:
pushed force to update commit message

alifeee:
is `safety check` worth using if we ignore it every time?

alifeee (approved):

lavigne958:
> is `safety check` worth using if we ignore it every time?

it is for other matters, mostly for the google API we use, its current version being used and other potential libraries like `requests` etc.
for sure our use of it is very very limited and we could remove it, I'm fine with it. 

May be let's find-out online what it really brings, and if we remove it what do we risk ? based on that we'll decide if we remove it

PR #1487, opened 2024-06-24
Added a range option to `Worksheet.get_notes` [Issue #1482]

### LINKED ISSUE #1482: get_notes from a range
I think get_notes would be more useful if if could be applied to a range in addition to a worksheet.

muddi900:
Fixes #1482


alifeee:
hi! thanks for the change! your code was all good, but the test cassette was breaking things. This is common, and what I usually do is completely delete the JSON test cassette and re-create it, effectively running:

```bash
rm tests\cassettes\WorksheetTest.test_get_notes.json
GS_CREDS_FILENAME="tests/creds.json"
GS_RECORD_MODE="all"
pytest -k "test_get_notes" -v -s tests/ 
# check it worked
GS_RECORD_MODE="none"
pytest -k "test_get_notes" -v -s tests/ 
```

I have reformatted the test a little so it is easier to read for me, and changed `range` to `grid_range`, as `range` is a python built-in keyword that we should not change.

Also, in testing, I noticed that `get_notes` did not work with worksheets that were not the primary sheet, so I have fixed this (see [16429af](https://github.com/burnash/gspread/pull/1487/commits/16429af50bbf563e0038fc2a58ce0772c1c4d6d8) and [983dcb5](https://github.com/burnash/gspread/pull/1487/commits/983dcb5b7989674d7b1b88da374e5f464d2d18bf) above) too.

again thanks for the change! I will let @lavigne958 read this and merge when happy with it :)

nbwzx:
> Also, in testing, I noticed that `get_notes` did not work with worksheets that were not the primary sheet, so I have fixed this (see [16429af](https://github.com/burnash/gspread/pull/1487/commits/16429af50bbf563e0038fc2a58ce0772c1c4d6d8) and [983dcb5](https://github.com/burnash/gspread/pull/1487/commits/983dcb5b7989674d7b1b88da374e5f464d2d18bf) above) too.

Wait is that wrong?

For example, there are three worksheets (Sheet1, Sheet2, Sheet3). 

`res["sheets"][self.index]["data"][0].get("rowData", [{}])` returns the notes of current worksheet. 
self.spreadsheet.worksheets()[0].get_notes() get the notes of Sheet1
self.spreadsheet.worksheets()[1].get_notes() get the notes of Sheet2
self.spreadsheet.worksheets()[2].get_notes() get the notes of Sheet3

If you change `self.index` to `0`, you can only get the notes of the primary sheet (Sheet1), 
self.spreadsheet.worksheets()[0].get_notes() get the notes of Sheet1
self.spreadsheet.worksheets()[1].get_notes() get the notes of Sheet1
self.spreadsheet.worksheets()[2].get_notes() get the notes of Sheet1
how can you get the notes of other worksheets (Sheet2, Sheet3)?

alifeee:
> Wait is that wrong?

As it turns out... yes and no! ;]

When you provide a range, i.e., here

https://github.com/muddi900/gspread/blob/d4a57746816ad097d56969cfab4c9c12a483dff6/gspread/worksheet.py#L2657-L2658

Then the returned results has only one sheet. Printing `res`, we get

```python
> get_notes()
# res = 
{'sheets':
  [
    {'data': [{}]},
    {'data': [{'rowData': [{'values': [{'note': 'the first time'}]}, {}, {'values': [{}, {'note': 'two sheets'}]}]}]}
  ]
}
> get_notes("A2:C3")
# res = 
{'sheets':
  [
    {'data': [{'rowData': [{}, {'values': [{}, {'note': 'two sheets'}]}]}]}
  ]
}
```

So in the initial case, we do need to index it with `worksheet.index`, i.e, `1`

In the second case, we need to use `0`. I had spotted the latter case, which is why I made my change. However, I didn't make a very good test to be able to spot the reason it had changed.

I have made the test clearer (split it out) and 'fixed' it by always providing `worksheet.title`, so we can only ever expect one sheet in the response.

Thanks for spotting @nbwzx ! :)

alifeee:
sorry for jumping on your PR @muddi900. wouldn't have happened without you ;] so thanks for proposing the change!

will let @lavigne958 review and with an approval I will merge, and this can be included in 6.2.0 :)

muddi900:
No need to apologize. Thank you for the help.

nbwzx:
@alifeee 
I tried your code on one of my sheets. The code `get_notes()` raises an error: gspread.exceptions.APIError: APIError: [400]: Range (Introduction!UF3) exceeds grid limits. Max rows: 966, max columns: 27.
But the code `get_notes()` runs well before this PR, even on a very big sheet.
I think the modifications about `get_notes` when `grid_range=None` reduces the limitation significantly.
So in my opinion, When grid_range is not given, using the previous code is a better way.

muddi900:
Do you have a public big sheet I can test this on?

nbwzx:
> Do you have a public big sheet I can test this on?

The one I mentioned is

https://docs.google.com/spreadsheets/d/158F-jyu8ld8kbdD4I_Lqy4fsX5vvp_MQOq7Ofg8NdSI/

on Sheet name `UF3`.

lavigne958 (changes requested):

lavigne958:
Thank you everyone for the hard work and the all the edge case testing ! we managed to catch a very particular use case that we would not see otherwise :muscle: 

I tried it and it works fine, it is well explained that the resulting matrix is not square.
it's quite complicated for a novice to use the example with `fill_gaps` I believe, should we simply offer an option to fill the gaps for the end user ? :thinking: 

something like:
- option name: `auto_fill_gaps`
- option type: bool
- option default value: `False` (in order to prevent breaking change_

that requires us to do some computation based on the resulting matrix, meaning on a big big sheet we need to iterate over all rows to find the longest row before iterating again on all rows to _expand_ them.

I'm fine with it, what do you guys think about it ?

gspread/worksheet.py
```diff
@@ -2631,21 +2636,36 @@ def get_notes(self, default_empty_value: Optional[str] = "") -> List[List[str]]:
...
+                ["", "B2"]
+            ]
         """
-        params: ParamsType = {"fields": "sheets.data.rowData.values.note"}
+        params: ParamsType = {
+            "fields": "sheets.data.rowData.values.note",
+            "ranges": (
+                self.title
+                if grid_range is None
+                else absolute_range_name(self.title, grid_range)
```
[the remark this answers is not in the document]
nbwzx: Nice explanation on Introduction!UF3. I was confused about it before your comment.

nbwzx:
I agree. I think it is not convenient for users to copy the relevant code from the documentation.

muddi900:
@alifeee Sorry for the `git reset` shenanigans. I missed @lavigne958 comment, and tried to redo everything.

I have incorporated the changes suggested in the comment.

alifeee:
> @alifeee Sorry for the git reset shenanigans. I missed @lavigne958 comment, and tried to redo everything.

no worries :) I have refreshed the test cassette (by deleting `tests/cassettes/WorksheetTest.test_get_notes.json` and re-running the test with `GS_RECORD_MODE=all`)

Also, the `git reset` has removed the extra test that I made for 2nd sheet notes. Please could you add this back? You can see it in here:

https://github.com/muddi900/gspread/compare/add/notes-range...burnash:gspread:add/notes-range-tests

.github/CONTRIBUTING.md
```diff
@@ -110,23 +112,23 @@ In the following cases, you must record new HTTP requests:
 - an existing test is updated and does a new HTTP request
 - gspread is updated and does a new HTTP request
 
-### Run test, capturing *all* HTTP requests
+### Run test, capturing _all_ HTTP requests
```
lavigne958, line 115: why change this part ?

.github/CONTRIBUTING.md
```diff
@@ -110,23 +112,23 @@ In the following cases, you must record new HTTP requests:
...
 
 In some cases if the test suite can't record new episodes, or it can't replay them offline, you can run a complete update of the cassettes.
 
 ```bash
 GS_CREDS_FILENAME=<./YOUR_CREDS.json> GS_RECORD_MODE=all tox -e py
 ```
 
-### Run test, capturing *only new* HTTP requests
+### Run test, capturing _only new_ HTTP requests
```
lavigne958, line 123: same here why change this part ?

lavigne958 (commented):
the changes are good.
The CI is failing due to outdated/missing cassette.
Please remove the recorded cassette: `tests/cassettes/WorksheetTest.test_get_notes.json` and record the cassettes again.

You can use the value `new_episodes` to only record the missing cassette.

.github/CONTRIBUTING.md
muddi900: I don't know why that happened.

lavigne958: Bo worries, probably "auto-format" in your Editor. 
Juste revert the changes and we'll be fine.

lavigne958 (commented):
your tests are still failing, please remove the file `tests/cassettes/WorksheetTest.test_get_notes.json` and run the test with the following command: `GS_RECORD_MODE='new_episodes' CREDS_FILENAME=<path to credentials> tox -e py`

alifeee:
the test `test_get_notes_2nd_sheet` is still missing, please see https://github.com/burnash/gspread/pull/1487#issuecomment-2210964256 for context :)

tests/worksheet_test.py
```diff
@@ -1631,20 +1631,64 @@ def test_worksheet_notes(self):
...
         all_notes = w.get_notes()
 
         self.assertEqual(empty_notes, [[]])
-        self.assertEqual(all_notes, notes_array)
+        self.assertEqual(all_notes, expected_notes)
+        self.assertEqual(range_notes, expected_range_notes)
+
+    @pytest.mark.vcr()
+    def test_get_notes_2nd_sheet(self):
+        self.spreadsheet.worksheets()[0]
```
lavigne958, line 1664: I feel like this line is a no-op

muddi900:
So should autofill be a different issue. Because that would be a rewrite of the method.

lavigne958:
> So should autofill be a different issue. Because that would be a rewrite of the method.

I am sorry I don't understand your comment, could you please elaborate ?

muddi900:
> Thank you everyone for the hard work and the all the edge case testing ! we managed to catch a very particular use case that we would not see otherwise 💪
> 
> I tried it and it works fine, it is well explained that the resulting matrix is not square. it's quite complicated for a novice to use the example with `fill_gaps` I believe, should we simply offer an option to fill the gaps for the end user ? 🤔
> 
> something like:
> 
>     * option name: `auto_fill_gaps`
> 
>     * option type: bool
> 
>     * option default value: `False` (in order to prevent breaking change_
> 
> 
> that requires us to do some computation based on the resulting matrix, meaning on a big big sheet we need to iterate over all rows to find the longest row before iterating again on all rows to _expand_ them.
> 
> I'm fine with it, what do you guys think about it ?

I was referring to this

lavigne958:
I see, thank you.
we can make it 2 different issues, though we can't make a release between the time each issue is merged, we want to provide the feature once and not provide the feature then provide a different way to use it.

lavigne958:
I come back to the point about `fill_gaps`, the current feature requires the user to call `fill_gaps`. so for now we can't change the the output, we can merge this as is, once the cassettes have been updated and the no-op line has been removed.

If you don't have time or struggle with the cassettes @muddi900, just let me know and I'll cherry-pick you commits, you'll be mentioned as contributor still and we'll merge these changes soon.

lavigne958 (approved):

lavigne958:
Thanks @muddi900 for this contribution !

I fixed the missing cassettes and the small conflict with master branch, this is a go for mege.

PR #1491, opened 2024-07-16
Export only one worksheet by it's Id from all document added

ShevchenkoVadim:
Added the ability to export one page from the entire document by ID

### DIFF
--- a/gspread/http_client.py
+++ b/gspread/http_client.py
@@ -39,6 +39,7 @@ from .urls import (
     SPREADSHEET_VALUES_BATCH_URL,
     SPREADSHEET_VALUES_CLEAR_URL,
     SPREADSHEET_VALUES_URL,
+    SPREADSHEET_DRIVE_URL,
 )
 from .utils import ExportFormat, convert_credentials, quote
 
@@ -352,6 +353,57 @@ class HTTPClient:
             raise UnSupportedExportFormat
 
         url = "{}/{}/export".format(DRIVE_FILES_API_V3_URL, file_id)
+
+        params: ParamsType = {"mimeType": format}
+
+        r = self.request("get", url, params=params)
+        return r.content
+
+    def export_worksheet(self, file_id, sheet_id: str, format: str = ExportFormat.PDF) -> bytes:
+        """Export the worksheet in the given format.
+
+        :param str file_id: The key of the spreadsheet to export
+
+        :param str sheet_id: The key of the worksheet to export
+
+        :param str format: The format of the resulting file.
+            Possible values are:
+
+                * ``ExportFormat.PDF``
+                * ``ExportFormat.EXCEL``
+                * ``ExportFormat.CSV``
+                * ``ExportFormat.OPEN_OFFICE_SHEET``
+                * ``ExportFormat.TSV``
+                * ``ExportFormat.ZIPPED_HTML``
+
+            See `ExportFormat`_ in the Drive API.
+
+        :type format: :class:`~gspread.utils.ExportFormat`
+
+        :returns bytes: The content of the exported file.
+
+        .. _ExportFormat: https://developers.google.com/drive/api/guides/ref-export-formats
+        """
+
+        if format not in ExportFormat:
+            raise UnSupportedExportFormat
+
+        if format not in ExportFormat:
+            raise UnSupportedExportFormat
+
+        format_str = "pdf"
+        if format == ExportFormat.PDF:
+            format_str = "pdf"
+        elif format == ExportFormat.CSV:
+            format_str = "csv"
+        elif format == ExportFormat.ZIPPED_HTML:
+            format_str = "zip"
+        elif format == ExportFormat.EXCEL:
+            format_str = "xlsx"
+        elif format == ExportFormat.TSV:
+            format_str = "tsv"
+
+        url = "{}/export?gid={}&format={}".format(SPREADSHEET_DRIVE_URL % file_id, sheet_id, format_str)
 
         params: ParamsType = {"mimeType": format}
 
--- a/gspread/client.py
+++ b/gspread/client.py
@@ -250,6 +250,33 @@ class Client:
 
         return self.http_client.export(file_id=file_id, format=format)
 
+    def export_worksheet(self, file_id, sheet_id: str, format: str = ExportFormat.PDF) -> bytes:
+        """Export the spreadsheet in the given format.
+
+                :param str file_id: The key of the spreadsheet to export
+
+                :param str sheet_id: The key of the worksheet to export
+
+                :param str format: The format of the resulting file.
+                    Possible values are:
+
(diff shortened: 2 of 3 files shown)


gspread/http_client.py
```diff
@@ -358,6 +359,57 @@ def export(self, file_id: str, format: str = ExportFormat.PDF) -> bytes:
...
+        :returns bytes: The content of the exported file.
+
+        .. _ExportFormat: https://developers.google.com/drive/api/guides/ref-export-formats
+        """
+
+        if format not in ExportFormat:
+            raise UnSupportedExportFormat
+
+        if format not in ExportFormat:
+            raise UnSupportedExportFormat
```
lavigne958, line 392: this is a duplicate

gspread/http_client.py
```diff
@@ -358,6 +359,57 @@ def export(self, file_id: str, format: str = ExportFormat.PDF) -> bytes:
...
+        if format == ExportFormat.PDF:
+            format_str = "pdf"
+        elif format == ExportFormat.CSV:
+            format_str = "csv"
+        elif format == ExportFormat.ZIPPED_HTML:
+            format_str = "zip"
+        elif format == ExportFormat.EXCEL:
+            format_str = "xlsx"
+        elif format == ExportFormat.TSV:
+            format_str = "tsv"
```
lavigne958, line 404: formats are `StrEnum` objects. so you can use the value of the Enum and it will be in string format.

gspread/http_client.py
```diff
@@ -358,6 +359,57 @@ def export(self, file_id: str, format: str = ExportFormat.PDF) -> bytes:
...
+
+        url = "{}/export?gid={}&format={}".format(SPREADSHEET_DRIVE_URL % file_id, sheet_id, format_str)
```
lavigne958, line 406: are you sure this is the right URL ? I did not find it on the google API reference documentation. can you please send me the link with the API reference ?

usually this is URL is used only to return the URL for a user to use in web browser.

gspread/http_client.py
ShevchenkoVadim: You are right. My bad. I'll fix it.

gspread/http_client.py
ShevchenkoVadim: Yeah, I'm sure! I'm testing it in my home project.

alifeee:
you can run these commands to check that the workflow will pass :)

```bash
pip install -r lint-requirements.txt
tox -e lint
# optionally but you have not changed tests/docs
pip install -r docs/requirements.txt
pip install -r test-requirements.txt
tox -e doc # build docs
tox -e py # run tests
```

is this a feature that you think is possible to add a test for, or not?

gspread/http_client.py
```diff
@@ -358,6 +359,54 @@ def export(self, file_id: str, format: str = ExportFormat.PDF) -> bytes:
...
+
+        .. _ExportFormat: https://developers.google.com/drive/api/guides/ref-export-formats
+        """
+
+        if format not in ExportFormat:
+            raise UnSupportedExportFormat
+
+        format_str = "pdf"
+        if format == ExportFormat.PDF:
+            format_str = "pdf"
```
lavigne958, line 393: if the default value is PDf, then you don't have to check if it's PDF in `format`

gspread/http_client.py
lavigne958: never mind we use the mimetype as value so it won't work.

gspread/http_client.py
```diff
@@ -358,6 +359,54 @@ def export(self, file_id: str, format: str = ExportFormat.PDF) -> bytes:
...
+        elif format == ExportFormat.CSV:
+            format_str = "csv"
+        elif format == ExportFormat.ZIPPED_HTML:
+            format_str = "zip"
```
lavigne958, line 397: this one does not work.
as well as you don't handle the value: `OPEN_OFFICE_SHEET` which results as `PDF` due to default value set at the start of the `if`s here.

lavigne958 (commented):
the list of possible formats are slightly different from the export format from the DriveAPI.
I suggest we create a new Enum list with the appropriate Names and Values.

I am not too keen on introducing this feature for one reason:
- in any error scenario we receive an HTML error page.
- the whole API is made to handle JSON response in all error scenarios, so here we fail to read the JSON error response, because its HTML format.
- This is because the URL we use is originally made for humans to use on a web browser.

@alifeee  what do you think of this detail ?

alifeee:
> @alifeee what do you think of this detail ?

I'm not fully sure why this feature is desired. What can `gspread` currently do, and what is missing?

Could someone provide a (pseudo)code snippet example of what is desired with this function?

lavigne958:
> > @alifeee what do you think of this detail ?
> 
> I'm not fully sure why this feature is desired. What can `gspread` currently do, and what is missing?
I believe it's desired when you wish to export (to PDF, to CSV, ...) a single sheet instead of a the whole spreadsheet (which creates multiple pages PDF for example).
> 
> Could someone provide a (pseudo)code snippet example of what is desired with this function?

sure, I tried it, here it is :upside_down_face: 

```python
client = gspread.service_account()
file = client.open("xxxxxxxx")
res = file.sheet1.export(gspread.utils.ExportFormat.PDF)
with open("result.pdf", "wb") as fp:
    fp.write(res)
```

My major concerns are:
1. the error handling, which returns HTML code, and not a JSON
2. the exported format that are not fully covered in this MR. (this is easily fixed with a new Enum)

lavigne958:
Hi I thought about this feature and I really don't like the use of an HTML base API. that is dedicated to server content for humans on a web browser and mostly because we can't handle errors from this URL. 

it should not be part of gspread unfortunately until the API allows us to export a single sheet.

For that matter I still want to help you and I found a way for you to download your spreadsheet and extract a single sheet using this pure python library that can handle PDFs: https://github.com/py-pdf/pypdf 

You can find the appropriate documentation here to read a pdf, choose  page then write that page only to a new PDF file.
https://pypdf.readthedocs.io/en/stable/user/cropping-and-transforming.html#cropping-and-transforming-pdfs

With this solution that can solve your problem I suggest we don't merge this feature but I still want to wait for @alifeee opinion on this, so what do you think ?

alifeee:
I see it being quite useful. I don't know what exists already to export a sheet (say, if you want to download PDF as a file). If you want a CSV or TSV, then you can probably just use `worksheet.get` just fine.

But, perhaps if you want a PDF, you would want to customise how it looks, so you would want to use the GUI on the Google Sheets web browser anyway.

As for errors: I do not know what errors exist when trying to export, apart from an incorrect format string (or if Google changes which are accepted)

So, in the end: I see this adding a nice feature, but at a cost of bad errors if anything goes wrong, which is a maintenance burden cost. I say we could leave this open, or close it without merging, and if other people come requesting the same feature, then we can reconsider.?

Issue #677, opened 2019-05-02
Feature Request: Insert record


thp44:
get_all_records() 
return nicely formatted dictionaries (thanks!)

**set_records()** (new feature)
inset multiple in list dictionaries to g-spread (values under right heading) 

**set_record()** (new feature)
inset one dictionary to g-spread (values under right heading)

muddi900:
This would be simple to implement, but it would be very slow.

We can add a mapping of the columns when the `get_all_records` are generated to the index.

Most of the work would be validation and testing.

alifeee:
The way I could see this feature working is similar to Python's [CSV DictWriter](https://docs.python.org/3/library/csv.html#csv.DictWriter) class

```python
def append_records(
  headers: List[str],
  records: List[Dict[str, str]],
  default_blank: Any = None,
  ignore_extra_dict_items: bool = False,
  # extra kwargs?
)
  # use a provided "headers" array, probably obtained with worksheet.row_values(1)
  # new_rows = []
  # for each object in records
    # for each header in headers
      # if header not in headers, raise error! (unless kwarg)
      # if header not in object, raise error! (unless default blank)
      # new_rows[i].append(object[header])
  # use worksheet.append_rows to add new_rows to worksheet
```

and `append_record` (singular) would call `append_records` with a wrapped record, like `[record]` (like `append_row` and `appeend_rows`)

Questions I would have:

- should we provide value input option kwargs?
- should we only offer starting from column 1? (probably easiest)

> This would be simple to implement, but it would be very slow.

slow? not sure what you mean

> We can add a mapping of the columns when the get_all_records are generated to the index.

also not sure what you mean

if you desire to implement it, I think this could be a nice feature @muddi900 :)

lavigne958:
I agree with your code @alifeee , small detail here, remember that `data` is `List[Dict[str,str]]` as we have a dict per row.
it looks good and the slowness here is depending on the length of `data`, with 1M rows of course it would take some time to run. Though I don't think a sheet can hold 1M rows :thinking: (may be it does ? )

so far we need to wrap the above code in a `for` loop iterating on `data` and run that for each dict found in the list `data`.
it takes sometime to run if we have a lot of rows in `data` but nothing to long I believe.

alifeee:
yes. my example, effectively, was an implementation of `set_record` (singular `data`), instead of `set_records` (plural `data`)

I think slowness will not be a problem. We will see if it is after making the code.

lavigne958:
[...]

PR #1494, opened 2024-07-21
Added `set_records` and `set_record` to update sheet via records.[Issue #677]

muddi900:


alifeee:
thanks for the implementation!

I find the test hard to parse, would it be possible to write it with very explicit objects? e.g.,

```python
table_before = [
  ["header 1", "header 2"]
  ["value 1",  "value 2"]
]
table_after = [
  ["header 1",    "header 2"]
  ["value 1",     "value 2"]
  ["new value 1", "new value 2"]
]
```

gspread/worksheet.py
```diff
@@ -606,6 +618,150 @@ def get_all_records(
...
+                raise GSpreadException("Extra headers found in the data set")
+
+            insert_row = []
+            for col in cols:
+                insert_row.append(row.get(col, default_blank))
+            insert_rows.append(insert_row)
+
+        self.append_rows(
+            insert_rows,
+            value_input_option=ValueInputOption.user_entered,
```
lavigne958, line 654: why should we force the user to use that valueInputOption type ? :thinking:

gspread/worksheet.py
```diff
@@ -606,6 +618,150 @@ def get_all_records(
...
+
+            ins_row = []
+            for col in cols:
+                ins_row.append(row.get(col, default_blank))
+            insert_rows.append(ins_row)
+
+        self.insert_rows(
+            insert_rows,
+            row=insert_row,
+            value_input_option=ValueInputOption.user_entered,
```
lavigne958, line 726: same here

gspread/worksheet.py
```diff
@@ -197,6 +197,17 @@ def __init__(
...
+
+    @property
+    def column_headers(self) -> List[str]:
+        if not self._column_headers:
+            self._column_headers = self.row_values(1)
+        return self._column_headers
+
+    @column_headers.setter
+    def column_headers(self, value: List[str]) -> None:
+        self._column_headers = value
```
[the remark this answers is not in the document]
muddi900: Fixed this as well

PR #1498, opened 2024-08-02
Created a `batch_merge` function [Issue #1473]

### LINKED ISSUE #1473: Feature request. Batch Merge in worksheet class
I am using gspread to create quite a large spreadsheet in a particular format. One of the things that is very present in the worksheet are merged cells. The `batch_update`, `batch_format`, and `batch_get` have been invaluable to me. However, there is no way from the worksheet side (that i can see) to do a batch merge.

I understand that logically merging cells should be done before you insert data. One thing that i find myself doing is having a long list of `merge_cells` at the beginning of my script, which sets up all the merges. However, each makes an API request.

I would like a `batch_merge` method defined in the `Worksheet class` that takes in a list of ranges and uses the `batch_update` method of the Google Sheets API, resulting in only 1 API request. It would most likely need to append multiple `mergeCells` to a request body.

Currently one of the ways i get around it by making the request myself and then using the `Speadsheet.batch_update` method. However, it would be better if i could do it through some sort of abstraction and keep everything in the `Worksheet` class

**Additional context**
N/A

I am more than happy to attempt to write the method on my own and create a pull request. I understand it is a bit of a niche issue

muddi900:
Fixes #1473

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2611,6 +2611,20 @@ class Worksheet:
         merges: List[Dict[Literal["range", "mergeType"], Union[str | MergeType]]],
         merge_type: MergeType = MergeType.all,
     ) -> Any:
+        """Merge multiple ranges at the same time.
+
+        :param merges: list of dictionaries with the ranges(is A1-notation), and
+            an optional ``MergeType`` field.
+            See `MergeType`_ in the Sheets API reference.
+        :type merges: List[Dict[Literal["range", "mergeType"], Union[str | MergeType]]]
+        :params merge_type: (optional) default ``MergeType`` for all merges missing the merges.
+            defaults to ``MergeType.all``.
+        :type merge_type: ``MergeType``
+
+        :returns: The body of the request repsonse.
+        :rtype: dict
+        """
+
         requests = [
             {
                 "merges": {
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2606,6 +2606,23 @@ class Worksheet:
 
         return self.client.batch_update(self.spreadsheet_id, body)
 
+    def batch_merge(
+        self,
+        merges: List[Dict[Literal["range", "mergeType"], Union[str | MergeType]]],
+        merge_type: MergeType = MergeType.all,
+    ) -> Any:
+        requests = [
+            {
+                "merges": {
+                    "range": a1_range_to_grid_range(i["range"], self.id),
+                    "mergeType": i.get("mergeType", merge_type),
+                }
+            }
+            for i in merges
+        ]
+
+        return self.client.batch_update(self.spreadsheet_id, {"requests": requests})
+
     def get_notes(self, default_empty_value: Optional[str] = "") -> List[List[str]]:
         """Returns a list of lists containing all notes in the sheet.
 
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2608,7 +2608,7 @@ class Worksheet:
 
     def batch_merge(
         self,
-        merges: List[Dict[Literal["range", "mergeType"], Union[str | MergeType]]],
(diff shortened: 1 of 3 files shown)


lavigne958:
thanks for this contribution ! as mentioned in the associated issue the input format should be simpler. see last comment: https://github.com/burnash/gspread/issues/1473#issuecomment-2295227936

alifeee:
looks wonderful ;)

this should have a test

alifeee:
@muddi900 I have updated your test and cassette

the function is called like 


```python
       sheet.batch_merge(
            {
                "A1:B2": utils.MergeType.merge_all,
                "C2:D2": utils.MergeType.merge_all,
                "C3:C4": utils.MergeType.merge_all,
            }
        )
```

The normal merge function (I think) can be called like (i.e., without merge type, uses default `merge_all`)

```bash
sheet.merge("A1:B2")
```

It would be nice if `batch_merge` could be called without specifying the merge type for each range, as I imagine most people will want to use `merge_all` and not `merge_column` or `merge_row`, so it would be nice to just use the default, something like

```python
sheet.batch_merge(
  [
    "A1:B2",
    "C2:D2",
    "C3:C4"
  ]
)
```

I think this is what @muddi900 and @lavigne958 were discussing in #1473 with regard to the type of the argument

What do you both think ?

muddi900:
I, as a user, would prefer my implementation, as that would not require a merge type declaration for each merge. And it would allow adding a merge type for a specific case.

Thank you.

alifeee:
I agree @muddi900. What do you think @lavigne958?

I see three options

1

```python
sheet.batch_merge(
  [
    "A1:B2",
    "C2:D2",
    "C3:C4"
  ]
)
```

2

```python
sheet.batch_merge(
  [
    {"range": "A1:B2"},
    {"range": "C2:D2"},
    {"range": "C3:C4", "merge_type": utils.MergeType.merge_all}
  ]
)
```

3

```python
sheet.batch_merge(
            {
                "A1:B2": utils.MergeType.merge_all,
                "C2:D2": utils.MergeType.merge_all,
                "C3:C4": utils.MergeType.merge_all,
            }
        )
```

Under the assumption that most of the time, a user will not care to specify the merge type...

I think 1 is the most user friendly option, but I'm not sure how to specify merge types (a second array, and zip them? I don't like it so much...)

I think 2 works, but is still quite complicated.

I think 3 is the most complicated as it does not allow missing out merge type.

I suggest some experimental options

4

```python
sheet.batch_merge(
  [
    "A1:B2",
    ("C2:D2", utils.MergeType.merge_all),
    "C3:C4"
  ]
)
```

5

```python
sheet.batch_merge(
  [
    "A1:B2",
     {"range": "C3:C4", "merge_type": utils.MergeType.merge_all},
    "C3:C4"
  ]
)
```

I think my overall preference is 4, as it does not require knowledge of what strings to use (i.e., `"range"` and `"merge_type"`) nor to specify merge_type for every range.

muddi900:
It would be better if the list has a unified type.


the reason I opted for `dict` instead of `tup` because the latter may lead to user errors.

lavigne958:
Hi I agree with @muddi900 , mixed inputs that can be `string` or `tuple(string, MergeType)` is confusing.

I still think option 3 is the most explicit one with the least necessary inputs to pass (a `string` for the key and a `MergeType` for the value). 

I understand that option 2 could be simpler for the users, it allows them to reduce code and it's very easy for us to set the default value when reading the dictionary.

Option 1 to me does not provide enough flexibility for the user, if any user needs to do some column or row merge then the feature is useless for that user. 

I would accept option 2 if that suits everyone ?

alifeee (approved):
hi @muddi900, thanks again for the work !! I made some small changes and created the test cassette

it looks good to me now :)

I will let @lavigne958 approve and merge ;]

PR #1505, opened 2024-08-19
fix: fix type annotation for default_blank

hiro-o918:
fix type annotation

### DIFF
--- a/gspread/utils.py
+++ b/gspread/utils.py
@@ -306,7 +306,7 @@ def numericise_all(
     :param list values: Input row
     :param bool empty2zero: (optional) Whether or not to return empty cells
         as 0 (zero). Defaults to ``False``.
-    :param str default_blank: Which value to use for blank cells,
+    :param Any default_blank: Which value to use for blank cells,
         defaults to empty string.
     :param bool allow_underscores_in_numeric_literals: Whether or not to allow
         visual underscores in numeric literals
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -529,7 +529,7 @@ class Worksheet:
             be rendered in the output. See `ValueRenderOption`_ in
             the Sheets API.
         :type value_render_option: :class:`~gspread.utils.ValueRenderOption`
-        :param str default_blank: (optional) Determines which value to use for
+        :param Any default_blank: (optional) Determines which value to use for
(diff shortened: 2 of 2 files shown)


lavigne958:
Hi thank you for contributing, I'm sorry but I don't understand the issue behind your changes.

The input `default_blank` is here to set a value in case of blank cells, when a user wants a specific value. This is typed string so gspread can safely work knowing that all input values are strings.

Could you please explain your changes ?

hiro-o918:
@lavigne958 
Sorry for slow responding.
I found `default_black` passed into [`numericise`](https://github.com/burnash/gspread/blob/60bc77f73501d5d5fe40e6ed179e092c4a759aea/gspread/utils.py#L224) finally.
This seems that the value can accept `Any` value.

For my case, I want to pass [`pd.NA`](https://pandas.pydata.org/docs/reference/api/pandas.NA.html#pandas.NA) as a default value for empty data which is not `str` type.

lavigne958:
> @lavigne958 Sorry for slow responding. I found `default_black` passed into [`numericise`](https://github.com/burnash/gspread/blob/60bc77f73501d5d5fe40e6ed179e092c4a759aea/gspread/utils.py#L224) finally. This seems that the value can accept `Any` value.
> 
> For my case, I want to pass [`pd.NA`](https://pandas.pydata.org/docs/reference/api/pandas.NA.html#pandas.NA) as a default value for empty data which is not `str` type.

I see, this is a different issue then, you request the ability to use a different type and you know the util functions `numericise` will work. though the upper method `get_all_records` will block you from using it.

Then I agree, we need to double check it does not break anything but if this works fine, I'm fine with it.

@alifeee any objection ?

alifeee:
this sounds good to me

only thing to remember is that multiple functions use `default_blank` so they should all be updated

not sure of the need for a test since it's just typing.

hiro-o918:
@lavigne958 @alifeee 
Thank you for your reviews.
I found there is no additional type annotations `str` for `default_blank`
please double check just in case

alifeee:
I think you are correct! Then this is good with me :)

alifeee (approved):

lavigne958 (approved):
thank you for this contribution, I checked I found no typing issue. I hope this helps you in the future.

PR #1509, opened 2024-09-03
Added Conditional Formatting Rules logic

mcorrigan:
Added some logic for handling the retrieval and removal of conditional formatting rules. 

I can't run the tests on my work account, but I believe they are correct. Feel free to do what you will with it.

### DIFF
--- a/gspread/spreadsheet.py
+++ b/gspread/spreadsheet.py
@@ -718,6 +718,22 @@ class Spreadsheet:
         self._properties["locale"] = locale
         return res
 
+    def list_conditional_formatting_rules(self, sheetid: int) -> List[Any]:
+        """Lists the spreadsheet's conditional formats"""
+        sheets: List[Mapping[str, Any]] = self.fetch_sheet_metadata(
+            params={"fields": "sheets.properties,sheets.conditionalFormats"}
+        )["sheets"]
+
+        try:
+            sheet = finditem(
+                lambda sheet: sheet["properties"]["sheetId"] == sheetid, sheets
+            )
+
+        except StopIteration:
+            raise WorksheetNotFound("worksheet id {} not found".format(sheetid))
+
+        return sheet.get("conditionalFormats", [])
+
     def list_protected_ranges(self, sheetid: int) -> List[Any]:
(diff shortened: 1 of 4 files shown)


alifeee:
Hi

I ran your tests locally with (I use PowerShell on Windows), and fixed a couple of errors.

```powershell
> $env:GS_RECORD_MODE="all"
> $env:GS_CREDS_FILENAME="tests/creds.json"   
> pytest -k "conditional_format" -v -s
```

Please see the [contributing guide](https://github.com/burnash/gspread/blob/master/.github/CONTRIBUTING.md) for more help on how to run tests

However, the tests still fail.

Please could you explain the motivation for your change? :)

mcorrigan:
> Hi
> 
> I ran your tests locally with (I use PowerShell on Windows), and fixed a couple of errors.
> 
> ```powershell
> > $env:GS_RECORD_MODE="all"
> > $env:GS_CREDS_FILENAME="tests/creds.json"   
> > pytest -k "conditional_format" -v -s
> ```
> 
> Please see the [contributing guide](https://github.com/burnash/gspread/blob/master/.github/CONTRIBUTING.md) for more help on how to run tests
> 
> However, the tests still fail.
> 
> Please could you explain the motivation for your change? :)

Thanks for catching those, it would definitely be easier if I could run the tests myself but I'm not set up to do that at the moment. 

The reason for the changes was just to get and set conditional formats in the Google Sheet via your API instead of directly through the client. 

https://developers.google.com/sheets/api/guides/conditional-format

alifeee:
before I continue, have you looked at gspread formatting?

find https://gspread-formatting.readthedocs.io/en/latest/index.html on https://docs.gspread.org/en/latest/community.html

mcorrigan:
I had no idea that existed, it looks awesome! Probably would work for what I'm trying to do here. I'll look at trying it out with my code. Thanks!

mcorrigan:
Maybe you already can do this, but one thing that would be nice is if there was a simple interface to performing batch operations in gspread like exists in gspread-formatting but with simple function calls like hide_columns() and hide_rows() instead of aggregating request objects.

PR #1510, opened 2024-09-08
better handler API error parsing.

### LINKED ISSUE #1504: Exception requests.exceptions.JSONDecodeError is raised in APIError constructor
**Describe the bug**
My code called  sh.sheet1.batch_get([gsrange], Dimension.cols) which sends request to google sheets. It looks like some error response was returned, but the following exception was raised when APIError was trying to be constructed:
```
  File "helper.py", line 37, in get_capabilities_list
    capabilities = sh.sheet1.batch_get([gsrange], Dimension.cols)
  File ".local/lib/python3.10/site-packages/gspread/worksheet.py", line 1096, in batch_get
    response = self.client.values_batch_get(
  File ".local/lib/python3.10/site-packages/gspread/http_client.py", line 255, in values_batch_get
    r = self.request("get", url, params=params)
  File ".local/lib/python3.10/site-packages/gspread/http_client.py", line 128, in request
    raise APIError(response)
  File ".local/lib/python3.10/site-packages/gspread/exceptions.py", line 45, in __init__
    self.error: Mapping[str, Any] = response.json()["error"]
  File ".local/lib/python3.10/site-packages/requests/models.py", line 978, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

**To Reproduce**
Unfortunately the bug prevents me from knowing what error actually happened. Usually this step succeeds, but once it has failed.
1. call sh.sheet1.batch_get(['G1:S1'], Dimension.cols)

**Expected behavior**
Correct APIError is returned

**Environment info:**
- Ubuntu
- Python 3.10.12
- gspread Version: 6.1.2

lavigne958:
Possibly we encountered an error while parsing the API error JSON, like connection closes early etc and the received JSON is invalid.

Better handle such scenario and build a valid error Dict to keep the exception stack flow running and raise this message to the application.

extract the returned text so if we any bit of valid information at least we can pass that to the application.

This could be helpfull to if we add calls to the API made for web browser that returns HTML error message, it will fail to decode the recieved error and raise a proper error.

closes #1504

### DIFF
--- a/gspread/exceptions.py
+++ b/gspread/exceptions.py
@@ -9,6 +9,7 @@ Exceptions used in gspread.
 from typing import Any, Mapping
 
 from requests import Response
+from requests.exceptions import JSONDecodeError
 
 
 class UnSupportedExportFormat(Exception):
@@ -40,7 +41,19 @@ class APIError(GSpreadException):
     such as when we attempt to retrieve things that don't exist."""
 
     def __init__(self, response: Response):
-        error = dict(response.json()["error"])
+        try:
+            error = response.json()["error"]
+        except JSONDecodeError:
+            # in case we failed to parse the error from the API
+            # build an empty error object to notify the caller
+            # and keep the exception raise flow running
+
+            error = {
+                "code": -1,
+                "message": response.text,
+                "status": "invalid JSON",
+            }
+
         super().__init__(error)
         self.response: Response = response
         self.error: Mapping[str, Any] = error


alifeee (approved):
looks good to me.

is there a way to test this?

lavigne958:
I could possibly manually test it, but that's quiet hard to do. otherwise, using automatic test no :disappointed: 
that would require us to close the connection perfectly in the middle of the transfer of the response from the server, way too hard to be automatic :laughing: 

we do not regress in anything so far.

alifeee:
we could make a test that uses a HTTP client and mock the response?

lavigne958:
> we could make a test that uses a HTTP client and mock the response?

Possibly yes 🤔
We could use a custom http client for a single request that returns an invalid JSON.

I'm worried how this is gonna play with the cassette recorder too 🤔 it worth the try.

dhuang:
appreciate the quick fix here, we started seeing these as well. any chance we'll get a release with this change?

lavigne958:
It's done :heavy_check_mark: 
in #1512  I managed to create a fake HTTP client that can be tested and raises the appropriate exception (the regular APIError exception).

lavigne958:
> appreciate the quick fix here, we started seeing these as well. any chance we'll get a release with this change?

so far we can't tell, we have quite a few new commits but nothing minor/major except this one.
we need to discuss what is waiting to be reviewed or merged and check our priorities.

alifeee:
I have turned the release into an issue... we will do it soon :)

https://github.com/burnash/gspread/issues/1516

gspread/exceptions.py
```diff
@@ -40,7 +41,19 @@ class APIError(GSpreadException):
...
-        error = dict(response.json()["error"])
+        try:
+            error = response.json()["error"]
+        except JSONDecodeError:
+            # in case we failed to parse the error from the API
+            # build an empty error object to notify the caller
+            # and keep the exception raise flow running
+
+            error = {
+                "code": -1,
```
natebessa, line 52: Hey @lavigne958 I'm wondering if the error code here should instead be `response.status_code`. I noticed this because the BackOffHTTPClient will not retry to send the request because -1 is not an expected error code [here](https://github.com/burnash/gspread/blob/master/gspread/http_client.py#L573).

lavigne958: Hi that's a good point. 

Unfortunately we can't simply set the return code as we failed to parse the returned error.

This could be 2 reasons: 1. The response is incomplete/connection lost etc. 2. The response is in html or some other format we don't handle nor expect but the response is valid !

I will check with alifee what is best here, I have 2 ideas:
1. Instead of returning a mock http response, we raise a new Gspread issue and the user is in charge of retrying or not. We don't take a decision here.

2. We add a new setting for the http client to retry on -1 errors. So the user can enable that setting and the http client will retry on -1 error code too.

@alifee what you do think ?

natebessa: For what it's worth, I imagine there would still be HTTP codes coming back with the scenarios you mentioned, e.g. 408 (timeout), 502 (server error, with html response).

Kirkman:
> Helpful change, thanks for fixing. We need this fix because suddenly we've been getting a lot of 502 errors from the Google Sheets API. The response on the 502 does not contain a json object but instead the HTML/CSS of an error page, which triggers a JSONDecodeError. We implemented gspreads `BackOffHTTPClient` to retry on those 502s but the JSONDecodeError breaks that.

Glad to know I'm not the only one seeing this the last few weeks.

natebessa:
@Kirkman absolutely. It's almost 1 in 10 calls for me at this rate. Not something I'd expect from Google, even if it's a free service. I can't find much chatter about it online, and no acknowledgement from Google on any social media. Meanwhile my Google Cloud dashboard is massively undercounting the API error rate I'm seeing.

eitchtee:
The intermittent 502 errors seems to be a bug on Google's end, related issue: https://issuetracker.google.com/issues/369670473

alifeee:
hi all

a new release `6.1.3` has been released ([on github](https://github.com/burnash/gspread/releases/tag/v6.1.3)) ([on PyPi](https://pypi.org/project/gspread/6.1.3/)) which contains a fix for this issue

we now catch the non-JSON error and turn it into a `GSpreadException` which should be easier to catch and find out what went wrong with the request

see more context here: https://github.com/burnash/gspread/pull/1510

az22-atl:
Heya, ever since 6.1.3 was released, I can no longer import gspread using Databricks notebooks in Python 3.8; I get a `cannot import name 'JSONDecodeError' from 'requests.exceptions' (/databricks/python/lib/python3.8/site-packages/requests/exceptions.py) `error. Is this a known issue?

PR #1512, opened 2024-09-22
Add test on receiving an invalid JSON in the APIError exception handler.

lavigne958:
Manually build an invalid JSON body. Using HTML to fake using the wrong endpoint that returns HTML instead of JSON.

Use that new http client to do some request and make sure the error matches.

### DIFF
--- a/tests/cassettes/SpreadsheetTest.test_bad_json_api_error.json
+++ b/tests/cassettes/SpreadsheetTest.test_bad_json_api_error.json
@@ -1,0 +1,232 @@
+{
+    "version": 1,
+    "interactions": [
+        {
+            "request": {
+                "method": "POST",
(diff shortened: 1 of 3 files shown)


alifeee:
looks splendid ! :)

alifeee (approved):

alifeee:
hi all

a new release `6.1.3` has been released ([on github](https://github.com/burnash/gspread/releases/tag/v6.1.3)) ([on PyPi](https://pypi.org/project/gspread/6.1.3/)) which contains a fix for this issue

we now catch the non-JSON error and turn it into a `GSpreadException` which should be easier to catch and find out what went wrong with the request

see more context here: https://github.com/burnash/gspread/pull/1510

PR #1525, opened 2024-10-25
rearrange flow to remove reference to `creds`

alifeee:
I don't fully know how to test this

@lavigne958 please advise :)

### DIFF
--- a/gspread/auth.py
+++ b/gspread/auth.py
@@ -284,9 +284,10 @@ def oauth_from_dict(
 
     if authorized_user_info is not None:
         creds = OAuthCredentials.from_authorized_user_info(authorized_user_info, scopes)
-
-    if not creds and credentials is not None:
+    elif credentials is not None:
         creds = flow(client_config=credentials, scopes=scopes)
+    else:
+        raise ValueError("no credentials object supplied")
 
     client = Client(auth=creds, http_client=http_client)
 


lavigne958:
The fox looks good ! I just want to run a test to confirm.
To test this code you need to generate oauth credentials, then extract the JSON information and pass them as `dict` and see what happens 🧐

alifeee:
I try to test using 

```python
import gspread

gc = gspread.oauth(
    credentials_filename="credentials.json",
    authorized_user_filename="authorised_user.json",
)

sht = gc.open_by_url(
    "https://docs.google.com/spreadsheets/d/89fua83fu983w8fu89s38fu/"
)
ws = sht.sheet1
print(ws.get())
```

but I get `gspread.exceptions.APIError: APIError: [400]: This operation is not supported for this document`. I have not used oauth before so it confuses me. Perhaps I will attempt this again soon, but if I do not, feel free to

lavigne958:
> I try to test using
> 
> ```python
> import gspread
> 
> gc = gspread.oauth(
>     credentials_filename="credentials.json",
>     authorized_user_filename="authorised_user.json",
> )
> 
> sht = gc.open_by_url(
>     "https://docs.google.com/spreadsheets/d/89fua83fu983w8fu89s38fu/"
> )
> ws = sht.sheet1
> print(ws.get())
> ```
> 
> but I get `gspread.exceptions.APIError: APIError: [400]: This operation is not supported for this document`. I have not used oauth before so it confuses me. Perhaps I will attempt this again soon, but if I do not, feel free to

the way oauth works is as follow:
- the first time you run the call to `gspread.oauth()` it uses the credentials file for authentication
- then it creates a file called `authorized_user.json` where it saves some token and some infos.
- you obtain some credentials to use for any request you make

later (like some hours, days laters)
- you run the oauth call again
- it see you `authorized_user.json` file
- it tries to use it to authenticate yourself as the token is possibly still valid !
- if it works you get credentials from that file and you keep making requests
- if it does not you need to re-authenticate using the original credential file

Here it seems you make a request to a wrong url, the spreadsheet ID look too short and does not look like it's valid, usually it contains letters+numbers and it's case sensitive so it has a mix of lowercase + uppercase letters + numbers.

lavigne958 (approved):
I ran the code and tried using received authentication token, running with new credentials etc. everything works as expected.

alifeee:
I'd love to make a test for this, but for now I will accept…

> I ran the code and tried using received authentication token, running with new credentials etc. everything works as expected.

…as enough of a test, as I don't have time/effort to create one (someone could feel free to!)

so I will merge, and release "soon". for now, you can pull the code from the `master` branch, if you'd like.

PR #1536, opened 2024-12-27
explicitly list top-level exports

afeld:
Fixes Pylance/Pyright reporting `"oauth" is not exported from module "gspread"` when using `gspread.oauth()` (for example). Screenshot from VSCode:

image

[More about the warning.](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportPrivateImportUsage)

Note that I have not tested the change.


alifeee:
hey, thanks for the suggestion !!

I believe this is a duplicate of https://github.com/burnash/gspread/pull/1531 and as such can be closed without merging 

please let me know if this is not the case :]

afeld:
Ah, missed that, thanks!

PR #1543, opened 2025-02-16
Google Sheet Format

Newbern:
I had to use gspread for a project and wanted to see this file would be beneficial for you


alifeee:
hi ! thanks for the PR !

unfortunately we do not want to deal with formatting in gspread, but check out [`gspread-formatting`](https://pypi.org/project/gspread-formatting/) !

PR #1548, opened 2025-04-22
Include the duplicate column names in error message

NickCrews:
When I get this error, I would like to go find the problematic column in the sheet and then fix it. This was difficult before, now it is easy.

Also, fixup the lint action in CI, because I was running into errors:

- We don't need to install mypy (the tox env does that for us)
- We only run this on python versions 3.8+, so the version check isn't needed
- Upgrade the usage of `safety` from the 2.x CLI version to the 3.x version (https://docs.safetycli.com/safety-docs/safety-cli/migrating-from-safety-cli-2.x-to-safety-cli-3.x). Before, we were getting `DEPRECATED: this command (`check`) has been DEPRECATED, and will be unsupported beyond 01 June 2024.` and this caused an error in https://github.com/burnash/gspread/actions/runs/14607328648/job/40978873165?pr=1548
- Pin the version of safety so that this regression doesn't happen again

I can split the CI stuff into a separate PR, but I prefer not to. I can't test the updated CI workflow because I don't have the creds for the `safety` account (and the workflow run, running against my branch, doesn't have access to the creds either). So, I think the only thing to do would be for

1. you @lavigne958 to pull this PR locally and try it using your local creds for `safety`
2. just try merging this to `master` and seeing if it works, then reverting if it doesn't (the YOLO approach ;) )

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -24,11 +24,11 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - run: pip install -U pip
-      - run: pip install -U bandit pyupgrade safety==3.3.1 tox setuptools
+      - run: pip install -U bandit pyupgrade pip-audit tox setuptools
       - run: bandit --recursive --skip B105,B110,B311,B605,B607 --exclude ./.tox .
       - run: tox -e lint
       - run: tox -e py
       - run: shopt -s globstar && pyupgrade --py3-only **/*.py # --py36-plus
-      - run: safety scan -i 42559 -i 62044 -i 67599 -i 70612 # pip version issue, we don't care about it, we don't ship it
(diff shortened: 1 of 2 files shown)


alifeee:
hi @NickCrews, thanks for the PR! wonderfully formed ;]

seems like a useful error message to me. thanks for contributing it!

sorry about the CI ... I hold to allegiance to `safety`, especially as they have (newly) required an account, so I suggested removing it (https://github.com/burnash/gspread/pull/1551) - I pulled that change over to this PR, too

I will merge this hopefully in a few days… after #1551

NickCrews:
Oops, just noticed a bug in get_dupes(). I'll push a fix in a sec. Thanks for the quick review.

alifeee:
thanks for the change @NickCrews :]

PR #869, opened 2021-05-16
GitHub Action to lint Python code

cclauss:
Because Travis CI is no longer running on PRs.
Output: https://github.com/cclauss/gspread/actions

PR #1551, opened 2025-05-06
switch safety for pip-audit

alifeee:
safetycli was added in https://github.com/burnash/gspread/pull/869

it recently switched to requiring an account -> https://github.com/pyupio/safety/issues/663

image

in order to keep maintenance burden low, I switched it to pip-audit, which seems to do the same thing

I do not think the need for such a step in the CI is not really strong, but I have endeavoured not to "decrease" the security of the project.

this will allow https://github.com/burnash/gspread/pull/1548 to be more easily merged

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -24,12 +24,12 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - run: pip install -U pip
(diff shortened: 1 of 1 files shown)


alifeee:
I will merge on @lavigne958's review

alifeee:
merging without review

PR #1568, opened 2025-10-12
test: change service account to oauth client

xoiga123:
I got hit with
> `The user's Drive storage quota has been exceeded`

when trying to test with a service account created from my `gmail.com` account. Apparently service accounts have 0 Drive storage space available, and I couldn't do any test with `client.create(name)`, since that requires creating a sheet in the service account's Drive.

I tried some workarounds to no avail:
1. Share a folder with the service account with `Editor` permission and do `client.create(name, folder_id=...)`, still fails.
2. Enable domain-wide-delegation so that the service account can impersonate my `gmail.com` account, fails because I cannot login to `admin.google.com` with a `gmail.com` account.

Unless I'm seriously misunderstanding something, the only way for a regular `gmail.com` user to add new tests now is by authenticating with OAuth instead so that test sheets can be created.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -92,9 +92,9 @@ gspread uses [vcrpy](https://github.com/kevin1024/vcrpy) to record and replay HT
 
 ### `GS_CREDS_FILENAME` environment variable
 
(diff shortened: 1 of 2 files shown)


Ev2geny:
Hi, I think the problem is already solved now with this accepted PR https://github.com/burnash/gspread/pull/1593

antoineeripret:
Closing as per @Ev2geny comment. Thanks !

PR #1590, opened 2026-06-19
feat: add custom JSON serializer support to HTTPClient

### LINKED ISSUE #1556: Using a custom json serialiser in worksheet.update
I'm using GSpread to write values from a (polars) dataframe into Google Sheet via `worksheet.batch_update` and would like to be able to specify the JSON serialiser when doing that. This would help when dealing with dates serialisation issues for instance.

In the context of `batch_update`, the trail is:

* [here](https://github.com/burnash/gspread/blob/6074285628c88f3a73d024cefd2f6691ad4391d6/gspread/worksheet.py#L1368) the query is made via `self.client.values_batch_update(...)` 
* [here](https://github.com/burnash/gspread/blob/6074285628c88f3a73d024cefd2f6691ad4391d6/gspread/http_client.py#L139) is what it correspond to in the `http_client` code 

The `data=data` [here](https://github.com/burnash/gspread/blob/6074285628c88f3a73d024cefd2f6691ad4391d6/gspread/http_client.py#L119) in the `request` method gives me hope that this

https://github.com/burnash/gspread/blob/6074285628c88f3a73d024cefd2f6691ad4391d6/gspread/http_client.py#L139

could maybe be replaced by something like

```python
if json_serialiser is not None:
    r = self.request("post", SPREADSHEET_BATCH_UPDATE_URL % id, data=json_serialiser(body))
else:
    r = self.request("post", SPREADSHEET_BATCH_UPDATE_URL % id, json=body)
```

Gardner-Programs:
## Summary

Closes #1556.

Adds the ability to configure a custom JSON serializer for request bodies, so users can write values that the standard library `json` module cannot encode by default (e.g. `datetime`, `Decimal`) without pre-stringifying their data.

```python
import json
import datetime

gc.set_serializer(lambda body: json.dumps(body, default=lambda o: o.isoformat()))

ws.update([[datetime.date(2026, 6, 19)]], "A1")  # now works
```

Pass `None` to restore the default behavior.

## Design

The issue suggested adding a `serializer` argument at the `Worksheet.update` / `batch_update` level. I went with configuring it once on the client instead, for a few reasons:

- It's a single choke point in `HTTPClient.request()` rather than threading a parameter through several method signatures, so it's smaller and less likely to break existing write paths.
- It applies uniformly to every write (`update`, `batch_update`, etc.).
- It mirrors the existing `set_timeout` pattern (setter on both `HTTPClient` and `Client`), so it fits the codebase's conventions.

If you'd prefer a per-call argument, that can be added on top of this foundation later. Happy to adjust the API in review.

## Implementation

- `SerializerType` alias + optional `self.serializer` (defaults to `None`).
- `HTTPClient.set_serializer()` and a `Client.set_serializer()` forwarder.
- In `HTTPClient.request()`, when a serializer is set and there is a JSON body, the body is serialized and sent as `data=` with a `Content-Type: application/json` header instead of via `json=`. Without a serializer, behavior is unchanged.

## Backward compatibility

Fully backward compatible. `serializer` defaults to `None`, and the existing `json=` code path is untouched when no serializer is set.

## Tests

New `tests/http_client_test.py` (mocked session, no cassette needed since this is request-layer logic):

- `test_default_uses_json_kwarg`: no serializer, so the body still goes out via `json=` (proves unchanged behavior).
- `test_custom_serializer_uses_data_and_header`: serializer set, so the body is serialized into `data=`, `json` is nulled, and the JSON content-type header is added.
- `test_custom_serializer_handles_non_native_types`: a `datetime.date` (which stdlib `json.dumps` rejects) is written successfully.

Also manually verified end-to-end against the live Sheets API: without a serializer a `date` write raises `TypeError`, and with one the value lands correctly in the cell.

## Docs

Added a short "Using a Custom JSON Serializer" section to the user guide.

## Notes

- I left off a `.. versionadded::` directive since I don't know the target release. Happy to add `6.3.0` (or whatever you prefer) on request.

### DIFF
--- a/gspread/http_client.py
+++ b/gspread/http_client.py
@@ -12,6 +12,7 @@ from http import HTTPStatus
 from typing import (
     IO,
     Any,
+    Callable,
     Dict,
     List,
     Mapping,
@@ -44,6 +45,7 @@ from .urls import (
 from .utils import ExportFormat, convert_credentials, quote
 
 ParamsType = MutableMapping[str, Optional[Union[str, int, bool, float, List[str]]]]
+SerializerType = Optional[Callable[[Mapping[str, Any]], Union[str, bytes]]]
 
 FileType = Optional[
     Union[
@@ -83,6 +85,7 @@ class HTTPClient:
             self.session = AuthorizedSession(self.auth)
 
         self.timeout: Optional[Union[float, Tuple[float, float]]] = None
+        self.serializer: SerializerType = None
 
     def login(self) -> None:
         from google.auth.transport.requests import Request
@@ -102,16 +105,45 @@ class HTTPClient:
         """
         self.timeout = timeout
 
+    def set_serializer(self, serializer: SerializerType) -> None:
+        """Set a custom JSON serializer used to encode request bodies.
+
+        The serializer is a callable that takes the request body (a mapping)
+        and returns its JSON-encoded form as ``str`` or ``bytes``. It is
+        applied to every request that sends a JSON body, which is useful for
+        handling types the standard library ``json`` module cannot serialize
+        by default (e.g. ``datetime`` or ``Decimal``).
+
+        Use value ``None`` to restore the default serialization behavior.
+
+        :param serializer: A callable with the same signature as
+            ``json.dumps`` (e.g. ``functools.partial(json.dumps, default=...)``),
+            or ``None`` to use the default.
+
+        Example::
+
+            import json
+
+            client.set_serializer(
+                lambda body: json.dumps(body, default=str)
+            )
+        """
+        self.serializer = serializer
+
     def request(
         self,
         method: str,
         endpoint: str,
         params: Optional[ParamsType] = None,
-        data: Optional[bytes] = None,
+        data: Optional[Union[str, bytes]] = None,
         json: Optional[Mapping[str, Any]] = None,
         files: FileType = None,
         headers: Optional[MutableMapping[str, str]] = None,
     ) -> Response:
+        if self.serializer is not None and json is not None:
+            data = self.serializer(dict(json))
+            json = None
+            headers = {**(headers or {}), "Content-Type": "application/json"}
         response = self.session.request(
             method=method,
             url=endpoint,
--- a/gspread/http_client.py
+++ b/gspread/http_client.py
@@ -9,6 +9,7 @@ Google API.
 
 import time
 from http import HTTPStatus
+from json import dumps as json_dumps
 from typing import (
     IO,
     Any,
@@ -45,7 +46,7 @@ from .urls import (
 from .utils import ExportFormat, convert_credentials, quote
 
 ParamsType = MutableMapping[str, Optional[Union[str, int, bool, float, List[str]]]]
-SerializerType = Optional[Callable[[Mapping[str, Any]], Union[str, bytes]]]
+DefaultSerializerType = Optional[Callable[[Any], Any]]
 
 FileType = Optional[
     Union[
@@ -85,7 +86,6 @@ class HTTPClient:
             self.session = AuthorizedSession(self.auth)
 
         self.timeout: Optional[Union[float, Tuple[float, float]]] = None
-        self.serializer: SerializerType = None
 
     def login(self) -> None:
         from google.auth.transport.requests import Request
@@ -105,31 +105,6 @@ class HTTPClient:
         """
         self.timeout = timeout
 
-    def set_serializer(self, serializer: SerializerType) -> None:
-        """Set a custom JSON serializer used to encode request bodies.
-
-        The serializer is a callable that takes the request body (a mapping)
-        and returns its JSON-encoded form as ``str`` or ``bytes``. It is
-        applied to every request that sends a JSON body, which is useful for
-        handling types the standard library ``json`` module cannot serialize
-        by default (e.g. ``datetime`` or ``Decimal``).
-
-        Use value ``None`` to restore the default serialization behavior.
-
-        :param serializer: A callable with the same signature as
-            ``json.dumps`` (e.g. ``functools.partial(json.dumps, default=...)``),
-            or ``None`` to use the default.
-
-        Example::
-
-            import json
-
-            client.set_serializer(
-                lambda body: json.dumps(body, default=str)
-            )
-        """
-        self.serializer = serializer
-
     def request(
         self,
         method: str,
@@ -139,9 +114,10 @@ class HTTPClient:
         json: Optional[Mapping[str, Any]] = None,
         files: FileType = None,
         headers: Optional[MutableMapping[str, str]] = None,
+        default_serializer: DefaultSerializerType = None,
     ) -> Response:
-        if self.serializer is not None and json is not None:
-            data = self.serializer(dict(json))
+        if default_serializer is not None and json is not None:
+            data = json_dumps(dict(json), default=default_serializer)
             json = None
             headers = {**(headers or {}), "Content-Type": "application/json"}
         response = self.session.request(
@@ -179,6 +155,7 @@ class HTTPClient:
         range: str,
         params: Optional[ParamsType] = None,
         body: Optional[Mapping[str, Any]] = None,
+        default_serializer: DefaultSerializerType = None,
     ) -> Any:
         """Lower-level method that directly calls `PUT spreadsheets/<ID>/values/<range> <https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/update>`_.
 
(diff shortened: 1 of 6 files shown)


gspread/http_client.py
```diff
@@ -102,16 +105,45 @@ def set_timeout(self, timeout: Optional[Union[float, Tuple[float, float]]]) -> N
...
         endpoint: str,
         params: Optional[ParamsType] = None,
-        data: Optional[bytes] = None,
+        data: Optional[Union[str, bytes]] = None,
         json: Optional[Mapping[str, Any]] = None,
         files: FileType = None,
         headers: Optional[MutableMapping[str, str]] = None,
     ) -> Response:
+        if self.serializer is not None and json is not None:
+            data = self.serializer(dict(json))
```
tlienart: why is the dict needed here?

tlienart (approved):

tlienart:
cool, that would definitely solve the problem I had and I think your point about putting this at the higher level makes total sense 👍

gspread/http_client.py
Gardner-Programs: I was just mirroring the existing `json=dict(json) if json else None` line below it. Since `json` is typed as `Mapping[str, Any]` it isn't guaranteed to be a concrete `dict`, so this makes sure the serializer gets one it can encode.

antoineeripret:
Hey @Gardner-Programs & @tlienart, 

Can't we implement something simpler to fix the issue? I'm not a fan of ... 

`gc.set_serializer(lambda body: json.dumps(body, default=lambda o: o.isoformat()))`

... because it's a global state. 

Wouldn't it make more sense to let users serialize their data before running the update call? We'd have to implement the following: 

1. Add a serializer parameter. When provided, convert json → data before passing to requests (`http_client.py` // `request()`) 
2. Accept and forward serializer (`http_client.py` // `values_batch_update()`)
3. Add the parameter and pass it through (`worksheet.py` // `Worksheet.update()` + `Worksheet.batch_update()`)

User would be able to run 

```python 
ws.update(
    [[datetime.date(2026, 6, 23)]],
    "A1",
    serializer=lambda b: json.dumps(b, default=lambda o: o.isoformat()),
)
```

Overall, we'd achieve the same but we'd remove layers of complexity. 

What do you think?

Gardner-Programs:
Hey @antoineeripret, thanks for taking a look 🙏

I actually modeled `set_serializer()` on the existing `set_timeout()`. It's stored per-client and applied down in `request()`, so it follows a pattern the library already uses rather than introducing a new one. It's also a no-op until you opt in, since `serializer` defaults to `None` and `request()` only does anything custom when it's actually set. So nothing changes for anyone who doesn't touch it.

On why I went this way: in my own usage my serialization needs are per-spreadsheet, not per-call. A given script talks to one spreadsheet and every write needs the same handling (e.g. `datetime` to ISO), so I set it once at the top and I'm done. Having to re-pass `serializer=...` on every `update`/`batch_update` would be pretty painful from where I sit. That said, I'm aware my usage shouldn't dictate everyone's, so I do get the appeal of explicit per-call control.

The other thing that pushed me toward the setter is coverage. It hooks in at `request()`, which is the single point all the JSON-body calls flow through, so one setting covers `update`, `batch_update`, `format`, `append_rows`, permissions, and the rest.

That said, here's a possible compromise. What if we support both instead of picking one? Keep `set_serializer()` as a client-wide default, and also add an optional `serializer=` to the write methods that overrides it per call. Precedence would be per-call arg, then client default, then the built-in. So you'd get exactly the explicit, local API you're after:

```python
ws.update(
    [[datetime.date(2026, 6, 23)]],
    "A1",
    serializer=lambda b: json.dumps(b, default=lambda o: o.isoformat()),
)
```

while folks with uniform per-spreadsheet needs can still set it once. It stays fairly small because `request()` already centralizes the logic, so it just takes the override and falls back to the client default.

I do want to be honest about the trade-offs though:

- It's a bit more complexity. Two ways to do the same thing means more API surface to document, plus a precedence rule people have to understand.
- Coverage would be partial. Realistically I'd thread `serializer=` through the data-writing methods (`update`/`batch_update`), so per-call control would exist for those but not for every method that sends a body. We could thread it through all of them for full parity, it's just more plumbing and more signatures to maintain. Either way the client-wide setter stays as the catch-all that covers everything.

Happy to go whichever direction you think is best, whether that's the setter as-is, the per-call param on its own, or the combined approach. Let me know what you'd prefer and I'll adjust.

antoineeripret:
Hi @Gardner-Programs, 

That's a very good point you raise, I actually have other scripts with other libraries where I have to do just that, and it's a pain 😢. That being said, I wouldn't go with the double-implementation because **I'm unsure it's a good idea for a library that is currently short of maintainers**. 

Let me suggest a different approach as a compromise, w/o adding too much complexity. 

- We'd implement the per-call solution 
- To achieve a per-spreadsheet behavior, you could use [functool](https://docs.python.org/3/library/functools.html) as follows: 

```python 
import json, functools
import gspread

def _iso_serializer(body):
    return json.dumps(body, default=lambda o: o.isoformat())

#we define the per_spreadsheet here 
def open_sheet(name):
    gc = gspread.oauth()
    ws = gc.open(name).sheet1
    ws.update = functools.partial(ws.update, serializer=_iso_serializer)
    ws.batch_update = functools.partial(ws.batch_update, serializer=_iso_serializer)
    return ws

ws = open_sheet("Report A")
ws.update(...)  # serializer already baked in
```

What do you think? Please do not hesitate to let me know if you think that your approach is better: I don't want to push mine, I just want to make sure we implement the right solution given our objectives and the library's context. 

Thank you !

Gardner-Programs:
Hey @antoineeripret, the setter is what worked for my personal usage, but I agree it's not the best call for the repo. Let's go per-call and drop it.

The global default fit the majority of the cases I've personally hit, so it felt like the natural choice at the time. But I've never worked with a team that shared a single client across multiple scripts or functions, and once you do, I can see how a mutable setter would cause real problems. That's a context I just wasn't designing for. The `functools.partial` recipe gets me the same set-once feel on my side without any of that risk, so that works great as the alternative.

One small thing I wanted to run by you before I rework it. Instead of a full serializer that the user owns:

```python
ws.update([[date(2026, 6, 23)]], "A1",
          serializer=lambda b: json.dumps(b, default=lambda o: o.isoformat()))
```

what if we expose just the `default=` hook from `json.dumps` and let gspread keep ownership of the encoding?

```python
ws.update([[date(2026, 6, 23)]], "A1", json_default=str)  # str covers datetime, Decimal, UUID...
```

Same result for the datetime/Decimal case, but since the library still owns `json.dumps`, the user can't accidentally hand back invalid JSON or break the `Content-Type`. Smaller surface, fewer ways to get it wrong, which felt in line with your point about keeping things light for the maintainers.

I'm happy either way, so just let me know which you'd prefer. One thing I want to flag with the per-call approach: unlike the setter, it only applies to the methods we thread it through, so any future data-writing method would need the same param to support it. Nothing blocking, just worth deciding up front. Do you want me to thread it through `update`/`batch_update` only, or `append_row(s)` too, and should we treat "new write methods accept `json_default`" as a loose convention going forward?

antoineeripret:
Hey @Gardner-Programs, 

Great idea, really love it ! 

I'd maybe use a different name for the parameter though: the name makes sense if you already know the `json` library, less so in `gspread`'s context. `serializer` or `value_serializer_default` could be more self-describing IMO. 

Regarding your questions: 

1/ **Which methods to thread it through**: I'd add `append_row(s)` since they write cell data just like `update`. Non-data methods don't need it though. 
2/ **Convention going forward:**: yes, establishing it as a soft convention ("write methods accept json_default") is the right call IMO. 

Thank you !

Gardner-Programs:
Perfect, let's lock it in.

On the name, I went with `default_serializer`. `value_serializer_default` was my other thought but it felt a touch long, and I wanted to steer clear of plain `serializer` since that's what we'd been calling the full-body version earlier in the thread, didn't want to overload it. Happy to switch to the more explicit name if you'd rather, though.

For scope I'll thread it through `update`, `batch_update`, and `append_row(s)`, and leave the non-data methods alone. I'll keep "write methods accept `default_serializer`" as the soft convention going forward.

I'll get the rework up shortly. Thanks for the back and forth on this!

antoineeripret:
Hey @Gardner-Programs, sorry I didn't catch your last commit until now. Do you have missing changes to apply? I'll take some time tomorrow to review it and hopefully merge it :)

Gardner-Programs:
Hey @antoineeripret, no worries. I forgot to send a follow up when I pushed the commit, it's up to date so its ready for review whenever you have the time.

antoineeripret:
Hey @Gardner-Programs, 

I've checked the PR this morning and LGTM 🥳 . Thank you very much, I'll merge it now !

Issue #1592, opened 2026-06-22
Online tests can no longer be run with a service account (Drive storageQuotaExceeded)


Ev2geny:
## Summary

It is no longer possible to run online tests and to record test cassettes using a **service account**, which is the method described in [`CONTRIBUTING.md`](https://github.com/burnash/gspread/blob/master/.github/CONTRIBUTING.md). The test suite creates spreadsheets (`client.create(...)`), and Google now rejects file creation by service accounts.

## Steps to reproduce

1. Follow the contributing guide to record cassettes online with a service account:
   ```bash
   GS_CREDS_FILENAME=<service_account.json> GS_RECORD_MODE=all tox -e py
   ```
2. Any test that creates a spreadsheet fails during setup (e.g. `tests/cell_test.py` → `client.create(name)`).

## Actual result

The first `POST https://www.googleapis.com/drive/v3/files` returns:

```
APIError: [403]: The user's Drive storage quota has been exceeded.
```

Because this 403 carries `"domain": "usageLimits"`, `BackOffHTTPClient` treats it as a rate-limit and retries with exponential backoff, so the run appears to hang for a long time before failing.

## Root cause

  - Service accounts historically seem to had the standard **15 GB** Drive quota like any Google account, so creating files worked. See e.g. ["Handle Google Drive's service account limit of 15 GB"](https://harshalrj25.medium.com/handle-google-drives-service-account-limit-of-15gb-6d67d468c7f9) — the article exists precisely because service accounts *had* storage to fill. This contrasts with the current Google docs above, which now state service accounts have no quota and cannot own files.

- Google policy is different now: **service accounts no longer have Drive storage quota and cannot own files.** Per the official Google Drive API documentation:
  
> Service accounts don't have storage quota and can't own any files. Instead, they must upload files and folders into shared drives, or use OAuth 2.0 to upload items on behalf of a human user.

— <https://developers.google.com/workspace/drive/api/guides/handle-errors> (see the `storageQuotaExceeded` section)

Since the tests rely on the service account *creating and owning* spreadsheets, this method no longer works.

## Evidence it used to work

- The recorded cassettes in `tests/cassettes/` (committed since 2018) contain successful live `POST /drive/v3/files` → `200 OK` responses creating spreadsheets — i.e. the recording account could create/own files, following the documented service-account workflow.

## Proposed fix

[...]

PR #1593, opened 2026-06-22
Fix issue 1592 (Online tests can no longer be run with a service account)

Ev2geny:
Test code is updated to allow the test suite to authenticate using  human user credintials

CONTRIBUTING.md is updated with  instructions on how to obtain these human user credintials (an authorized_user.json file)

This fixes the issue https://github.com/burnash/gspread/issues/1592


Ev2geny:
@antoineeripret , I have implemented the changes, as we discussed in [the issue](https://github.com/burnash/gspread/issues/1592#issuecomment-4777554882). I also slightly updated the relevant section of the `docs/oauth2.rst` to better reflect the current procedure and to allow to refer to it from the `.github/CONTRIBUTING.md`

Issue #1594, opened 2026-07-14
test_add_timeout leaks a 1-second timeout into the shared client, which makes online test runs fail randomly:


Ev2geny:
## Summary

When the test suite is run online (against the live API), a varying handful of tests in
`tests/client_test.py` fail with `requests.exceptions.ReadTimeout ... (read timeout=1)`.
The failures look random — a different subset fails on each run — but they are not.

`ClientTest.test_add_timeout` calls `self.gc.set_timeout(1)` and never restores the default.
The `client` fixture in `tests/conftest.py` is **module-scoped**, so `self.gc` is a single
`Client` object shared by every test in the module. The 1-second read timeout therefore leaks
into every test that runs after `test_add_timeout`, and those tests fail whenever a real
round-trip to the Google API takes longer than one second — which, online, it often does.

## Reproduction

```bash
GS_CREDS_FILENAME=<creds.json> tox -e py -- --disable-vcr -k ClientTest
```

Expect a random-looking subset of the tests after `test_add_timeout` to fail with
`read timeout=1`. Re-running produces a different subset, drawn from the same ten tests.

## Evidence

pytest runs `unittest.TestCase` methods in alphabetical order, and `test_add_timeout` is third
of the thirteen tests in `ClientTest`. Across three online runs, **every** failure came from the
ten tests that sort after it; the two that sort before it never failed, and `test_add_timeout`
itself never failed:

| # | Test (alphabetical order) | Failed in any run? |
|---|---------------------------|--------------------|
| 1 | `test_access_non_existing_spreadsheet` | no |
| 2 | `test_access_private_spreadsheet` | no |
| 3 | **`test_add_timeout`** ← sets timeout=1 | no |
| 4–13 | `test_client_export_spreadsheet` … `test_openall` | **yes — all failures are here** |

One of the three runs is a clean demonstration: *all ten* tests after `test_add_timeout` failed
or errored, while the two before it passed.

```
FAILED tests/client_test.py::ClientTest::test_import_csv - requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='www.googleapis.com', port=443): Read timed out. (read timeout=1)
FAILED tests/client_test.py::ClientTest::test_open_all_has_metadata - requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='sheets.googleapis.com', port=443): Read timed out. (read timeout=1)
FAILED tests/client_test.py::ClientTest::test_openall - requests.exceptions.ReadTimeout: ...
[...]

PR #1595, opened 2026-07-14
Fix issue 1594 (test_add_timeout leaks a 1-second timeout into the shared client)

Ev2geny:
Restore the default timeout when the test test_add_timeout finishes, so it cannot leak into the shared client.

This fixes this issue https://github.com/burnash/gspread/issues/1594

After this change all online tests passed for me.


janmarsino98 (approved):
Reviewed the module-scoped client fixture and ran the modified test successfully. addCleanup is registered immediately after set_timeout, so the default is restored even when the test fails later. No blocking issues found.

PR #1596, opened 2026-07-15
Reset backoff state after non-retryable API errors

lovewave02:
## What changed

A non-retryable API error currently increments `BackOffHTTPClient`'s retry
counter before the error is returned. If the same client later hits a retryable
error, its first delay starts at 4 seconds instead of 2 seconds.

This resets the counter before propagating a terminal API error. The regression
test exercises a 403 -> 500 -> success sequence and verifies that the 500 starts
with the normal 2-second delay.

## Verification

- `tox -e lint,py,build,doc`
- 161 tests passed
- package and Sphinx warning-as-error builds passed

### DIFF
--- a/gspread/http_client.py
+++ b/gspread/http_client.py
@@ -618,6 +618,7 @@ class BackOffHTTPClient(HTTPClient):
                 return response
 
             # failed too many times, raise APIEerror
+            self._NR_BACKOFF = 0
             raise err
         except RefreshError as err:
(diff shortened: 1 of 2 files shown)


antoineeripret:
Hey @lovewave02, 

Thank you for the PR, love it :) 

Can you apply the same change to `RefreshError ` that suffers from the same issue though? 

Thank you !

Issue #1589, opened 2026-06-13
Add method to batch delete multiple row ranges in one API call


Ev2geny:
**Is your feature request related to a problem? Please describe.**
Deleting multiple non-contiguous row ranges requires either multiple HTTP requests (one per contiguous block via `delete_rows()`) or dropping down to `Spreadsheet.batch_update()` directly. When using `batch_update()` for deletion, the worksheet's locally cached `row_count` is not updated, unlike `delete_rows()` which keeps it in sync. This creates problems aleter.  E.g.  any subsequent call to `add_rows()` uses a stale count and resizes the sheet incorrectly — adding far more rows than needed.

**Describe the solution you'd like**
A new `Worksheet.delete_rows_blocks(blocks: list[tuple[int, int]])` method that deletes multiple row ranges in a single API call and correctly updates the cached `row_count` afterwards. Blocks would use 1-based inclusive row numbers, consistent with the existing `delete_rows(start, end)` signature. 
The method shall check, that blocks are not overlapping.

Example:
```python
# Delete rows 3–5 and rows 8–9 in one API call
worksheet.delete_rows_blocks([(8, 9), (3, 5)]) 
```

**Describe alternatives you've considered**
Loop over `delete_rows()` — keeps row_count in sync but makes one API call per block instead of one total.
Call `Spreadsheet.batch_update()` directly — one API call, but leaves row_count stale. Workaround requires patching `worksheet._properties['gridProperties']['rowCount']` manually, which couples user code to private gspread internals.

**Additional context**
n/a

antoineeripret:
Hey @Ev2geny,

Love the feature request.

Would you like to work on a PR to implement this?

Ev2geny:
@antoineeripret , thanks for the feedback. let me try to create the PR

antoineeripret:
Fixed with https://github.com/burnash/gspread/pull/1598

PR #1598, opened 2026-07-16
Implement worksheet methods to delete non-contiguous blocks of rows/columns in one API call (implements issue 1589)

Ev2geny:
This implements https://github.com/burnash/gspread/issues/1589

## What this PR does

Adds new `Worksheet` methods that delete multiple non-contiguous blocks of rows or
columns with a **single** `batchUpdate` API call, while keeping the locally cached
`row_count` / `col_count` in sync (the stale-cache problem described in the issue):

```python
# Delete rows 3–5 and rows 8–9 in one API call
worksheet.delete_rows_blocks([(3, 5), (8, 9)])
```

## Note: slightly expanded scope vs. what was agreed in the issue

The issue discussion agreed on `delete_rows_blocks` only. While implementing it I
followed the existing structure of the module, where `delete_rows` and
`delete_columns` are thin wrappers around a generic `delete_dimension`. Mirroring
that pattern gave the column variant essentially for free, so this PR adds:

- `delete_dimension_blocks(dimension, blocks)` — generic implementation
- `delete_rows_blocks(blocks)` — what was agreed in the issue
- `delete_columns_blocks(blocks)` — the symmetric column variant

## Design

- Blocks are 1-based inclusive `(start, end)` pairs, consistent with
  `delete_rows(start, end)`. They may be tuples or lists
  (`Sequence[Sequence[int]]`), in any order; they are sorted internally and the
  `deleteDimension` requests are emitted bottom-up, so earlier deletions in the
  batch don't shift the indexes of later ones.
- Strict validation raises `ValueError` **before any API call**: empty blocks
  list, a block without exactly 2 elements, `start < 1`, `end < start`, a block
  exceeding the current sheet size, and overlapping blocks (adjacent blocks are
  allowed).
- After the call, the cached `rowCount`/`columnCount` is decremented by the total
  number of deleted rows/columns, so subsequent `add_rows()`/`resize()` calls work
  with correct values.

## Note on `.. versionadded:: 6.3.0`

The docstrings mark the new methods as `versionadded:: 6.3.0`. That version is a
**guess** (current release is 6.2.1 and this is a new feature, so the next minor
seemed likely). If the actual release number ends up different, these three
directives will need updating.

## Testing

- Three new tests: `test_delete_rows_blocks`, `test_delete_columns_blocks`
  (both pass blocks unsorted and mixing lists/tuples) and
  `test_delete_dimension_blocks_validation` (all `ValueError` cases).
- Cassettes recorded; the full suite passes offline (`tox -e py`).
- The new tests were additionally verified against the live API with
  `--disable-vcr`.
- new API is correctly documented in HTML files after running `tox -e doc`

### DIFF
--- a/gspread/worksheet.py
+++ b/gspread/worksheet.py
@@ -2242,6 +2242,158 @@ class Worksheet:
         """
         return self.delete_dimension(Dimension.cols, start_index, end_index)
 
+    def delete_dimension_blocks(
+        self, dimension: Dimension, blocks: Sequence[Sequence[int]]
+    ) -> JSONResponse:
+        """Deletes multiple non-contiguous blocks of rows or columns from
+        the worksheet with a single API call.
+
+        :param dimension: A dimension to delete. ``Dimension.rows`` or ``Dimension.cols``.
+        :type dimension: :class:`~gspread.utils.Dimension`
+        :param blocks: A sequence of ``(start_index, end_index)`` pairs, one
+            per block to delete. Each block may be a tuple or a list of
+            exactly two ints. Indexes are 1-based and inclusive, consistent
+            with :meth:`~gspread.worksheet.Worksheet.delete_rows`. Blocks may
+            be listed in any order but must not overlap.
+        :type blocks: Sequence[Sequence[int]]
+
+        :raises ValueError: If ``blocks`` is empty, a block does not contain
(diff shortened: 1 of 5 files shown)


antoineeripret:
@Ev2geny: merged :)

PR #1599, opened 2026-07-26
Stop BackOffHTTPClient retrying past the maximum backoff

uttam12331:
`BackOffHTTPClient` never stops retrying a persistently failing request.

The retry guard checks `wait <= self._MAX_BACKOFF`, but `wait` is computed as `min(2 ** self._NR_BACKOFF, self._MAX_BACKOFF)`, so it is clamped to `self._MAX_BACKOFF` (128) and the comparison is always true. The comment above it says the intent is to retry only "AND we did not reach the max retry limit" — but that limit check is dead. The same dead guard appears in both the `APIError` and `RefreshError` paths.

So a request that keeps returning 429 / 5xx keeps retrying (each retry is a recursive call) until Python's recursion limit (~1000), sleeping up to 128 s between attempts, instead of giving up after the backoff is exhausted.

Reproduced with a session that always returns 429: the request recurses ~994 times before failing, rather than stopping after ~7 retries.

The fix compares the un-clamped `2 ** self._NR_BACKOFF` against `self._MAX_BACKOFF`, so retries stop once the backoff reaches the maximum (after `_NR_BACKOFF` reaches 8, i.e. the initial call plus 7 retries). The sleep duration still uses the clamped `wait`. With the fix, the same permanently-failing request stops after 8 attempts and raises `APIError`.

The now-unused `wait` parameter of the nested `_should_retry` helper is removed.

Added `test_gives_up_after_max_backoff` (a session that always returns 429 must raise `APIError` after a bounded number of calls); it fails on the current code and passes with the change. Existing `tests/http_client_test.py` passes; `flake8` is clean.

### DIFF
--- a/gspread/http_client.py
+++ b/gspread/http_client.py
@@ -574,7 +574,6 @@ class BackOffHTTPClient(HTTPClient):
         def _should_retry(
             code: int,
             error: Mapping[str, Any],
-            wait: int,
         ) -> bool:
             # Drive API return a dict object 'errors', the sheet API does not
             if "errors" in error:
@@ -594,7 +593,7 @@ class BackOffHTTPClient(HTTPClient):
             return (
                 code in self._HTTP_ERROR_CODES
                 or code >= HTTPStatus.INTERNAL_SERVER_ERROR
-            ) and wait <= self._MAX_BACKOFF
+            ) and 2**self._NR_BACKOFF <= self._MAX_BACKOFF
 
         try:
             return super().request(*args, **kwargs)
@@ -606,7 +605,7 @@ class BackOffHTTPClient(HTTPClient):
             wait = min(2**self._NR_BACKOFF, self._MAX_BACKOFF)
 
             # check if error should retry
-            if _should_retry(code, error, wait) is True:
+            if _should_retry(code, error) is True:
                 time.sleep(wait)
 
                 # make the request again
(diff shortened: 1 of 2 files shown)


antoineeripret:
Hey @uttam12331, thanks for this  I reproduced it and the diagnosis is correct. 

On master a permanently 429 request retries 954 times before dying on the recursion limit, and your new test fails there (954 != 8) and passes on your branch. Full suite is green.

One thing before this can go in though: 

`_should_retry` returns True early for 403 + usageLimits, before reaching the comparison you patched. So that path never hits the new limit. I confirmed it's genuinely unbounded rather than incidentally deep: attempts scale with the recursion limit (91/291/591 at limits of 100/300/600). At the default limit that's ~950 attempts and roughly 35 hours of cumulative sleeping. This is the case the class docstring names as its known issue, so it's the one most likely to hurt in practice. 

**The early return True needs the same guard, plus a test case covering it.**

Thanks !

uttam12331:
Good catch, thanks - you're right, the 403 usageLimits branch returned `True` before ever reaching the backoff comparison, so that path stayed unbounded.

I've pulled the `2**self._NR_BACKOFF <= self._MAX_BACKOFF` check into a `within_max_backoff` local and the usageLimits branch now returns that too, so every retryable path shares the same bound. Added `test_gives_up_after_max_backoff_on_usage_limits` covering it - it fails on the old code (retries to the recursion limit) and stops at 8 calls now. Full suite green, flake8 clean.

